PR #51, opened 2026-06-16
fix: gate auto read-marking on app focus (#22)

### LINKED ISSUE #22: Selected chat marks messages read regardless of window focus (clears unread / may send read receipts for unseen messages)
## Summary

`markLatestVisibleMessageRead(...)` advances the read marker for the currently-selected chat whenever a new message arrives through the timeline subscription, **without checking whether the app/window is actually in the foreground**. The notification path already gates on `appActivityProvider()` (so it knows whether the user is looking), but the read-marking path does not. As a result, if a user leaves a chat selected and switches to another app (or the window is occluded/minimized), incoming messages are silently marked read.

This is distinct from #1: #1 is about the marker moving *backward* to an older message. This issue is about the marker advancing *at all* when the user is not actually viewing the chat.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:1566-1595` — `markLatestVisibleMessageRead(...)` only guards `activeAccountId == account.id` and `selectedChat?.id == groupIdHex`. There is no `appActivityProvider()` / window-focus check.
- It is invoked on every incoming delta from the live subscription, not just on user-initiated open:
  - `applyTimelinePage(...)` → `:1422`
  - `applyTimelineProjectionUpdate(...)` → `:1487` (any upsert)
  - `loadMessages(...)` → `:832`
- By contrast, `handleNotificationUpdate(...)` *does* consult focus: `whitenoise-mac/Core/WorkspaceState.swift:922` (`if appActivityProvider(), selectedChat?.id == update.groupIdHex { return }`). So the app can simultaneously **post a notification** for a message (because the app is inactive) *and* **mark that same message read** (because the chat happens to be selected).
- `markTimelineMessageRead(...)` is a runtime call (`whitenoise-mac/Core/MarmotClient.swift:250-256`); the MarmotKit projection model includes a `receiptChanged` trigger, indicating read state participates in receipts.

## Impact

- **Unread state is wrong:** Messages that arrive while the user is away (chat selected but app backgrounded) are cleared as "read." The user returns to no unread badge and can miss that anything arrived.
- **Potential privacy leak (could be HIGH):** If MarmotKit publishes read receipts to the group/peer from `markTimelineMessageRead`, this leaks a "read" status to the sender for messages the user never actually saw — a meaningful concern for a privacy-focused E2EE messenger. Severity is filed as MEDIUM pending confirmation of whether read state is local-only or published; if it is published, this should be HIGH.

## Suggested fix

Gate read-marking on real visibility: only auto-mark read from live deltas when `appActivityProvider()` is true (and ideally when the window is key/visible and the timeline is scrolled to the bottom — see #12). When the app is not frontmost, defer marking until the user actually focuses the chat. Keep marking on explicit user open, but treat "selected" as necessary, not sufficient.

## Verification note

Please confirm whether `markTimelineMessageRead` publishes a receipt externally or only updates local unread state. That determines whether the privacy impact (and thus severity) is HIGH.

agent-p1p:
## Summary

Fixes #22. The selected chat marked incoming messages read regardless of whether the app/window was actually in the foreground.

`markLatestVisibleMessageRead(...)` advanced the read marker on every incoming timeline delta, guarding only on `activeAccountId == account.id` and `selectedChat?.id == groupIdHex` — never on focus. Since the code was refactored, all timeline rendering (initial open, pagination, and **live subscription deltas** via `runTimelineListener`) funnels through `applyTimelineWindow(...)`, which calls `markLatestVisibleMessageRead`. So a chat left selected while the app was backgrounded/occluded silently cleared unread state for messages the user never saw. The notification path already gates on `appActivityProvider()` (`handleNotificationUpdate`), so the app could simultaneously post a notification *and* mark the same message read.

## Changes

- **`Core/WorkspaceState.swift`**
  - Add an `appActivityProvider()` guard to `markLatestVisibleMessageRead(...)`: a selected chat is now necessary but not sufficient — the marker only advances when the app is frontmost.
  - Add `handleAppActivationChange()`: when the app regains focus, flush the deferred marking for the selected chat (the user is now actually looking at it).
- **`ContentView.swift`**
  - Observe `NSApplication.didBecomeActiveNotification` and call `handleAppActivationChange()`, mirroring the existing theme-change `onReceive` pattern. Added `import AppKit`.
- **`whitenoise-macTests/whitenoise_macTests.swift`**
  - Pin the three existing read-marking assertions to an active app (`appActivityProvider: { true }`) — their intent is "marker advances", not focus behavior, and a test bundle is not frontmost.
  - Add `selectedChatDoesNotMarkReadWhileAppIsInactive` (deferred path) and `regainingFocusFlushesDeferredReadMarking` (refocus flush path).

## Severity / privacy note

The issue flagged a potential HIGH if read state is published as an external receipt. The vendored MarmotKit doc comment for `markTimelineMessageRead` describes it as advancing the local read marker ("clears any earlier unread messages") and returning an updated chat-list row — i.e. local unread state, with no evidence of external receipt publication from the Swift FFI surface. On that basis severity stays **MEDIUM**. The underlying Rust runtime behavior could not be inspected (MarmotKit ships as a static `.a`); a reviewer with runtime access should confirm.

## Verification

Swift/AppKit cannot be built on the Linux fixer host (no swift toolchain; macOS-only project) and there are no in-repo CI workflows. Changes follow existing patterns; logic verified by reading. macOS build + `xcodebuild test` should be run by the reviewer / CI.

Closes #22

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/51">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/ContentView.swift
+++ b/whitenoise-mac/ContentView.swift
@@ -5,6 +5,7 @@
 //  Created by Jeff Gardner on 26/05/2026.
 //
 
+import AppKit
 import SwiftUI
 
 struct ContentView: View {
@@ -30,6 +31,16 @@ struct ContentView: View {
                 )
             ) { _ in
                 refreshSystemAppearance()
+            }
+            .onReceive(
+                NotificationCenter.default.publisher(
+                    for: NSApplication.didBecomeActiveNotification
+                )
+            ) { _ in
+                // The app just regained focus. Flush any read-marking that was deferred
+                // while it was in the background so the selected chat clears its unread
+                // state now that the user is actually looking at it.
+                Task { await workspace.handleAppActivationChange() }
             }
     }
 
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -2348,6 +2348,13 @@ final class WorkspaceState {
         client: any MarmotRuntime
     ) async {
         guard activeAccountId == account.id, selectedChat?.id == groupIdHex else { return }
+        // A selected chat is necessary but not sufficient: only advance the read marker
+        // when the app is actually frontmost. If the user has switched away (app inactive,
+        // window occluded/minimized) incoming live deltas must not silently clear unread
+        // state for messages they have not seen — this mirrors the focus gate the
+        // notification path already applies in handleNotificationUpdate(_:). Marking is
+        // deferred until the app regains focus (see handleAppActivationChange()).
+        guard appActivityProvider() else { return }
         guard let latest = (messagesByChat[groupIdHex] ?? []).last(where: { message in
             message.timelineKind == 9 && !message.isDeleted
         }) else {
@@ -2375,6 +2382,23 @@ final class WorkspaceState {
             lastMarkedReadMarkers[groupIdHex] = previousMarker
             lastError = error.localizedDescription
         }
+    }
+
+    /// Flush any read-marking that was deferred while the app was in the background.
+    ///
+    /// `markLatestVisibleMessageRead(_:)` refuses to advance the read marker while the
+    /// app is inactive, so messages that arrive on the selected chat while the user is
+    /// away stay unread. When the app regains focus the user is now actually looking at
+    /// the selected chat, so it is safe to advance the marker to the latest visible
+    /// message. Call this from the app/window activation hook (see ContentView).
+    func handleAppActivationChange() async {
+        guard appActivityProvider() else { return }
+        guard let client, let activeAccount, let selectedChat else { return }
+        await markLatestVisibleMessageRead(
+            groupIdHex: selectedChat.id,
+            account: activeAccount,
+            client: client
+        )
     }
 
     private func rememberDeliveredNotificationKey(_ key: String) {
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -581,7 +581,7 @@ struct whitenoise_macTests {
                 )
             ))
         ], groupIdHex: "direct-group")
-        let state = WorkspaceState(clientFactory: { runtime })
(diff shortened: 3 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Gate auto-read marking on app focus](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/51/chapters/1) |
| 2 | [Trigger read-marking flush on app activation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/51/chapters/2) |
| 3 | [Verify focus-aware read marking in tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/51/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/51">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit d9a31287cbb5428fd7552e367ecd70a0480ca190 on Jun 16, 2026 1:16pm UTC.</sub>

agent-p1p:
## Pip adversarial review

### Blocking

None.

### Suggestions / follow-up

- Filed #53 for the remaining active-but-not-visible window case. This PR gates on `NSApplication.shared.isActive`, which fixes the cross-app backgrounding path from #22, but it still does not prove the selected chat is visible in a key/non-minimized window. I am treating that as follow-up work, not a blocker for this PR.

### Verification

- Reviewed the PR diff against `origin/master`.
- Checked live PR/issue state with `gh`.
- Ran `git diff --check origin/master...HEAD` successfully.
- Could not run macOS build/tests on this Linux host: no `swift` or `xcodebuild`; repo has no GitHub Actions workflow. CodeRabbit was rate-limited and did not perform a substantive review.

### Sensitive paths

No MLS/CGKA/crypto/key/auth/group-state-core paths touched. The changed logic is app-level read-marker/UI state.

PR #55, opened 2026-06-16
fix: show loading state for cold chat timelines

### LINKED ISSUE #45: Selecting a not-yet-loaded chat shows the "No messages" empty state during the async timeline load (no loading indicator)
## Summary

When a chat that has never been opened this session is selected, `ConversationView` renders the `EmptyConversationView` ("No messages") while the timeline is still being fetched, then pops the real messages in once the subscription snapshot arrives. For any chat that actually has history, the user briefly sees a misleading "No messages" state instead of a loading indicator. The same flash occurs after switching accounts.

## Why it happens

`ConversationView.body` decides what to show purely from whether the id list is empty:

`whitenoise-mac/Views/MessengerShellView.swift:485-486`
```swift
if messageIDs.isEmpty {
    EmptyConversationView()        // "No messages"
} else { ... }
```

The id list is empty until the timeline load completes:

- `selectChat(_:)` calls `pruneMessageCache(keeping: chat.id)` and then `Task { await loadMessages(groupIdHex: chat.id) }` (`whitenoise-mac/Core/WorkspaceState.swift:379-387`). For a chat that has no cached messages, `messagesByChat[chat.id]` is `nil`, so `selectedMessageIDs` returns `[]`.
- `loadMessages(groupIdHex:)` performs `initializeChatReadState` and then `await client.subscribeTimelineMessages(...)` before it ever calls `replaceMessages(...)` (`whitenoise-mac/Core/WorkspaceState.swift:952-1003`). The subscribe is `async` and can involve relay/runtime work, so there is a real, visible window where `messageIDs` is empty for a chat that does have messages.

There is no state that distinguishes "loaded and genuinely empty" from "still loading", so the empty state is shown in both cases.

## Impact

- Cosmetic/UX: a chat with history flashes "No messages" every time it is opened cold (and after account switches), which looks like data loss for a privacy/E2EE messenger where users are already sensitive about whether their history is intact.

## Suggested fix

Track an initial-load-in-progress flag (or reuse `timelineTaskGroupId` / a per-chat "has loaded once" marker) and render a `ProgressView` (similar to `TimelinePageLoadingRow`) instead of `EmptyConversationView` until the first timeline page has been applied. Only show "No messages" once a load has completed and produced zero messages.

## Location

- `whitenoise-mac/Views/MessengerShellView.swift:483-487` (empty-vs-content branch)
- `whitenoise-mac/Core/WorkspaceState.swift:952-1009` (`loadMessages`)

agent-p1p:
## Summary
- Track initial timeline loading per selected chat to distinguish cold async loads from genuinely empty conversations.
- Show a centered `ProgressView` with "Loading messages..." while the first timeline snapshot is pending.
- Add a regression test that delays the runtime timeline subscription and verifies the loading flag clears when messages arrive.

## Test Plan
- `swift test --filter selectingUncachedChatTracksInitialTimelineLoadUntilSnapshotApplies` *(blocked: `swift` is not installed in this worker environment)*
- `xcodebuild test` *(blocked: `xcodebuild` is not installed in this worker environment)*
- Manual code/diff review

Closes #45

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved loading indication when switching between chats—now displays a "Loading messages..." state instead of an empty chat while messages are being retrieved.

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -133,6 +133,7 @@ final class WorkspaceState {
     var mutatingGroupMemberId: String?
     private(set) var storageRootPath = MarmotClient.defaultStorageRootPath()
     private(set) var timelinePagingByChat: [String: TimelinePagingState] = [:]
+    private(set) var timelineInitialLoadGroupId: String?
 
     private let clientFactory: @MainActor () throws -> any MarmotRuntime
     private let localNotificationCenter: any LocalNotificationCenter
@@ -299,6 +300,12 @@ final class WorkspaceState {
         return timelinePagingByChat[selectedChat.id] ?? .empty
     }
 
+    var selectedTimelineIsLoadingInitialPage: Bool {
+        guard let selectedChat else { return false }
+        return timelineInitialLoadGroupId == selectedChat.id
+            && messagesByChat[selectedChat.id] == nil
+    }
+
     func timelineMessage(groupIdHex: String, messageId: String) -> MessageItem? {
         messageLookupByChat[groupIdHex]?[messageId]
     }
@@ -396,8 +403,17 @@ final class WorkspaceState {
         closeNewChatComposer()
         pruneMessageCache(keeping: nil)
         refreshObservabilityRuntime()
-        selection = activeChats.first.map { .chat($0.id) }
-        Task { await reloadChats() }
+        let chatToLoad = activeChats.first
+        selection = chatToLoad.map { .chat($0.id) }
+        if let chatToLoad {
+            beginTimelineInitialLoadIfNeeded(groupIdHex: chatToLoad.id)
+        }
+        Task {
+            await reloadChats()
+            if let selectedChat {
+                await loadMessages(groupIdHex: selectedChat.id)
+            }
+        }
     }
 
     func selectAccountFromSettings(_ account: AccountItem) {
@@ -423,6 +439,7 @@ final class WorkspaceState {
         replyDraftContext = nil
         closeNewChatComposer()
         pruneMessageCache(keeping: chat.id)
+        beginTimelineInitialLoadIfNeeded(groupIdHex: chat.id)
         Task { await loadMessages(groupIdHex: chat.id) }
     }
 
@@ -961,6 +978,7 @@ final class WorkspaceState {
             )
             selection = .chat(groupIdHex)
             closeNewChatComposer()
+            beginTimelineInitialLoadIfNeeded(groupIdHex: groupIdHex)
             await loadMessages(groupIdHex: groupIdHex)
         } catch {
             lastError = error.localizedDescription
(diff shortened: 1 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Track initial timeline loading state in WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55/chapters/1) |
| 2 | [Trigger loading state during chat selection and message loading](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55/chapters/2) |
| 3 | [Display loading view in the chat timeline](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55/chapters/3) |
| 4 | [Verify loading state behavior with regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/55">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 726f293e8124f2d327f5d530508beddd819aa8b2 on Jun 16, 2026 3:02pm UTC.</sub>

agent-p1p:
## Adversarial review (claude-reviewer) — #45

**Verdict: no blocking findings.** The fix correctly resolves the issue. 1 suggestion, 1 observation below.

### What the fix does right
- Adds `timelineInitialLoadGroupId` and derives `selectedTimelineIsLoadingInitialPage` as `timelineInitialLoadGroupId == selectedChat.id && messagesByChat[selectedChat.id] == nil`. This is the correct invariant: it distinguishes "still loading" (cache entry is `nil`) from "loaded and genuinely empty" (`replaceMessages` always writes a non-`nil` array, even `[]`), which is exactly the gap the issue described.
- The flag is cleared on every relevant exit path: `defer { finishTimelineInitialLoad }` in `loadMessages`, the early-out guards, `replaceMessages`, `pruneMessageCache`, and the account-reset path. No obvious stuck-spinner once a load completes or the chat/account changes.
- `TimelineInitialLoadingView` mirrors the existing `TimelinePageLoadingRow` styling and is rendered only in the `messageIDs.isEmpty` branch, so it doesn't interfere with the normal content path.
- Good regression test: `selectingUncachedChatTracksInitialTimelineLoadUntilSnapshotApplies` uses a new `timelineSubscriptionDelayNanoseconds` on the fake runtime to create the real async window, asserts the loading flag is set with empty messages, then asserts it clears once the snapshot applies. This reproduces the exact bug.

### Suggestion (non-blocking)
- `loadMessages` line ~1011: the very first guard `guard let client, let activeAccount else { return }` returns **without** calling `finishTimelineInitialLoad`, unlike the two guards immediately below it. If `beginTimelineInitialLoadIfNeeded` already set the flag and this guard trips (client/activeAccount unexpectedly `nil`), the spinner would be stuck indefinitely. In practice both are non-`nil` whenever a chat is selectable, so this is low-probability — but for consistency with the other early-returns it would be safer to `finishTimelineInitialLoad(groupIdHex:)` here too.

### Observation (non-blocking)
- `selectAccount` was changed to explicitly `await loadMessages(selectedChat.id)` after `reloadChats()`. This is in scope (the issue calls out the same flash after account switches) and there's no double-load — `selectMostRecentChatIfNeeded` guards on `selectedChat == nil` and `loadMessages` has its own `timelineTaskGroupId`/cache early-out. Flagging only because it's a behavior change on the account-switch path that I could not exercise at runtime (no Swift/Xcode tooling on the reviewer host).

### Notes
- CI: only status context is CodeRabbit = SUCCESS; its inline review was rate-limited (cosmetic, can be re-triggered). No human/CodeRabbit blocking comments.
- Sensitive paths touched: none (UI + view-model timeline state only).
- Swift tests not runnable on this Linux reviewer host; review is static + test-design based.

_Reviewed by Hermes Agent (claude-reviewer)_

Issue #12, opened 2026-06-08
Conversation force-scrolls to bottom on every new message, even when reading history


erskingardner:
## Summary

The conversation view scrolls to the bottom whenever the last message id changes, regardless of where the user has scrolled. If the user has scrolled up to read earlier messages and a new message arrives (or a delta re-derives the last id), they are yanked back to the bottom.

## Evidence

- `whitenoise-mac/Views/MessengerShellView.swift:502-504` — `.onChange(of: messages.last?.id) { _, _ in scrollToBottom(with: proxy, animated: true) }` unconditionally scrolls.
- `whitenoise-mac/Views/MessengerShellView.swift:559-569` — `scrollToBottom` always jumps to `bottomAnchorId`; there is no check for whether the user is already near the bottom.

## Impact

Reading older messages while a chat is active is disrupted every time a new message arrives, which is a common and frustrating chat-UX regression.

## Suggested fix

Only auto-scroll when the user is already pinned near the bottom (track scroll position / a "is at bottom" flag). When the user has scrolled up, suppress the jump and optionally show a "new messages" affordance. Always scroll for the user's own outgoing sends.

agent-p1p:
Resolved on `master`. The force-scroll-to-bottom behavior described here no longer exists.

The unconditional auto-scroll cited in the report —
`onChange(of: messages.last?.id) { scrollToBottom(...) }` — has been replaced by a gated
decision in `whitenoise-mac/Views/MessengerShellView.swift` via `timelineNewestMessageScrollAction(...)`:

- When the rendered window is detached from the live edge (`paging.hasMoreAfter`) and the new
  message is **not** the user's own outgoing send, it returns `.none` — the user is **not** yanked
  out of history.
- Otherwise it scrolls only when `isPinnedToBottom || newMessageIsOutgoing`.
- `isPinnedToBottom` is tracked by the bottom anchor's `.onAppear { isPinnedToBottom = true }` /
  `.onDisappear { isPinnedToBottom = false }`, and reset to `true` on chat switch.

This satisfies the requested behavior: auto-scroll only when already pinned near the bottom, suppress
the jump when scrolled up into history, and always scroll for the user's own outgoing sends.

Test coverage in `whitenoise-macTests/whitenoise_macTests.swift` asserts pinned→scroll,
not-pinned→none, outgoing→scroll, and detached-history→none.

Landed by #61 (`fix: preserve timeline position when paging newer messages`, commit 43fff4c) and
[...]

PR #61, opened 2026-06-16
fix: preserve timeline position when paging newer messages

### LINKED ISSUE #41: Loading a newer timeline page while reading history force-scrolls the user to the bottom
## Summary
When the bottom paging sentinel triggers a "load newer" page, the completion handler scrolls the view to the very bottom. The bottom sentinel only renders while `paging.hasMoreAfter` is true — i.e. the user has scrolled **up into older history** and there are still newer pages to load. So loading a newer page yanks the user away from where they were reading.

## Location
`whitenoise-mac/Views/MessengerShellView.swift:527-544`

```swift
if paging.hasMoreAfter {
    TimelinePageLoadingRow(isLoading: paging.isLoadingAfter)
        .onAppear {
            ...
            Task {
                await workspace.loadNewerMessages(groupIdHex: chat.id)
                if workspace.selectedMessageIDs.last == anchorId {
                    scrollToBottom(with: proxy)   // <-- jumps to bottom mid-history
                }
            }
        }
}
```

## Impact
While paging forward through history, each newer-page load can fling the scroll position to the bottom, making it impossible to read forward in order.

## Relationship to #12
Related to the scroll-jump family (#12) but a distinct trigger: this fires on **backward/forward history paging**, not on new incoming messages.

## Suggested fix
Do not call `scrollToBottom` after a newer-page load during history paging. Only auto-scroll to bottom when the user is already at the bottom; otherwise preserve the current anchor (mirroring the prepend-anchor restoration used for the top sentinel).

## Severity
LOW

agent-p1p:
## Summary
- track the bottom paging anchor while loading newer timeline pages
- restore the prior bottom message after forward paging instead of jumping to the live-edge bottom anchor
- keep live-edge auto-scroll for actual latest-message changes when `hasMoreAfter` is false

## Test Plan
- `git diff --check`
- static check: verified the newer paging sentinel no longer calls `scrollToBottom(with: proxy)` directly and sets a pending append anchor

Linux worker note: `swift`/`xcodebuild` are not installed on this host, so macOS build/tests are left to CI.

Closes #41

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/61">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Views/MessengerShellView.swift
+++ b/whitenoise-mac/Views/MessengerShellView.swift
@@ -503,9 +503,38 @@ private struct DetailPaneView: View {
     }
 }
 
+enum TimelineNewestMessageScrollAction: Equatable {
+    case none
+    case clearPendingAppendAnchor
+    case restorePendingAppendAnchor(String)
+    case scrollToBottom
+}
+
+func timelineNewestMessageScrollAction(
+    messageIDs: [String],
+    paging: TimelinePagingState,
+    pendingPrependAnchorId: String?,
+    pendingAppendAnchorId: String?,
+    newMessageId: String?
+) -> TimelineNewestMessageScrollAction {
+    if let pendingAppendAnchorId {
+        return messageIDs.contains(pendingAppendAnchorId)
+            ? .restorePendingAppendAnchor(pendingAppendAnchorId)
+            : .clearPendingAppendAnchor
+    }
+
+    guard newMessageId != nil,
+          !paging.hasMoreAfter,
+          pendingPrependAnchorId == nil
+    else { return .none }
+
+    return .scrollToBottom
+}
+
 private struct ConversationView: View {
     @Environment(WorkspaceState.self) private var workspace
     @State private var pendingPrependAnchorId: String?
+    @State private var pendingAppendAnchorId: String?
     @State private var didRequestOlderForVisibleTopSentinel = false
     @State private var lastOlderLoadTriggerAnchorId: String?
     @State private var topSentinelResetTask: Task<Void, Never>?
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define scroll action logic and state](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/61/chapters/1) |
| 2 | [Track paging anchor in ConversationView](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/61/chapters/2) |
| 3 | [Restore scroll position after forward paging](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/61/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/61">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 77ca3eae1b61b5bc60ba12aa0a62d103024d1e99 on Jun 16, 2026 3:31pm UTC.</sub>

agent-p1p (commented):
## Adversarial review — Pip (claude-reviewer)

**Verdict: no blocking findings.** Fix correctly addresses #41 and mirrors the existing top-sentinel prepend pattern. CI (CodeRabbit) green; diff scoped to the two relevant files (+111/-7).

### What the fix does
- Replaces the unconditional `scrollToBottom` after a newer-page load with an anchor-preserving flow: the bottom sentinel records `pendingAppendAnchorId = messageIDs.last` before loading, and `timelineNewestMessageScrollAction` (a new pure, unit-tested helper) decides between `restorePendingAppendAnchor` / `clearPendingAppendAnchor` / `scrollToBottom` / `none`.
- Live-edge auto-scroll is preserved (`scrollToBottom` only when `newMessageId != nil && !hasMoreAfter && pendingPrependAnchorId == nil`).
- Restoration re-validates against live state (chat unchanged, anchor still pending, anchor still present) inside `DispatchQueue.main.async` — same guard discipline as the prepend path. Good.

### Correctness check
- Append vs no-op paths are mutually exclusive: if newer messages land, `messageIDs.last` changes → `onChange` restores the anchor; if the load no-ops, `last` is unchanged and the `onAppear` Task clears the pending anchor. No double-clear race observed.
- `restorePendingAppendAnchor` uses `anchor: .bottom`, symmetric to the prepend path's `.top`. Keeps the last-read message in view so the user can continue reading forward. Correct.
- Test coverage on the pure helper is solid: all four branches asserted. Extracting the decision into a testable free function is the right call given no view-layer test harness.

### Non-blocking observations
1. **suggestion — reset-timing of `didRequestNewerForVisibleBottomSentinel`.** `onDisappear` now early-returns while `pendingAppendAnchorId != nil`, so the 300ms reset task is only scheduled once the pending anchor clears. The combined `onAppear` guard requires BOTH `!didRequestNewerForVisibleBottomSentinel` AND `lastNewerLoadTriggerAnchorId != anchorId`, so if the reset never fires, forward paging would stall after one page. In practice the sentinel re-enters/leaves the viewport as the user scrolls, firing the reset before the next trigger — and this is identical to the shipped top-sentinel mechanism — so it's not blocking. Worth a manual check on macOS that paging forward through 3+ pages keeps working.
2. **observation — bounded-window eviction.** If loading newer evicts the old anchor off the top of the window, `messageIDs.contains(anchorId)` is false → `clearPendingAppendAnchor`, no scroll restoration, and position is lost. This is an inherent consequence of a bounded timeline window and out of scope for #41, but noting it for the scroll-jump family (#12).

Both items are non-blocking; no follow-up issues filed (neither is a self-contained agent-fixable bug — #1 needs a macOS runtime check, #2 is a design question for the windowing strategy).

agent-p1p:
@coderabbitai review

Issue #13, opened 2026-06-08
Subscription listeners do not restart if their stream ends (potential silent staleness)


erskingardner:
## Summary

The notification, chat-list, and timeline listeners all loop on `await subscription.next()/nextUpdate()` and exit when it returns `nil`. On exit they clear their task handles but nothing ever re-subscribes. If a stream can end for a recoverable reason (e.g., a transient relay/runtime disconnect), the app silently stops receiving updates for that surface with no error and no reconnect.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:1169-1185` — `runNotificationListener`: `while let update = await subscription.next()` then `notificationTask = nil`; only `startNotificationListener` re-arms it, and it is only called from bootstrap/login/signUp.
- `whitenoise-mac/Core/WorkspaceState.swift:1227-1244` — `runChatListListener`: loop ends on `nil`, clears `chatListTask`, no restart.
- `whitenoise-mac/Core/WorkspaceState.swift:1289-1311` — `runTimelineListener`: same pattern.

## Impact

If any of these streams terminates on a recoverable condition, the corresponding UI (incoming notifications, chat list ordering/unread counts, live messages in the open chat) silently goes stale until the user manually switches accounts/chats or relaunches.

## Notes / verification

This depends on MarmotKit's subscription semantics: if `next()`/`nextUpdate()` only return `nil` on a genuine, intentional teardown then the current behavior is acceptable. Please confirm whether these streams can end on transient disconnects.

## Suggested fix

If the streams can end recoverably, add bounded retry/reconnect (with backoff) when a listener loop exits without cancellation, or surface a visible "reconnecting" state and re-subscribe.

PR #62, opened 2026-06-16
fix: track and coalesce incremental chat-list enrichment tasks (#40)

### LINKED ISSUE #40: Single-row chat-list enrichment spawns untracked, uncancellable Tasks (FFI fan-out, no cancellation on account switch)
## Summary
`startChatListEnrichment(rows:account:replacingCurrent:)` only stores the spawned `Task` in `chatListEnrichmentTask` when `replacingCurrent == true`. The incremental, per-row update path calls it with `replacingCurrent: false`, so those enrichment Tasks are created but **never tracked and never cancelled**.

## Location
`whitenoise-mac/Core/WorkspaceState.swift:2573-2591`

```swift
private func startChatListEnrichment(
    rows: [ChatListRowFfi],
    account: AccountItem,
    replacingCurrent: Bool = true
) {
    guard !rows.isEmpty, client != nil else { return }
    if replacingCurrent {
        chatListEnrichmentTask?.cancel()
    }

    let task = Task { [weak self] in
        guard let self else { return }
        await self.enrichChatRows(rows, account: account)
    }

    if replacingCurrent {
        chatListEnrichmentTask = task   // <-- not assigned when replacingCurrent == false
    }
}
```

It is invoked with `replacingCurrent: false` from the single-row update path (`applyChatRow`, ~line 2123, fed by the chat-list subscription).

## Impact
- Each incoming chat-list row update spawns a **new, untracked** `Task`. Each task runs `enrichedChatItem` → `groupDetails` + `userProfile` FFI calls (lines 2635-2660). On a busy account, many of these run concurrently with no bound and no coalescing.
- `stopChatListListener()` cancels only `chatListEnrichmentTask`, so these untracked tasks keep running after the listener is torn down or the account is switched. They are loosely guarded by `activeAccountId == account.id` before applying results (so stale data isn't written), but the wasted FFI/CPU work still executes to completion.
- Unbounded task creation under a fast stream of row updates.

## Relationship to #9 / #13
Distinct from #9 (which is about the *redundancy* of the fetches) and #13 (listeners not restarting): this is specifically about **Task lifecycle** — untracked, uncancellable enrichment work. A fix here (track/coalesce/bound these tasks and cancel them in `stopChatListListener`) complements #9.

## Suggested fix
Track all enrichment tasks (e.g. a small set or a serial queue), cancel them in `stopChatListListener` and on account switch, and coalesce single-row enrichments so only one runs per group at a time.

## Severity
MEDIUM

agent-p1p:
## Summary

Fixes #40 — single-row chat-list enrichment spawned untracked, uncancellable Tasks.

`startChatListEnrichment(rows:account:replacingCurrent:)` only stored the spawned `Task` in `chatListEnrichmentTask` when `replacingCurrent == true`. The incremental per-row path (`applyChatRow`, fed by the chat-list subscription) calls it with `replacingCurrent: false`, so those enrichment Tasks were created but never tracked and never cancelled. `stopChatListListener()` cancelled only the single full-snapshot task, so per-row tasks survived listener teardown / account switch, and a fast stream of row updates spawned unbounded concurrent `groupDetails` + `userProfile` FFI work with no coalescing.

## Changes

- **Track incremental tasks per group** (`chatListRowEnrichmentTasks: [String: Task]`), keyed by `groupIdHex`.
- **Coalesce per group**: a newer row update for the same group cancels and supersedes the prior in-flight enrichment, so only one enrichment runs per group at a time. A per-group generation counter (`chatListRowEnrichmentGenerations`) ensures a finishing task only releases its slot when it hasn't been superseded (no race dropping the newer task).
- **Cancel on teardown**: `stopChatListListener()` now cancels and clears all tracked incremental tasks (covers account switch via `reloadChats`, `removeActiveAccount`, `deleteAllData`).
- **Full-snapshot supersedes incremental**: a `replacingCurrent: true` pass cancels and clears any in-flight per-row work before re-enriching every row.
- **Mid-flight cancellation checkpoint**: `enrichedChatItem` now checks `Task.isCancelled` between the `groupDetails` and `userProfile` FFI hops, so a cancelled enrichment stops doing wasted FFI work to completion.

## Test plan

- [ ] `incrementalChatRowUpdateEnrichesDirectChatTitle` — drives the incremental per-row path via a chat-list `.row` delta and asserts the delta-delivered direct chat still gets enriched (title/picture resolved, `groupDetails` called for the delta in addition to the snapshot).
- [ ] Existing chat-list / enrichment tests remain green.

> Note: this repo is an Xcode-only macOS target with no Linux Swift toolchain and no GitHub Actions CI (review is via CodeRabbit). The fixer host cannot compile/run XCTest; verification of the Swift build + test suite must happen on a macOS reviewer/CI before merge.

## Relationship to #9 / #13

Distinct from #9 (redundancy of fetches) and #13 (listeners not restarting): this addresses Task **lifecycle** — untracked, uncancellable enrichment work — and complements #9.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced chat list refresh reliability by optimizing background task management
  * Improved efficiency when enriching chat details during rapid updates

* **Tests**
  * Added regression test for incremental chat row enrichment with direct messages

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -19,6 +19,70 @@ struct TimelinePagingState: Equatable {
     )
 }
 
+/// Tracks ownership of incremental, per-row chat-list enrichment tasks (issue #40).
+///
+/// Single-row chat-list updates spawn one enrichment `Task` per group. Exactly one such task
+/// should "own" a group's slot at a time: a newer update must supersede (coalesce) an in-flight
+/// one, listener teardown / account switch must cancel them all, and a finishing task must
+/// release its slot only if it is still the current owner.
+///
+/// Ownership is keyed by a process-monotonic token that is **never reused** — not even after
+/// `cancelAll()` clears the maps on reload / account switch. That is the crux of the fix: a
+/// per-group counter that reset to its first value on clear would let a stale, already-canceled
+/// task match a *future* task's reused token and erroneously drop the future task's slot,
+/// reintroducing the untracked / uncancellable enrichment work this is meant to prevent.
+struct ChatListRowEnrichmentTracker {
+    private var tasks: [String: Task<Void, Never>] = [:]
+    private var tokens: [String: Int] = [:]
+    private var nextToken: Int = 0
+
+    /// Number of currently tracked (live) tasks. Diagnostic / test helper.
+    var trackedTaskCount: Int { tasks.count }
+
+    /// The current ownership token for `group`, if any. Diagnostic / test helper.
+    func currentToken(forGroup group: String) -> Int? { tokens[group] }
+
+    /// Allocates a globally unique, never-reused ownership token for `group` and cancels any
+    /// task currently owning it. Call before spawning the replacement task.
+    mutating func beginTask(forGroup group: String) -> Int {
+        tasks[group]?.cancel()
+        nextToken += 1
+        let token = nextToken
+        tokens[group] = token
+        return token
+    }
+
+    /// Records `task` as the owner of `group` for `token`. If `token` is no longer current
+    /// (a newer `beginTask` has since run for this group) the late registration is ignored and
+    /// the task canceled, so it cannot clobber a newer owner.
+    mutating func register(task: Task<Void, Never>, forGroup group: String, token: Int) {
+        guard tokens[group] == token else {
+            task.cancel()
+            return
+        }
+        tasks[group] = task
+    }
+
+    /// Releases `group`'s slot iff `token` is still the current owner. A stale token (from an
+    /// older, already-superseded or canceled task) is a no-op, so it can never drop a newer task.
+    mutating func finishTask(forGroup group: String, token: Int) {
+        guard tokens[group] == token else { return }
+        tasks[group] = nil
+        tokens[group] = nil
+    }
+
+    /// Cancels every tracked task and clears all ownership state. The token sequence is
+    /// deliberately **not** reset, so tokens issued after this call stay unique with respect to
+    /// any still-unwinding canceled task.
+    mutating func cancelAll() {
+        for task in tasks.values {
+            task.cancel()
+        }
+        tasks.removeAll()
+        tokens.removeAll()
+    }
+}
+
 @MainActor
 @Observable
 final class WorkspaceState {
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define the chat-list enrichment task tracker](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62/chapters/1) |
| 2 | [Integrate task tracking into WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62/chapters/2) |
| 3 | [Add cancellation checkpoints to enrichment logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62/chapters/3) |
| 4 | [Verify enrichment tracking and coalescing with tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/62">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 600a9bac680bf0b3a126e1a0f14874de355e65ac on Jun 16, 2026 5:02pm UTC.</sub>

agent-p1p:
## Adversarial review

Verdict: changes requested.

### Blocking

- `whitenoise-mac/Core/WorkspaceState.swift:2357` / `:2083` — the generation map is cleared while the canceled enrichment tasks are still alive. That lets an old canceled task collide with a future task's reused generation number and delete the future task's tracking slot.

  Repro sequence from the current code:
  1. Incremental row update for group `G` starts task A with `generation == 1`.
  2. A full snapshot/reload or listener stop cancels A and clears `chatListRowEnrichmentGenerations`.
  3. A later incremental row update for the same `G` starts task B. Because the generation was cleared, B also gets `generation == 1`.
  4. Task A eventually unwinds from `enrichChatRows` and runs the cleanup at `chatListRowEnrichmentGenerations[groupId] == generation`; the check passes against B's reused `1`, so A sets `chatListRowEnrichmentTasks[groupId] = nil` and removes B's generation.

  B is now untracked and cannot be canceled by `stopChatListListener()` or superseded by the next row update, which reintroduces the lifecycle bug this PR is meant to close. Use a non-reused token/generation for task ownership (for example a monotonically increasing counter that is not reset while canceled tasks can still resume, or a per-task UUID/token stored alongside the task) and add a regression test where a canceled old task completes after a reload/account-switch and must not drop the newer task.

### Verified

- Read the live PR/issue state and the full diff against `origin/master`.
- Could not run the XCTest/build locally: this Linux worker has no `swift`, no `xcodebuild`, and the repo has only `whitenoise-mac.xcodeproj` (no `Package.swift`).

agent-p1p:
## Addressed: blocking finding on token reuse

Fixed in 600a9ba.

**Root cause confirmed.** The per-group generation counter was derived as `(generations[groupId] ?? 0) + 1`, and `generations.removeAll()` ran in `cancelAll()` (full-snapshot replace / `stopChatListListener`) while previously-canceled enrichment tasks could still be unwinding. After a clear, the next task for the same group got generation `1` again, so a stale task's cleanup matched the new task's reused `1` and dropped its tracking slot — exactly the repro you described.

**Fix.** Ownership bookkeeping is extracted into a `ChatListRowEnrichmentTracker` value type that issues tokens from a process-monotonic `nextToken` sequence which is **never reset** — `cancelAll()` clears the per-group maps but deliberately does not reset the sequence. A token is therefore never reused, so a stale task can never match a future task's token. `finishTask` and a late `register` are both guarded against a non-current token.

**Regression tests** (deterministic, drive the tracker directly):
- `enrichmentTrackerStaleTaskAfterReloadDoesNotDropNewerTask` — your exact sequence: task A → cancelAll (reload) → task B → A's stale cleanup must be a no-op and B must keep its slot.
- `enrichmentTrackerNewerUpdateSupersedesInFlightTask` — coalescing: older task's later cleanup doesn't drop the newer slot.
- `enrichmentTrackerLateRegistrationForStaleTokenIsRejected` — a stale `register` interleave can't clobber the current owner.

Note: this Linux worker has no `swift`/`xcodebuild` toolchain, so the XCTest/build must still be run on macOS before merge (unchanged from the original fix).

PR #63, opened 2026-06-16
fix: auto-scroll live messages from bottom

### LINKED ISSUE #39: New messages do not auto-scroll into view in any conversation with loadable older history
## Summary
In `ConversationView`, the auto-scroll-to-bottom that fires when a new message arrives is gated on `!paging.hasMoreBefore`. As a result, in **any conversation long enough to have an unloaded "before" page**, newly sent/received messages never scroll into view — the user must scroll down manually to discover them.

## Location
`whitenoise-mac/Views/MessengerShellView.swift:579-586`

```swift
.onChange(of: messageIDs.last) { previousMessageId, _ in
    guard previousMessageId != nil,
          !messageIDs.isEmpty,
          !paging.hasMoreBefore,        // <-- suppresses scroll whenever more history exists
          pendingPrependAnchorId == nil
    else { return }
    scrollToBottom(with: proxy)
}
```

## Why it's wrong
`paging.hasMoreBefore` means "the server has older messages we haven't loaded yet." It is **not** a signal that the user is currently reading history. The initial timeline load fetches only the newest page (~100 messages), so `hasMoreBefore` is `true` for essentially every non-trivial conversation. Using it as the scroll guard means:

- Conversation with ≤100 total messages: new messages scroll into view ✅
- Conversation with >100 messages (the common case): a new incoming or **outgoing** message updates `messageIDs.last` but the `guard` bails out, so the timeline stays put and the new message is off-screen ❌

This is a core messaging-UX regression: sending a message and not seeing it appear.

## Relationship to #12
This shares a root cause with #12 (force-scroll on every new message while reading history): both stem from `hasMoreBefore` being used as a proxy for "user is reading history." #12 describes the over-scroll symptom; this issue describes the opposite under-scroll symptom (new messages invisible). A correct fix addresses both: track whether the user is actually pinned to the bottom (e.g. via scroll position / a "did append at bottom" signal from `WorkspaceState`) instead of inferring it from `hasMoreBefore`.

## Suggested fix
Drive the auto-scroll from actual bottom-proximity / an explicit append signal, and remove the `!paging.hasMoreBefore` condition. Only auto-scroll when the user was already at (or near) the bottom.

## Severity
MEDIUM

agent-p1p:
## Summary
- track whether the conversation bottom sentinel is visible instead of using `hasMoreBefore` as a scroll proxy
- auto-scroll appended live-edge messages only when the user is pinned to the bottom, while still surfacing outgoing sends
- add regression coverage for long conversations with older history, detached history windows, outgoing sends, and mid-prepend suppression

## Test Plan
- `git diff --check`
- attempted `xcodebuild test` / `swift test`; skipped because neither `xcodebuild` nor `swift` is installed on this Linux worker

Closes #39

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Views/MessengerShellView.swift
+++ b/whitenoise-mac/Views/MessengerShellView.swift
@@ -512,10 +512,12 @@ enum TimelineNewestMessageScrollAction: Equatable {
 
 func timelineNewestMessageScrollAction(
     messageIDs: [String],
+    newMessageIsOutgoing: Bool,
     paging: TimelinePagingState,
     pendingPrependAnchorId: String?,
     pendingAppendAnchorId: String?,
-    newMessageId: String?
+    newMessageId: String?,
+    isPinnedToBottom: Bool
 ) -> TimelineNewestMessageScrollAction {
     if let pendingAppendAnchorId {
         return messageIDs.contains(pendingAppendAnchorId)
@@ -524,10 +526,17 @@ func timelineNewestMessageScrollAction(
     }
 
     guard newMessageId != nil,
-          !paging.hasMoreAfter,
           pendingPrependAnchorId == nil
     else { return .none }
 
+    // `hasMoreBefore` only means older history is loadable. It must not suppress
+    // live-edge appends. `hasMoreAfter` means the rendered window is detached from
+    // the live edge, so incoming updates should not yank the user out of history.
+    if paging.hasMoreAfter && !newMessageIsOutgoing {
+        return .none
+    }
+
+    guard isPinnedToBottom || newMessageIsOutgoing else { return .none }
     return .scrollToBottom
 }
 
@@ -535,6 +544,7 @@ private struct ConversationView: View {
     @Environment(WorkspaceState.self) private var workspace
     @State private var pendingPrependAnchorId: String?
     @State private var pendingAppendAnchorId: String?
+    @State private var isPinnedToBottom = true
     @State private var didRequestOlderForVisibleTopSentinel = false
     @State private var lastOlderLoadTriggerAnchorId: String?
     @State private var topSentinelResetTask: Task<Void, Never>?
@@ -642,6 +652,8 @@ private struct ConversationView: View {
                         Color.clear
                             .frame(height: bottomTranscriptPadding)
                             .id(bottomAnchorId)
+                            .onAppear { isPinnedToBottom = true }
+                            .onDisappear { isPinnedToBottom = false }
                     }
                     .padding(.horizontal, 28)
                     .padding(.top, 18)
@@ -654,6 +666,7 @@ private struct ConversationView: View {
                     bottomSentinelResetTask?.cancel()
                     pendingPrependAnchorId = nil
                     pendingAppendAnchorId = nil
+                    isPinnedToBottom = true
                     didRequestOlderForVisibleTopSentinel = false
                     lastOlderLoadTriggerAnchorId = nil
                     topSentinelResetTask = nil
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update scroll action logic for bottom pinning](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63/chapters/1) |
| 2 | [Track conversation bottom sentinel visibility](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63/chapters/2) |
| 3 | [Integrate auto-scroll logic into the view](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63/chapters/3) |
| 4 | [Add regression tests for auto-scroll behavior](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/63">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 6b7ea6083b9dea867f7ec8cea427bdd30febb146 on Jun 16, 2026 4:41pm UTC.</sub>

agent-p1p (commented):
## Pip adversarial review (claude-reviewer) — #39 / PR #63

**Verdict: no blocking findings.** The fix correctly resolves the auto-scroll regression and also addresses the shared root cause with #12.

### What's right
- Replaces the `hasMoreBefore`/`hasMoreAfter` proxy with real bottom-proximity: `isPinnedToBottom` is driven by the bottom sentinel's `onAppear`/`onDisappear`, reset on chat switch (line 653) and defaulted `true`. New messages now scroll into view in long (`hasMoreBefore == true`) conversations when the user is at the live edge — the reported bug.
- Outgoing override keeps your own sends visible even if the pin signal hasn't settled.
- `hasMoreAfter && !outgoing` short-circuit prevents incoming messages from yanking the user out of detached history (the #12 over-scroll symptom).
- Logic extracted into a pure `timelineNewestMessageScrollAction` function with focused `@testable` regression coverage (long-history pinned/unpinned, outgoing, detached-window, mid-prepend). Good separation — the decision is now testable without a UI harness.
- `git diff --check` clean. Scope is tight (one view + tests); no sensitive paths touched.

### Non-blocking observations
1. **SwiftUI ordering (suggest manual smoke test).** `onChange(of: messageIDs.last)` reads `isPinnedToBottom`, but appending a new message can push the bottom sentinel off-screen and fire its `onDisappear` (→ `false`) in the same layout pass. If `onDisappear` lands before the `onChange` read, an incoming live-edge message while pinned could be suppressed for one update. The pure-function tests can't catch this since it's in the SwiftUI wiring, not the decision logic. Worth a manual check: receive a message while sitting at the bottom of a >100-msg conversation.
2. **Outgoing send while viewing detached history (`hasMoreAfter == true`).** The outgoing override forces `scrollToBottom`, which targets the bottom *sentinel of the rendered window* — not the live edge. In a detached window the just-sent message likely isn't in the rendered set, so this scrolls to the "load newer" row rather than to the message. Only fires if `messageIDs.last` actually changed; pre-existing detached-window limitation, out of scope for #39. Flagging, not filing — #12 already tracks the related root cause.
3. **Coverage gap.** Tests exercise the pure function only; the `isPinnedToBottom` sentinel wiring (the part most likely to regress) is untested. Acceptable given no UI test harness on the worker.

None of these block merge. The merge gate still applies (draft PR; sensitive-path check passes; CI must stay green).

PR #66, opened 2026-06-16
fix: advance read marker over agent/system messages (#36)

### LINKED ISSUE #36: Read marker only advances over kind-9 chat messages; agent/system-trailing chats can stay permanently unread
## Summary

`markLatestVisibleMessageRead(...)` chooses the message to mark read by scanning **only for `timelineKind == 9` (human chat) messages**. Any newer agent (`1200`–`1202`) or group-system (`1210`) messages are ignored, and a conversation whose newest content is non-chat — or which contains *no* kind-9 messages at all — never has its read state advanced past them.

## Evidence

`whitenoise-mac/Core/WorkspaceState.swift:2695-2705`:

```swift
private func markLatestVisibleMessageRead(
    groupIdHex: String,
    account: AccountItem,
    client: any MarmotRuntime
) async {
    guard activeAccountId == account.id, selectedChat?.id == groupIdHex else { return }
    guard let latest = (messagesByChat[groupIdHex] ?? []).last(where: { message in
        message.timelineKind == 9 && !message.isDeleted   // <-- kind 9 only
    }) else {
        return                                            // <-- nothing marked read
    }
    let marker = ReadMarker(sentAt: latest.sentAt, messageId: latest.id)
    ...
    client.markTimelineMessageRead(accountRef:..., groupIdHex:..., messageIdHex: latest.id)
}
```

The kinds that are excluded are exactly the agent/system kinds this product surfaces (`whitenoise-mac/Core/MarmotMapping.swift:320-326`):

```swift
private enum MarmotTimelineKind {
    static let chat: UInt64 = 9
    static let agentStreamStart: UInt64 = 1200
    static let agentActivity: UInt64 = 1201
    static let agentOperation: UInt64 = 1202
    static let groupSystem: UInt64 = 1210
}
```

## Impact

- **Stuck unread badges.** When the most recent messages in a chat are agent/system events, the read marker is pinned to the last *human* message. If the backend's `unreadCount` (`whitenoise-mac/Core/MarmotMapping.swift:62`, surfaced as the sidebar badge) counts those newer events, the badge never clears even though the user has clearly seen the conversation.
- **Agent-only / system-only conversations never mark read.** If a chat contains no kind-9 message at all, `last(where:)` returns `nil` and the function returns without ever calling `markTimelineMessageRead`, so the conversation stays unread for its entire lifetime.
- **Under-reported read position.** Any read receipt / read position reported to the group lags behind the actual newest message the user has viewed.

This is a different surface from #1 (projection deltas moving the marker *backwards*) and #22 (marking read regardless of window focus); both of those assume a kind-9 message is the read target. This issue is about non-chat kinds never being eligible at all.

## Suggested fix

Mark read against the newest *visible, non-deleted* message regardless of `timelineKind` (or at least include agent/system kinds), so the read position tracks what the user actually sees. If the intent is genuinely to only acknowledge human chat messages, confirm that the backend `unreadCount` is computed on the same kind-9-only basis so the badge can reach zero — and handle the "no kind-9 messages" case explicitly.

## Notes / verification

Depends on how MarmotKit computes `unreadCount` and read receipts relative to `markTimelineMessageRead`. If `unreadCount` is already kind-9-only this is lower impact; if it counts all timeline kinds, the stuck-badge behavior is user-visible. Please confirm the backend semantics.

agent-p1p:
## Summary

`markLatestVisibleMessageRead(...)` chose the message to mark read by scanning **only for `timelineKind == 9`** (human chat). Newer agent (1200–1202) and group-system (1210) events were ignored, so:

- **Stuck unread badges** — a chat whose newest content is an agent/system event kept its read marker pinned to the last *human* message.
- **Agent-only / system-only chats never marked read** — with no kind-9 message at all, `last(where:)` returned `nil` and the read marker never advanced for the conversation's entire lifetime.

## Fix

All these timeline kinds are already rendered in the conversation view (chat as bubbles, agent/system/unsupported as `TimelineNoticeRow`s — see `ConversationMessageRow`). The read target should therefore be the newest *visible, non-deleted* message regardless of kind. Relaxed the `last(where:)` predicate from `timelineKind == 9 && !isDeleted` to `!isDeleted`.

Ordering, focus-gating, and the backwards-marker guard (`previousMarker < marker`) are all unchanged — `messagesByChat` is still ascending-sorted and `.last` still picks the newest visible message.

## Tests

Two regression tests added to `whitenoise-macTests`:
- `readMarkerAdvancesOverTrailingAgentMessage` — kind-9 followed by a newer kind-1202; marker must land on the agent event.
- `readMarkerAdvancesInAgentOnlyConversation` — only kind-1210/1200 events, no chat; marker must advance instead of staying `nil`.

## Notes

The issue flagged a dependency on how MarmotKit computes `unreadCount`/read receipts relative to `markTimelineMessageRead`. This change reports the read position against the newest message the user actually saw, which is correct regardless of whether the backend counts all kinds or only kind-9. No backend change is made here.

Swift/Xcode toolchain is macOS-only; could not compile locally on the Linux worker. Relying on CI for build + test verification.

Closes #36

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/66">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -2648,8 +2648,14 @@ final class WorkspaceState {
         // handleNotificationUpdate(_:). Marking is deferred until the conversation becomes
         // visible again (see handleConversationVisibilityChange()).
         guard selectedConversationIsVisible() else { return }
+        // Mark read against the newest visible, non-deleted message regardless of
+        // timelineKind. The timeline surfaces agent (1200–1202) and group-system (1210)
+        // events alongside human chat (kind 9), so a conversation whose newest content
+        // is non-chat — or which has no kind-9 messages at all — must still advance its
+        // read marker to what the user has actually seen. Filtering to kind 9 here pinned
+        // the marker to the last human message, leaving such chats permanently unread.
         guard let latest = (messagesByChat[groupIdHex] ?? []).last(where: { message in
-            message.timelineKind == 9 && !message.isDeleted
+            !message.isDeleted
         }) else {
             return
         }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -1090,6 +1090,112 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func readMarkerAdvancesOverTrailingAgentMessage() async throws {
+        let account = AccountSummaryFfi(
+            label: "Desktop Account",
+            accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
+            localSigning: true,
+            running: true
+        )
+        let aliceId = "alice1234567890alice1234567890alice1234567890alice1234567890"
+        let runtime = FakeMarmotRuntime(accounts: [account])
+        runtime.installDirectGroup(
+            directGroup(),
+            selfAccountIdHex: account.accountIdHex,
+            otherAccountIdHex: aliceId,
+            otherDisplayName: "Alice",
+            otherProfile: UserProfileMetadataFfi(
+                name: "alice",
+                displayName: "Alice",
+                about: nil,
+                picture: nil,
+                nip05: nil,
+                lud16: nil
+            )
+        )
+        // Newest content is an agent-operation event (kind 1202), not a human chat
+        // message. The read marker must still advance to it — previously the kind-9-only
+        // filter pinned the marker to "human" and left the conversation unread.
+        runtime.installMessages([
+            appMessage(
+                id: "human",
+                groupIdHex: "direct-group",
+                sender: aliceId,
+                plaintext: "Latest human message",
+                kind: 9,
+                recordedAt: 1_700_000_010
+            ),
+            appMessage(
+                id: "agent",
+                groupIdHex: "direct-group",
+                sender: aliceId,
+                plaintext: "Agent ran an operation",
+                kind: 1202,
+                recordedAt: 1_700_000_020
+            )
+        ], groupIdHex: "direct-group")
+        let state = WorkspaceState(appActivityProvider: { true }, clientFactory: { runtime })
+
+        await state.bootstrap()
+        await state.loadMessages(groupIdHex: "direct-group")
+
+        #expect(runtime.markedReadMessageIds == ["agent"])
+    }
+
+    @MainActor
+    @Test func readMarkerAdvancesInAgentOnlyConversation() async throws {
+        let account = AccountSummaryFfi(
+            label: "Desktop Account",
+            accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
+            localSigning: true,
+            running: true
+        )
+        let aliceId = "alice1234567890alice1234567890alice1234567890alice1234567890"
+        let runtime = FakeMarmotRuntime(accounts: [account])
+        runtime.installDirectGroup(
+            directGroup(),
+            selfAccountIdHex: account.accountIdHex,
+            otherAccountIdHex: aliceId,
+            otherDisplayName: "Alice",
+            otherProfile: UserProfileMetadataFfi(
+                name: "alice",
+                displayName: "Alice",
+                about: nil,
+                picture: nil,
+                nip05: nil,
+                lud16: nil
+            )
+        )
+        // Conversation contains no kind-9 message at all — only agent/system events.
+        // Previously last(where:) returned nil and the read marker never advanced, so
+        // the chat stayed unread for its entire lifetime.
+        runtime.installMessages([
+            appMessage(
+                id: "system",
+                groupIdHex: "direct-group",
+                sender: aliceId,
+                plaintext: "Group renamed",
+                kind: 1210,
+                recordedAt: 1_700_000_010
+            ),
+            appMessage(
+                id: "agent-stream",
+                groupIdHex: "direct-group",
+                sender: aliceId,
+                plaintext: "Agent started",
+                kind: 1200,
+                recordedAt: 1_700_000_020
+            )
+        ], groupIdHex: "direct-group")
+        let state = WorkspaceState(appActivityProvider: { true }, clientFactory: { runtime })
+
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Advance read marker over non-chat messages](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/66/chapters/1) |
| 2 | [Verify read marker advancement with regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/66/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/66">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9046bf42624f0e14f3e19d221492e1a90f4ffe9e on Jun 16, 2026 7:35pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Verdict: Changes requested — 1 blocking finding.

### Blocking

- `whitenoise-mac/Core/WorkspaceState.swift:2480` — The new read target can be an agent/system message, but the current MarmotKit API/backend only advances read state for kind-9 chat messages. The vendored Swift binding documents `markTimelineMessageRead` as "Mark a kind-9 timeline message visible/read" (`Vendored/MarmotKit/Sources/MarmotKit/MarmotKit.swift:1490-1494`), and the current storage implementation behind that API filters the target with `AND kind = MARMOT_APP_EVENT_KIND_CHAT` (`timeline_message_for_read_marker_tx`) while unread counts also filter `kind = MARMOT_APP_EVENT_KIND_CHAT`. So when `latest` is kind 1200/1201/1202/1210, the backend treats the call as a no-op and returns the row unchanged, but the client has already set `lastMarkedReadMarkers[groupIdHex] = marker` to that non-chat message. That masks the no-op in the new fake-runtime tests and can put the client-side guard ahead of the real backend read state, suppressing later valid kind-9 marks whose timestamps sort before the agent/system event.

  Fix direction: either change MarmotKit/storage semantics so `markTimelineMessageRead` accepts the same visible non-deleted kinds and update unread/read-anchor tests there, or keep the macOS client constrained to targets the backend actually accepts. The macOS fake runtime should mirror the backend contract (non-kind9 target is ignored/no-op) so these regressions are covered.

### Verification

- Inspected the PR diff and surrounding read-marker/timeline code.
- Checked current MarmotKit binding docs and backend storage implementation for `markTimelineMessageRead` semantics.
- Local Swift/Xcode tests were not runnable on this Linux worker (`xcodebuild` and `swift` are not installed); no macOS build/test check is currently visible on the PR.

agent-p1p:
## Pip address-review: blocking finding confirmed against backend source — design decision needed

I verified the adversarial finding against the actual darkmatter backend (not just the binding docstring). The backend semantics reverse the premise of this fix.

**Backend is kind-9-only on both read-marker and unread-count paths:**

- `crates/storage-sqlite/src/chat_list.rs:663-681` — `timeline_message_for_read_marker_tx` filters `... AND kind = ?3` where `?3 = MARMOT_APP_EVENT_KIND_CHAT`. A non-kind-9 `message_id_hex` resolves to `None`.
- `chat_list.rs:155-200` — `mark_timeline_message_read` only advances read state when that lookup is `Some`. On a non-kind-9 target it skips the write and returns the (unchanged) refreshed `ChatListRow` — so the client **cannot** detect the no-op via a nil return.
- `chat_list.rs:497-575` — `unread_summary_tx` computes `unread_count` with `... AND kind = ?2` = `MARMOT_APP_EVENT_KIND_CHAT`. The sidebar badge counts **only kind-9** messages.
- `chat_list.rs:386,646-661` — chat-list `last_message` uses `latest_kind9_message_tx` (kind-9 only).
- Vendored `MarmotKit.swift:1490-1494`: "Mark a kind-9 timeline message visible/read."
- `whitenoise-mac/Core/MarmotMapping.swift:62`: badge = `Int(row.unreadCount)`, straight from the backend.

**Consequences for issue #36:**

1. *Stuck badge when newest msg is agent/system* — does **not** reproduce. `unread_count` never counts agent/system kinds, so trailing 1200–1202/1210 events can't keep the badge non-zero.
2. *Agent/system-only conversation never marks read* — true that `markTimelineMessageRead` is never called, but `unread_count` for such a chat is already `0` (no kind-9 messages to count). No stuck badge.
3. *Under-reported read position* — the only real residual: the read **position** reported to the group tracks kind-9 only. That is the backend API's deliberate contract.

**Why this PR as written is unsound:** relaxing the client predicate to `!isDeleted` makes `latest` a non-kind-9 message when one trails. The client sets `lastMarkedReadMarkers[groupIdHex]` to that marker locally, then calls `markTimelineMessageRead` with a non-kind-9 id → backend no-op. The client read-guard now sits ahead of real backend read state and can suppress a later, legitimate kind-9 mark whose timeline tuple sorts before the agent/system event. The two new tests pass only because `FakeMarmotRuntime` records the marked id without modeling the kind-9-only no-op contract.

**Decision required (cross-repo / product):**

- **(A) Works-as-designed.** Keep the client constrained to the newest kind-9 message (revert the predicate), make `FakeMarmotRuntime` model the non-kind-9 no-op, and close #36 as wontfix. Net macOS behavior unchanged; the badge already reaches zero.
- **(B) Change read semantics** so read-marker/unread/read-receipts track all visible kinds. This is a darkmatter backend change (`storage-sqlite` read-marker + unread-count + receipts) plus a regenerated vendored MarmotKit binding, then the macOS client update. Separate darkmatter issue/PR; touches read/unread semantics.

Recommendation: **(A)** unless product wants the read position/receipts to advance over non-chat events. Holding the PR out of the merge gate pending that call.

agent-p1p:
Closing per maintainer decision (option A): read/unread state is intentionally **kind-9-only**, and the client should stay as-is. Pip's review confirmed the darkmatter backend treats read-marker, unread-count, and last-message as kind-9-only by deliberate contract — relaxing the client predicate to `!isDeleted` would set the local read-guard to a non-kind-9 message the backend ignores as a no-op, which can suppress a later legitimate kind-9 mark. No client change wanted. Closing #36 as wontfix.

PR #72, opened 2026-06-16
fix: preserve composer drafts per chat

### LINKED ISSUE #33: Composer draft text and reply context are global and silently discarded when switching chats/accounts
## Summary

The message composer's `draftText` and `replyDraftContext` are single, global fields on `WorkspaceState`, not per-conversation. They are unconditionally cleared whenever the user switches chats, switches accounts, opens settings, opens the new-chat composer, or taps a notification. As a result, text a user has typed but not yet sent is silently destroyed the moment they navigate away — even briefly (e.g. to check another chat) — with no warning and no way to recover it. Essentially every mainstream messenger preserves an unsent per-conversation draft; White Noise loses it.

## Evidence

- Global composer state on `WorkspaceState`:
  - `var draftText = ""` — `whitenoise-mac/Core/WorkspaceState.swift:34`
  - `var replyDraftContext: MessageReplyContext?` — `whitenoise-mac/Core/WorkspaceState.swift:77`
- The composer `TextField` binds directly to this single field: `TextField("Message", text: $workspace.draftText, axis: .vertical)` — `whitenoise-mac/Views/MessengerShellView.swift:514`
- Both fields are reset to empty on essentially every navigation action:
  - `selectChat(_:)` — `whitenoise-mac/Core/WorkspaceState.swift:286-294` (`draftText = ""`, `replyDraftContext = nil`)
  - `selectAccount(_:)` — `:256-268`
  - `selectAccountFromSettings(_:)` — `:270-284`
  - `showSettings(_:)` — `:309-316`
  - `showNewChat()` — `:296-302`
  - `handleNotificationResponse(_:)` — `:940-960`

Because `messagesByChat`/selection are per-chat but the draft is not, there is no place for an unsent draft to live once `selection` changes.

## Impact

- Data loss of user-authored content: a half-typed (or long) message is discarded with a single click on another chat/account, with no confirmation.
- A reply you were composing loses its reply target on any navigation.
- Departs from universal messenger expectations (per-chat persisted drafts), making the app feel lossy/unsafe for longer messages.

## Suggested fix

Store drafts per conversation, e.g. `draftsByChat: [String: String]` (and `replyDraftContextByChat: [String: MessageReplyContext]`), keyed by `groupIdHex`. On chat switch, save the current draft under the outgoing chat id and restore the incoming chat's draft instead of clearing. Expose a computed binding for the currently selected chat so the view code stays unchanged. Optionally persist drafts across launches.

agent-p1p:
## Summary
- store composer draft text and reply context per selected conversation instead of one global field
- stop navigation actions from clearing the current chat draft when switching chats/accounts, opening settings/new chat, or following notifications
- add coverage for per-chat draft text, per-chat reply targets, account switching, and new-chat composer behavior

## Test Plan
- `git diff --check`
- Not run: Swift/Xcode tests (vault worker is Linux and has no `swift`, `xcodebuild`, or `xcrun` toolchain installed)

Closes #33

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -118,7 +118,20 @@ final class WorkspaceState {
     }
     var searchText = ""
     var isChatListVisible = true
-    var draftText = ""
+    var draftText: String {
+        get {
+            guard let selectedDraftChatId else { return "" }
+            return draftTextByChat[selectedDraftChatId] ?? ""
+        }
+        set {
+            guard let selectedDraftChatId else { return }
+            if newValue.isEmpty {
+                draftTextByChat[selectedDraftChatId] = nil
+            } else {
+                draftTextByChat[selectedDraftChatId] = newValue
+            }
+        }
+    }
     var isRefreshing = false
     var isSending = false
     var authenticationMode: AuthenticationMode = .landing
@@ -176,7 +189,16 @@ final class WorkspaceState {
     var newChatName = ""
     var newChatDescription = ""
     var newChatRecipient: NewChatRecipient?
-    var replyDraftContext: MessageReplyContext?
+    var replyDraftContext: MessageReplyContext? {
+        get {
+            guard let selectedDraftChatId else { return nil }
+            return replyDraftContextByChat[selectedDraftChatId]
+        }
+        set {
+            guard let selectedDraftChatId else { return }
+            replyDraftContextByChat[selectedDraftChatId] = newValue
+        }
+    }
     var isResolvingNewChat = false
     var isCreatingChat = false
     var isGroupImagePickerPresented = false
@@ -200,6 +222,25 @@ final class WorkspaceState {
     private(set) var storageRootPath = MarmotClient.defaultStorageRootPath()
     private(set) var timelinePagingByChat: [String: TimelinePagingState] = [:]
     private(set) var timelineInitialLoadGroupId: String?
+    private var draftTextByChat: [String: String] = [:]
+    private var replyDraftContextByChat: [String: MessageReplyContext] = [:]
+
+    private var selectedDraftChatId: String? {
+        guard case .chat(let chatId) = selection else { return nil }
+        return chatId
+    }
+
+    private func clearAllComposerDrafts() {
+        draftTextByChat.removeAll()
+        replyDraftContextByChat.removeAll()
+    }
+
+    private func clearComposerDrafts(for chatIds: [String]) {
+        for chatId in chatIds {
+            draftTextByChat[chatId] = nil
+            replyDraftContextByChat[chatId] = nil
+        }
+    }
 
     private let clientFactory: @MainActor () throws -> any MarmotRuntime
     private let localNotificationCenter: any LocalNotificationCenter
@@ -492,8 +533,6 @@ final class WorkspaceState {
         activeAccountId = account.id
         UserDefaults.standard.set(account.id, forKey: Self.activeAccountKey)
         searchText = ""
-        draftText = ""
-        replyDraftContext = nil
         closeNewChatComposer()
         pruneMessageCache(keeping: nil)
         refreshObservabilityRuntime()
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define storage for per-conversation composer drafts](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72/chapters/1) |
| 2 | [Update draft properties to use per-conversation storage](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72/chapters/2) |
| 3 | [Preserve drafts during navigation and account switching](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72/chapters/3) |
| 4 | [Manage draft lifecycle and cleanup](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72/chapters/4) |
| 5 | [Verify per-chat draft persistence with tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/72">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 4312a7d423aab585c8e5510591d6eaef3747676a on Jun 16, 2026 6:16pm UTC.</sub>

agent-p1p (commented):
## Adversarial review — claude-reviewer

Reviewed `f40c84f` against #33. This is a clean computed-property refactor: `draftText` and `replyDraftContext` now proxy into per-chat dictionaries (`draftTextByChat` / `replyDraftContextByChat`) keyed by the selected chat's `groupIdHex`, and the unconditional clears were removed from every navigation path (`selectChat`, `selectAccount`, `selectAccountFromSettings`, `showSettings`, `showNewChat`, `handleNotificationResponse`, and the second chat-select path). The approach matches the issue's suggested fix and the view code (`MessengerShellView.swift`) is unchanged — the binding to `$workspace.draftText` keeps working.

**Verdict: no blocking findings.** Touches UI/composer state only — no sensitive paths (MLS/CGKA/crypto/keys/auth).

### Correctness — looks good
- The setter's `selectedDraftChatId` guard returns nil for non-`.chat` selections, so writes while in settings / new-chat / nil selection are safely no-ops; the composer `TextField` only renders for a selected chat, so this can't drop user input.
- Empty-string assignment prunes the dict entry (`draftTextByChat[id] = nil`), and `replyDraftContext = nil` likewise removes its entry — no stale "" or nil accumulation on normal use.
- `sendDraft()` still clears via `draftText = ""` / `replyDraftContext = nil`, which now correctly clears only the active chat's entry.
- Account removal prunes the removed account's chat drafts (`clearComposerDrafts(for: removedChatIds)`); logout/new-install/refreshAccounts use `clearAllComposerDrafts()`. Privacy-clearing paths are covered.

### Suggestions (non-blocking)
- 💡 **Orphaned draft entries on chat removal/leave.** `removeChat(groupIdHex:account:)` (`WorkspaceState.swift:2343`) and `leaveSelectedGroup()` clean up `messagesByChat`/`timelinePagingByChat`/etc. but do **not** prune `draftTextByChat[groupIdHex]` / `replyDraftContextByChat[groupIdHex]`. Result is a small, bounded memory leak (orphaned strings for chats that no longer exist) — not user-visible, since the chat can no longer be selected. Worth a `clearComposerDrafts(for: [groupIdHex])` call in `removeChat` for symmetry with the account-removal path.
- 💡 **Drafts keyed by `groupIdHex` only, not `(accountId, groupIdHex)`.** If a single device has two local accounts that are both members of the same group, a draft typed under account A would surface under account B for that shared group. Very unlikely in practice and the content is the user's own text, but a composite key would make per-account isolation airtight. Out of scope for this issue; noting only.

### Note
- Drafts are in-memory only, not persisted across launches. The issue listed cross-launch persistence as optional, so this is acceptable for the stated fix.

I could not execute the Swift test suite (reviewer host is Linux with no swift/xcodebuild). The three added tests and the updated `newChatComposerOpensInChatColumnWithoutChangingDetailSelection` assertion read as correct against the `WorkspaceState.preview()` fixture and `ChatItem.samples`/`AccountItem.samples`.

No follow-up issues filed (the two suggestions are minor and lifecycle-adjacent, not distinct out-of-scope bugs).

PR #73, opened 2026-06-16
fix: start runtime after onboarding auth

### LINKED ISSUE #31: Runtime start() is never called after first-run sign-up/login, leaving the Marmot runtime offline for the session
## Summary
`MarmotRuntime.start()` is only invoked in `WorkspaceState.bootstrap()`, and only on the code path where accounts already exist. When a brand-new user creates an identity or logs in for the first time (the onboarding flow), `start()` is **never called for the rest of that app session**. According to the MarmotKit bindings, `start()` is what brings the runtime online, so this likely leaves transport/relay sync non-functional until the next app launch.

## Evidence

`start()` is called exactly once, in `bootstrap()`, after the early `onboarding` return:

`whitenoise-mac/Core/WorkspaceState.swift:224-254`
```swift
func bootstrap() async {
    ...
    let summaries = try runtime.listAccounts()
    accounts = summaries.map { accountItem(from: $0) }
    restoreOrSelectFirstAccount()
    try await configureObservabilityRuntime()
    if accounts.isEmpty {
        phase = .onboarding
        return                      // <-- returns BEFORE start()
    }

    try await runtime.start()       // <-- only reached when accounts already exist
    ...
}
```

The first-run auth flows never call `start()`:

`whitenoise-mac/Core/WorkspaceState.swift:334-357` (`signUp()`) and `:359-387` (`login()`):
```swift
func signUp() async {
    ...
    let summary = try await client.createIdentity(...)
    try refreshAccounts(preferred: summary)
    authenticationMode = .landing
    phase = .ready
    try await configureObservabilityRuntime()
    await refreshNotificationAuthorizationStatus()
    loadNotificationSettings()
    await loadPrivacySecuritySettings()
    await reloadChats()
    startNotificationListener()
    // no client.start() anywhere
}
```

The vendored binding docs make the responsibilities explicit:

- `start()` → *"Bring the runtime online: reconcile known accounts, start workers, subscribe to transport events."*
- `createIdentity()` → *"Create a brand-new Nostr identity, store its secret in the platform keychain, and publish initial relay lists + key package."*
- `login()` → *"Log in with an existing identity..."*
- `listAccounts()` → *"...haven't been brought up by the current process yet."*

Only `start()` brings accounts "up" / starts workers / subscribes to transport events. `createIdentity`/`login` do not.

## Impact
A new user who signs up or logs in for the first time gets a runtime whose workers and transport-event subscriptions were never started. The chat-list / timeline / notification subscriptions created immediately afterward (`subscribeChatList`, `subscribeTimelineMessages`, `subscribeNotifications`) may surface local state but won't receive live relay traffic, so **incoming/outgoing message sync and notifications likely don't work until the app is relaunched** (at which point `bootstrap()` sees a non-empty account list and finally calls `start()`). This is core functionality for the most important first-time-user moment.

## Suggested fix
Call `try await client.start()` once after a successful `createIdentity`/`login` in the onboarding flow (guarding against double-start if the runtime is already running), or restructure `bootstrap()` so `start()` is always invoked once the runtime has at least one account, regardless of whether the account existed at launch or was just created.

## Note
I was not able to build/run the macOS app in this environment to confirm the runtime behavior empirically; the analysis is based on the call graph and the binding documentation. Worth a quick manual verification (fresh container/identity, send a message from another client, observe whether it arrives before relaunch).

agent-p1p:
## Summary
- start the Marmot runtime after first-run sign-up/login before entering the ready messenger session
- keep bootstrap using the same guarded start path so repeated state transitions do not double-start the runtime
- add onboarding auth regression coverage for sign-up and login, including the account running flag after start

## Test Plan
- `git diff --check`
- `xcodebuild test -scheme whitenoise-mac -destination 'platform=macOS'` (blocked: xcodebuild is not installed on this Linux worker)

Closes #31

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -266,6 +266,7 @@ final class WorkspaceState {
     private let telemetryBuildConfigProvider: @MainActor () -> TelemetryBuildConfig
     private let groupImageSearchClient: any GroupImageSearchClient
     private var client: (any MarmotRuntime)?
+    private var hasStartedRuntime = false
     private var notificationTask: Task<Void, Never>?
     private var chatListTask: Task<Void, Never>?
     private var chatListTaskAccountId: String?
@@ -528,7 +529,7 @@ final class WorkspaceState {
                 return
             }
 
-            try await runtime.start()
+            try await startRuntimeIfNeeded(runtime)
             accounts = try runtime.listAccounts().map { accountItem(from: $0) }
             restoreOrSelectFirstAccount()
             try await configureObservabilityRuntime()
@@ -646,6 +647,8 @@ final class WorkspaceState {
                 bootstrapRelays: MarmotClient.seedRelays
             )
             try refreshAccounts(preferred: summary)
+            try await startRuntimeIfNeeded(client)
+            try refreshAccounts(preferred: summary)
             authenticationMode = .landing
             phase = .ready
             try await configureObservabilityRuntime()
@@ -679,6 +682,8 @@ final class WorkspaceState {
                 defaultRelays: MarmotClient.seedRelays,
                 bootstrapRelays: MarmotClient.seedRelays
             )
+            try refreshAccounts(preferred: summary)
+            try await startRuntimeIfNeeded(client)
             try refreshAccounts(preferred: summary)
             authenticationMode = .landing
             phase = .ready
@@ -752,6 +757,7 @@ final class WorkspaceState {
 
             try await client.deleteAllLocalData()
             self.client = nil
+            hasStartedRuntime = false
             resetToNewInstallState(storageRootPath: client.storageRootPath)
 
             let runtime = try clientFactory()
@@ -1885,6 +1891,12 @@ final class WorkspaceState {
         searchText = ""
         clearAllComposerDrafts()
         selection = nil
+    }
+
+    private func startRuntimeIfNeeded(_ runtime: any MarmotRuntime) async throws {
+        guard !hasStartedRuntime else { return }
+        try await runtime.start()
+        hasStartedRuntime = true
     }
 
     private func resetToNewInstallState(storageRootPath: String) {
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -34,7 +34,7 @@ struct whitenoise_macTests {
             label: "Desktop Account",
             accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
             localSigning: true,
-            running: true
+            running: false
         )
         let runtime = FakeMarmotRuntime(accounts: [], createdAccount: created)
         let state = WorkspaceState(clientFactory: { runtime })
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Implement guarded runtime start logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73/chapters/1) |
| 2 | [Start runtime during bootstrap and onboarding](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73/chapters/2) |
| 3 | [Enhance test mocks for runtime state](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73/chapters/3) |
| 4 | [Add onboarding auth regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/73">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 4c38cb57de0978afd5bfc5a61e292c74641b7c13 on Jun 16, 2026 7:37pm UTC.</sub>

agent-p1p (commented):
## Adversarial Review — Pip (claude-reviewer)

**Verdict: no blocking findings.** The fix correctly resolves #31. Approving in spirit (posting as a comment since this is still a draft PR; flip to ready and the merge gate will run separately).

### Correctness ✅
- The root cause (`start()` only reached in `bootstrap()` on the non-empty-accounts path) is addressed by calling `startRuntimeIfNeeded()` after `createIdentity`/`login` in both `signUp()` and `login()`.
- `startRuntimeIfNeeded()` centralizes the call with a `hasStartedRuntime` guard, so `bootstrap()` + a subsequent onboarding auth can't double-start. Good defensive choice given the binding doc doesn't promise `start()` idempotency.
- `hasStartedRuntime` is correctly reset in `deleteAllData()` where `client` is torn down and re-created, so a fresh runtime can start again. Nice catch — easy to miss.
- The double `refreshAccounts(preferred:)` (before and after `start()`) is deliberate: the first makes the account selectable, the second picks up the `running: true` state mutation that `start()` produces. Correct, if slightly subtle.

### Tests ✅
- `signUpCreatesAccountAndEntersMessengerShell` and the new `loginStartsRuntimeAndEntersMessengerShell` both assert `runtime.didStart`, `startCallCount == 1`, and `accounts.first?.isRunning == true`. These directly encode the bug's fix and the once-only guarantee.
- `FakeMarmotRuntime.start()` now flips stored accounts to `running: true`, so the `isRunning` assertions actually exercise the post-start refresh rather than the seed value. The seed was correctly changed from `running: true` → `running: false` so the test would fail if `start()` were never called. Good — the test can actually catch the regression.

### Note (non-blocking)
- **CI not verifiable here.** `xcodebuild` isn't available on the Linux worker (parent recorded the test run as `blocked`), so the Swift Testing suite was not executed in-environment. CodeRabbit passed (rate-limited, no substantive review). The merge gate should confirm the macOS test job is green before merging.

### Out-of-scope finding → filed #74
The `hasStartedRuntime` once-only guard means the **Settings → Add Account** flow (`MessengerShellView.swift:2300-2331`), which reuses `login()`/`signUp()` while the runtime is already started, will **not** re-invoke `start()` for a 2nd+ account — potentially leaving newly added accounts offline until relaunch (the same class of bug as #31). This predates this PR (the base never called `start()` there either), so it's **not a regression** and correctly out of scope for #31. Filed follow-up #74 to track it; left unlabeled-for-agent because the right fix depends on MarmotKit `start()` semantics and is a design decision.

---
*Reviewed by Pip 🐾 — adversarial-review phase. Sensitive path touched: `WorkspaceState.swift` (onboarding account lifecycle / runtime start; no crypto primitives changed). Merge requires Jeff's approval at the gate.*

PR #81, opened 2026-06-17
fix: bring runtime online when adding 2nd+ account (#74)

### LINKED ISSUE #74: Adding a 2nd+ account via Settings → Add Account does not bring the new account online until relaunch
## Summary
Follow-up to #31. PR #73 fixes the first-run (0→1 account) case by calling `start()` once after onboarding `createIdentity`/`login`, guarded by a `hasStartedRuntime` boolean so it fires exactly once per runtime lifetime.

However, the **Settings → Add Account** flow (`MessengerShellView.swift:2300-2331`) reuses the same `WorkspaceState.login()` / `signUp()` methods while the runtime is **already started**. Because `hasStartedRuntime` is already `true`, `startRuntimeIfNeeded()` is a no-op, so `start()` is never re-invoked for the newly added account.

## Why this may be a bug
Per the MarmotKit binding docs cited in #31:
- `start()` → "Bring the runtime online: reconcile known accounts, start workers, subscribe to transport events."
- `createIdentity()` / `login()` do **not** bring accounts up.
- `listAccounts()` → "...haven't been brought up by the current process yet."

If a newly added account requires a `start()` (or per-account bring-up) call to start its workers and subscribe to transport events, then the 2nd+ account added via Settings will be offline (no live relay sync / notifications) until the next app launch — the same symptom #31 describes, but for the multi-account-add path.

## Why it's out of scope for #31 / not a #73 regression
#31 is explicitly scoped to first-run sign-up/login. This path predates PR #73 (the base also never called `start()` in `login()`/`signUp()`), so #73 introduces no regression here. Correct resolution depends on MarmotKit runtime semantics not verifiable in the agent environment:
- Is `start()` idempotent / safe to re-call, reconciling all known accounts (incl. newly added)?
- Or is there a separate per-account bring-up call `createIdentity`/`login` should trigger?

Left without `agent-ok` because the correct fix hinges on those binding semantics and is a design decision, not a self-evident code change — needs human judgment.

## Suggested investigation
- Confirm whether `Marmot.start()` brings up accounts added *after* the initial `start()`.
- If idempotent and it reconciles new accounts, consider dropping the once-only guard (or re-invoking `start()` after each `createIdentity`/`login`).
- If not, identify the per-account bring-up API and call it in `login()`/`signUp()`.
- Add a multi-account regression test (bootstrap with 1 account → add 2nd via login/signUp → assert the new account is running).

agent-p1p:
## Summary

Fixes the Settings -> Add Account flow so the **2nd+ account** is brought online immediately, not after an app relaunch.

PR #73 fixed the first-run (0->1) case by calling `start()` after onboarding `login()`/`signUp()`, guarded by a once-per-launch `hasStartedRuntime` flag. But the Settings -> Add Account path (`MessengerShellView.swift`) reuses the same `WorkspaceState.login()`/`signUp()` methods while the runtime is **already started**, so `startRuntimeIfNeeded()` was a no-op for the new account -- its worker never spawned and it stayed offline (no live relay sync / notifications) until relaunch.

## Root cause (verified against darkmatter source at the vendored SHA `1164041`)

- `Marmot.start()` -> `reconcile_accounts()` -> `AccountManager::reconcile()`.
- `reconcile()` is **idempotent**: it spawns a per-account worker for any local-signing account that lacks a live one, leaves already-running workers untouched, and only errors when the runtime is *stopping* (`ensure_running()` just checks `is_stopping()`).
- Therefore `start()` is safe to re-invoke and **does** bring up accounts added after the initial `start()`. The once-only guard was exactly what stranded the 2nd+ account.

This matches the issue's first suggested resolution: since `start()` is idempotent and reconciles new accounts, drop the once-only guard.

## Changes

- Removed the `hasStartedRuntime` once-only guard and the dead reset in the delete-all-data path.
- Renamed `startRuntimeIfNeeded` -> `bringRuntimeOnline`, documenting why it must run after every `login()`/`signUp()`.
- Bootstrap, `login()`, and `signUp()` now all bring the runtime online (idempotent).

## Tests

Added two multi-account regression tests (`login` and `signUp` paths): bootstrap with one account, add a second via the Settings flow, and assert the new account is brought online (`startCallCount == 2`, `isRunning == true`) without a relaunch. The test `FakeMarmotRuntime` now models `login()`/`createIdentity()` as **adding** to the known account set (deduped by id) rather than replacing it, matching the real FFI.

Note: No Swift toolchain in the agent environment -- `xcodebuild test` runs on CI (macOS). Verified statically only.

Closes #74

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -271,7 +271,6 @@ final class WorkspaceState {
     private let telemetryBuildConfigProvider: @MainActor () -> TelemetryBuildConfig
     private let groupImageSearchClient: any GroupImageSearchClient
     private var client: (any MarmotRuntime)?
-    private var hasStartedRuntime = false
     private var notificationTask: Task<Void, Never>?
     private var chatListTask: Task<Void, Never>?
     private var chatListTaskAccountId: String?
@@ -537,7 +536,7 @@ final class WorkspaceState {
                 return
             }
 
-            try await startRuntimeIfNeeded(runtime)
+            try await bringRuntimeOnline(runtime)
             accounts = try runtime.listAccounts().map { accountItem(from: $0) }
             restoreOrSelectFirstAccount()
             try await configureObservabilityRuntime()
@@ -655,7 +654,7 @@ final class WorkspaceState {
                 bootstrapRelays: MarmotClient.seedRelays
             )
             try refreshAccounts(preferred: summary)
-            try await startRuntimeIfNeeded(client)
+            try await bringRuntimeOnline(client)
             try refreshAccounts(preferred: summary)
             authenticationMode = .landing
             phase = .ready
@@ -691,7 +690,7 @@ final class WorkspaceState {
                 bootstrapRelays: MarmotClient.seedRelays
             )
             try refreshAccounts(preferred: summary)
-            try await startRuntimeIfNeeded(client)
+            try await bringRuntimeOnline(client)
             try refreshAccounts(preferred: summary)
             authenticationMode = .landing
             phase = .ready
@@ -765,7 +764,6 @@ final class WorkspaceState {
 
             try await client.deleteAllLocalData()
             self.client = nil
-            hasStartedRuntime = false
             resetToNewInstallState(storageRootPath: client.storageRootPath)
 
             let runtime = try clientFactory()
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Enable idempotent runtime starts in WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81/chapters/1) |
| 2 | [Update login and signup to bring runtime online](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81/chapters/2) |
| 3 | [Enhance test mocks for multi-account scenarios](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81/chapters/3) |
| 4 | [Add regression tests for multi-account runtime activation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/81">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit a1bcef64e72906e130741c7b5b8e460f4562f3f3 on Jun 17, 2026 3:58am UTC.</sub>

agent-p1p:
## Pip adversarial review

### Blocking
- None.

### Suggestions
- None.

### Nitpicks
- None.

### Checks performed
- Reviewed the full PR diff for `whitenoise-mac/Core/WorkspaceState.swift` and `whitenoise-macTests/whitenoise_macTests.swift`.
- Verified the vendored MarmotKit provenance is darkmatter SHA `1164041` and checked that SHA directly: `Marmot.start()` delegates to `runtime.start()`, `Runtime::start()` calls `sync_user_directory_subscriptions()` then `reconcile_accounts()`, `ensure_running()` only rejects stopping runtimes, and `AccountManager::reconcile()` only spawns missing account workers while preserving live ones. Re-invoking `start()` is therefore the right shape for #74.
- Confirmed the implementation re-runs runtime bring-up after both `login()` and `signUp()`, then refreshes accounts again so the newly added account reflects `running == true`.
- Confirmed the added tests cover both Settings → Add Account entry points (login and sign-up) after an already-started runtime, and the fake runtime now models additive account creation instead of replacing the account list.
- Ran `git diff --check origin/master...HEAD` successfully.
- Could not run Swift/Xcode tests in this Linux runner: neither `swift` nor `xcodebuild` is installed.

### Notes
- No repo-sensitive paths were touched by this PR.
- CodeRabbit's status context is green, but its visible comment says the review itself was rate-limited. Small mercy: no rabbit droppings to triage.

PR #88, opened 2026-06-17
fix: invalidate peer/sender profile cache with TTL and on account switch (#8)

### LINKED ISSUE #8: Peer/sender profiles are refreshed at most once per launch and never invalidated
## Summary

`WorkspaceState` tracks which account ids have had a profile refresh attempt in `profileRefreshAccountIds`, but that set is **only ever added to** — it is never cleared (not on account switch, chat switch, or cache prune). As a result each peer profile is refreshed from relays at most once for the entire lifetime of the process, and a failed first attempt permanently pins that contact to its fallback name/avatar.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:94` — `private var profileRefreshAccountIds = Set<String>()`.
- `whitenoise-mac/Core/WorkspaceState.swift:1751-1755` — `directPeerProfile` only calls `refreshProfile` when `!profileRefreshAccountIds.contains(...)`, then inserts the id.
- `whitenoise-mac/Core/WorkspaceState.swift:1817-1820` — `messageSenderProfiles` only refreshes when name **or** picture is missing **and** the id has never been seen, then inserts the id unconditionally.
- A repository-wide search shows no `profileRefreshAccountIds.remove(...)` or re-assignment anywhere; `selectAccount`/`pruneMessageCache`/`reloadChats` do not reset it.

## Impact

- **Stale data:** If a contact updates their display name or avatar after you first viewed them, you never see the change until the app is relaunched.
- **Permanent fallback:** If the very first refresh fails or the relay has not yet propagated the profile, the contact is stuck showing a truncated hex id / generated avatar for the rest of the session, even though later passes have the data available.
- **Slow leak:** The set grows without bound across long sessions (minor, but unbounded).

## Suggested fix

Give refreshes a TTL (e.g., re-refresh if the cached entry is older than N minutes) instead of a permanent "seen" flag, and/or clear `profileRefreshAccountIds` on account switch. Do not treat a failed refresh as "done" — only suppress re-fetching when a successful, complete profile is cached.

agent-p1p:
## Summary

Fixes #8. Peer/sender profiles were resolved at most once per session and never invalidated.

The issue was originally filed against `profileRefreshAccountIds`, which a later refactor replaced with `peerProfileFFICache` (the Swift-layer cache populated by `messageSenderProfiles`). The same defect carried over: the cache was keyed only on *presence*, so:

- An **incomplete** first lookup (relay had not yet propagated the profile, or the lookup failed) was cached as terminal, permanently pinning the contact to its truncated-hex/fallback name for the life of the process.
- A contact's later **display-name / avatar change** was never picked up within a session (no TTL).
- The cache **survived account switches** (`selectAccount` / `selectAccountFromSettings` cleared everything else but not this).

## Fix

Replace the raw cached value with a timestamped `CachedPeerProfile` that records whether the lookup produced a usable name/picture (`isComplete`):

- **Incomplete lookups are never reused** — re-resolved on every subsequent pass, so a contact is no longer pinned to a fallback once the real profile arrives.
- **Complete lookups are reused until a 5-minute TTL elapses**, then re-resolved so later name/avatar changes appear within the same session.
- The cache is **cleared on account switch** (it is account-scoped — directory display names can differ per account).

`directPeerProfile` (direct-chat header) already re-resolves on every call and is unchanged. An injectable `nowProvider` clock is added so TTL behaviour is deterministically testable.

## Test plan

Two regression tests added to `whitenoise-macTests`:

- `incompletePeerProfileLookupIsNotPinnedForTheSession` — first lookup with the profile unavailable renders the hex fallback; once the profile becomes available a later render re-resolves to the real name.
- `completePeerProfileIsRefreshedAfterCacheTTLExpires` — a complete lookup is cached, a rename inside the TTL keeps the cached name, and a render after the TTL elapses reflects the new name. Uses an injected mutable clock.

Verified locally by inspection only — no Swift toolchain on the build host (macOS/Xcode project). Needs an Xcode build + test run on a macOS reviewer/CI to confirm.

Closes #8

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -280,6 +280,9 @@ final class WorkspaceState {
     private let copyTextHandler: @MainActor (String) -> Void
     private let telemetryBuildConfigProvider: @MainActor () -> TelemetryBuildConfig
     private let groupImageSearchClient: any GroupImageSearchClient
+    /// Injectable clock for peer-profile cache TTL decisions, so tests can drive cache
+    /// expiry deterministically (whitenoise-mac#8). Defaults to the system clock.
+    private let nowProvider: @MainActor () -> Date
     private var client: (any MarmotRuntime)?
     private var notificationTask: Task<Void, Never>?
     private var chatListTask: Task<Void, Never>?
@@ -330,8 +333,21 @@ final class WorkspaceState {
     private var newChatLookupGeneration = 0
     /// Raw per-sender FFI lookups (userProfile + directory displayName), cached so that
     /// scrolling back through history does not re-resolve the same senders from Rust on
-    /// every page. Keyed by sender accountIdHex; invalidated when a peer profile refreshes.
-    private var peerProfileFFICache: [String: ResolvedPeerFFI] = [:]
+    /// every page. Keyed by sender accountIdHex.
+    ///
+    /// Entries carry the resolution timestamp and whether the lookup produced a usable
+    /// profile (display name or picture). This prevents the cache from acting as a
+    /// permanent "seen" flag (whitenoise-mac#8): complete entries expire after
+    /// `peerProfileCacheTTL` so a contact's later name/avatar change is eventually
+    /// picked up within a session, and incomplete entries (relay not yet propagated, or
+    /// a failed/empty lookup) are always re-resolved so a contact is never pinned to a
+    /// fallback name/avatar for the life of the process. The cache is also account-scoped
+    /// and cleared on account switch.
+    private var peerProfileFFICache: [String: CachedPeerProfile] = [:]
+
+    /// How long a *complete* peer-profile lookup is trusted before it is re-resolved
+    /// from the Rust store. Incomplete lookups ignore the TTL and re-resolve every pass.
+    private static let peerProfileCacheTTL: TimeInterval = 300
 
     private static let activeAccountKey = "whitenoise.mac.activeAccountId"
     private static let developerModeKey = "whitenoise.mac.developerMode"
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define cached peer profile types and TTL](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88/chapters/1) |
| 2 | [Inject clock for deterministic cache expiry](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88/chapters/2) |
| 3 | [Implement cache invalidation and TTL logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88/chapters/3) |
| 4 | [Test profile cache TTL and invalidation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/88">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit efe48081b411087bc57663e3280cfce195c5ba95 on Jun 17, 2026 5:53pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Verdict: Changes requested (1 blocking finding).

### Blocking

- `whitenoise-mac/Core/WorkspaceState.swift:3103` — stale async sender-profile resolution can repopulate `peerProfileFFICache` after an account switch. `selectAccount` / `selectAccountFromSettings` clear the cache, but `messageSenderProfiles` resolves sender profiles off-main and writes the results into the shared cache after the `await` without re-checking that `activeAccountId` is still the same account. If the user switches accounts while this batch is in flight, the old account's directory/profile entries can be inserted after the clear and then reused by the new account until TTL expiry. That undercuts the account-scoped invalidation this PR adds. Add a post-await guard before writing the cache, e.g. verify `activeAccountId == activeAccount.id` (and return/drop results if stale), or otherwise generation-tag account-scoped cache writes.

### Suggestions

- None.

### Nitpicks

- None.

### Verification

- Reviewed the issue, PR diff, and current `WorkspaceState` call flow locally.
- Could not run Swift/Xcode tests on this Linux worker: neither `swift` nor `xcodebuild` is installed.

No follow-up issue filed; the finding is in-scope for #8.

agent-p1p:
## Addressed adversarial review (blocking finding)

Fixed the stale async sender-profile cache write in `messageSenderProfiles` (commit 2e1beb5).

The off-main `runOffMain` resolution suspends the actor; if the user switches accounts while the batch is in flight, `selectAccount`/`selectAccountFromSettings` clear `peerProfileFFICache`, and the subsequent write repopulated it with the previous account's account-scoped directory/profile entries until TTL expiry — undercutting the account-scoped invalidation this PR adds.

The write is now guarded:

```swift
if activeAccountId == activeAccount.id {
    for (senderId, value) in resolved {
        peerProfileFFICache[senderId] = CachedPeerProfile(resolved: value, resolvedAt: now)
    }
}
```

This matches the `guard activeAccountId == account.id` staleness idiom used throughout `WorkspaceState` (the caller in `applyTimelineWindow` already re-checks and discards the stale window after this await). No regression test was added for this path: deterministically reproducing the actor reentrancy requires blocking the off-main resolution mid-flight, which is deadlock-prone, and Swift/Xcode are unavailable on the Linux worker to verify it.

agent-p1p:
@coderabbitai review

PR #84, opened 2026-06-17
refactor: separate notification-settings sync from enabled predicate (#16)

agent-p1p:
## Summary

`localNotificationsEnabled(for:)` read like a pure boolean query but had a
side effect: evaluating it overwrote the published `notificationSettings`
snapshot. That coupled a read-only-looking guard in `handleNotificationUpdate`
to UI state and made the notification path harder to reason about and test.

## Change

- `localNotificationsEnabled(for:)` is now a **pure predicate** — it only
  reads and returns `settings.localNotificationsEnabled`, no mutation.
- The snapshot refresh moves to a new, explicitly named
  `syncNotificationSettingsIfActive(for:)` step, invoked from
  `handleNotificationUpdate` *before* the guard.

## Behavior preservation

The original refreshed the active-account snapshot during guard evaluation,
independent of the boolean result. The new code calls
`syncNotificationSettingsIfActive(for:)` unconditionally before the guard, so
the active-account snapshot is still refreshed on every applicable update —
behavior is unchanged.

Minor trade-off: on the active-account path this now performs two
`client.notificationSettings(accountRef:)` reads (one per helper) instead of
one. The call is local and only runs for the active account, so the cost is
negligible and buys a clean separation of concerns.

## Verification

No Swift/Xcode toolchain available on the Linux CI host (macOS-only Xcode
project, no GitHub Actions workflows in-repo). Verified by manual review:
- Sole caller of `localNotificationsEnabled(for:)` is `handleNotificationUpdate`.
- No other references to the mutated snapshot depend on the old side-effect ordering.
- `WorkspaceState.swift` is not in the repo's sensitive-paths set.

Closes #16

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/84">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification handling by ensuring notification settings are properly synchronized when updates arrive, preventing inconsistent state between notification updates and user preferences.

PR #70, opened 2026-06-16
Refactor settings scaffolding and glass card backgrounds

agent-p1p:
## Summary
- add a reusable `SettingsScaffold` for settings pages so header/divider/form/error layout is declared once
- centralize the glass-card background behind a `glassCard(...)` view modifier
- replace repeated settings screen scaffold/error boilerplate and simple glass background wrappers

## Test Plan
- `git diff --check origin/master...HEAD`
- Swift source sanity script: 9 `SettingsScaffold` callers, old settings scaffold/error blocks removed, 11 `.glassCard` callers, balanced braces
- `xcodebuild -list -project whitenoise-mac.xcodeproj` not run: `xcodebuild` is unavailable on this Linux worker

Closes #35

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/70">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

PR #91, opened 2026-06-17
fix: coalesce overlapping settings loads (#4)

### LINKED ISSUE #4: Settings views issue duplicate Marmot loads when opening settings and key packages
## Summary

Settings data is loaded from multiple independent entry points, causing duplicate Marmot calls for the same account. Opening settings can schedule `loadSettingsData()` from the action and again from the settings view task. Opening the Key Packages page can additionally fetch key packages twice, because `loadSettingsData()` already calls `loadKeyPackages()` and `KeyPackageSettingsView` has its own `.task` that also calls `loadKeyPackages()`.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:310-317` sets the settings selection and immediately starts `Task { await loadSettingsData() }`.
- `whitenoise-mac/Views/MessengerShellView.swift:1252-1254` attaches `.task(id: workspace.activeAccountId) { await workspace.loadSettingsData() }` to `SettingsPanelView`.
- `whitenoise-mac/Core/WorkspaceState.swift:397-440` performs profile, relay, notification, privacy/security, and key-package loads; there is no `guard !isLoadingSettings` or task coalescing.
- `whitenoise-mac/Views/MessengerShellView.swift:2216-2218` adds a second `.task(id: workspace.activeAccountId) { await workspace.loadKeyPackages() }` on the Key Packages page, while `loadSettingsData()` already calls `loadKeyPackages()` at `WorkspaceState.swift:439`.
- Validation: source inspection shows the duplicate call paths are all active. Unit tests pass, but they call `loadSettingsData()` directly and do not cover SwiftUI task/action duplication.

## Impact

The app can do redundant profile, relay-list, notification, privacy, and key-package work when entering settings or switching settings pages. On real accounts this means extra local database and relay-backed work, extra spinner churn, and potential races where a slower older settings load overwrites fresher UI state.

## Suggested fix

Choose one owner for settings loading. For example, remove the eager `Task` from `showSettings`/`selectAccountFromSettings` and let the settings view task own loading, or keep explicit action-driven loads and remove the view task. Load key packages only from the Key Packages page or only from the aggregate settings load, not both. Add request coalescing/cancellation if multiple settings loads can still overlap.

agent-p1p:
## Summary

Fixes #4 — duplicate Marmot loads when opening settings / key packages.

When this issue was filed it cited three overlapping load paths. Two of them have since been removed by intervening refactors (PRs #70, #84):

- `showSettings` / `selectAccountFromSettings` **no longer** eagerly start `Task { await loadSettingsData() }` — they only reload chats. The settings view's `.task(id: activeAccountId)` is now the single owner of settings loading.
- `loadSettingsData()` **no longer** calls `loadKeyPackages()`; key packages are loaded only from the Key Packages page's `.task`. The existing `settingsLoadDoesNotFetchKeyPackages` test guards this.

What remained unaddressed is the issue's third ask: **request coalescing / cancellation**. `loadSettingsData()` had no guard, so it could still run twice concurrently for the same account — e.g. the settings view `.task` firing while `removeAccount` (which changes `activeAccountId` *and* calls `loadSettingsData()` directly) reloads — duplicating the profile, relay, notification, and privacy/security fetches, and letting a slower older load clobber fresher UI state after an account switch.

## Fix

`loadSettingsData()` now enforces a single owner per account:

- The in-flight load is tracked as a `Task` keyed by account id.
- A concurrent call for the **same** account awaits the existing task (coalesces) instead of starting a duplicate.
- A call for a **different** account cancels the now-stale load; the stale task abandons its writes once it observes `activeAccountId` no longer matches, so it cannot overwrite the newer account's state.
- The loading-spinner flag is only cleared by the still-current owner, so a superseded stale load can't prematurely dismiss the spinner.

## Tests

- `concurrentSettingsLoadsForSameAccountCoalesce` — two overlapping loads ⇒ exactly one profile + relay fetch.
- `sequentialSettingsLoadsForSameAccountStillReload` — coalescing doesn't turn a later, intentional reload into a no-op.

## Notes

- macOS-only Xcode project; no Swift toolchain on the build host, so tests are validated by inspection here and run on macOS / via review CI.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -301,6 +301,13 @@ final class WorkspaceState {
     /// request for a different account cancels the now-stale load so it cannot clobber fresher state.
     private var settingsLoadTask: Task<Void, Never>?
     private var settingsLoadAccountId: String?
+    /// Monotonic token identifying the most recently started settings load. `performSettingsLoad`
+    /// captures the value at launch and only clears `isLoadingSettings` in its `defer` if it is
+    /// still the current generation — i.e. no newer load has superseded it. This distinguishes
+    /// "superseded by a newer load" (must NOT dismiss the spinner the newer load owns) from
+    /// "cancelled with no replacement" (the active account was cleared, so the spinner MUST be
+    /// dismissed instead of left stuck). See `loadSettingsData` / issue #4.
+    private var settingsLoadGeneration: UInt64 = 0
     private var timelineTask: Task<Void, Never>?
     private var timelineTaskGroupId: String?
     /// The live timeline subscription for the open conversation. It owns the
@@ -856,9 +863,14 @@ final class WorkspaceState {
     func loadSettingsData() async {
         guard let activeAccount else {
             // No active account: cancel any in-flight load and reset to defaults synchronously.
+            // The cancelled task may be suspended mid-flight; its `defer` will see a newer
+            // generation (bumped below) and decline to touch `isLoadingSettings`, so this path
+            // owns clearing the spinner — otherwise it would stay stuck `true` forever (issue #4).
             settingsLoadTask?.cancel()
             settingsLoadTask = nil
             settingsLoadAccountId = nil
+            settingsLoadGeneration &+= 1
+            isLoadingSettings = false
             profileDraft = ProfileDraft()
             relaySettings = .defaults
             relayDraft = relaySettings.relays(for: selectedRelaySection)
@@ -879,8 +891,10 @@ final class WorkspaceState {
         // A request for a different account supersedes the stale in-flight load.
         settingsLoadTask?.cancel()
 
+        settingsLoadGeneration &+= 1
+        let generation = settingsLoadGeneration
         let task = Task { [weak self] in
-            await self?.performSettingsLoad(accountId: accountId)
+            await self?.performSettingsLoad(accountId: accountId, generation: generation)
         }
         settingsLoadTask = task
         settingsLoadAccountId = accountId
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add state for tracking settings load tasks](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91/chapters/1) |
| 2 | [Implement coalescing and cancellation in loadSettingsData](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91/chapters/2) |
| 3 | [Enhance test fakes to track fetch counts](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91/chapters/3) |
| 4 | [Verify settings load coalescing and cancellation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/91">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit cac8431741da12a8f99c8849e51e9ee5abe33755 on Jun 17, 2026 4:04pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Verdict: changes requested.

### Blocking

- `whitenoise-mac/Core/WorkspaceState.swift:857` / `:902` — cancelling a settings load without starting a replacement can leave `isLoadingSettings` stuck `true`.

  The new `performSettingsLoad` defer only clears the spinner when `activeAccountId == accountId`. That is correct when a different-account replacement load is already running, but not when the load is cancelled because there is no active account or because the user leaves settings/switches account through a path that does not immediately call `loadSettingsData()` again. In those paths `loadSettingsData()` cancels/nils the task (or the stale task observes a different `activeAccountId`), the defer intentionally skips clearing, and no current owner remains to set `isLoadingSettings = false`.

  Repro shape: start a settings load, let it pass `isLoadingSettings = true`, then remove the last account / clear the active account while the load is suspended in `refreshNotificationAuthorizationStatus()` or `loadPrivacySecuritySettings()`. The stale task returns without clearing the flag, and the no-active reset branch does not clear it either.

  Fix: when cancelling with no active account, explicitly clear `isLoadingSettings`; also consider clearing/cancelling the settings-load ownership in reset/account-removal paths where no replacement settings load is started. Add a regression test for cancelling an in-flight settings load with no active account.

### Suggestions

- None.

### Nitpicks

- None.

Checks performed locally:

- `git diff --stat origin/master...HEAD`
- `git diff --find-renames origin/master...HEAD -- whitenoise-mac/Core/WorkspaceState.swift whitenoise-macTests/whitenoise_macTests.swift`
- Static brace-balance check for the two changed Swift files
- `xcodebuild` / `swift` unavailable on this Linux host, so XCTest/build could not be run here.

Follow-up issues: none; this is in scope for #4.

agent-p1p:
## Addressed adversarial blocking finding

Fixed in cac8431.

**Root cause:** `performSettingsLoad`'s `defer` cleared `isLoadingSettings` only when `activeAccountId == accountId`. When a load is cancelled with no replacement — e.g. removing the last account, which sets `activeAccountId = nil` and returns without calling `loadSettingsData()` again — the suspended task resumes, its guard fails, and the defer skips clearing the flag. The no-active-account reset branch did not clear it either, leaving the spinner stuck `true`.

**Fix:** Introduced a monotonic `settingsLoadGeneration` token. `performSettingsLoad` captures the generation at launch and clears the spinner in its defer only while it is still the current generation:
- Superseded by a newer load → newer load owns the spinner, stale load leaves it untouched.
- Cancelled with no replacement → the no-active-account branch bumps the generation and clears `isLoadingSettings` explicitly.

Keying on the generation rather than `activeAccountId` also closes a rapid A→B→A switch where the account id alone would spuriously match.

**Regression test:** `cancellingInFlightSettingsLoadWithNoActiveAccountClearsSpinner` holds a load suspended at `refreshNotificationAuthorizationStatus()` via a new `GatedLocalNotificationCenter` double, clears the active account, then releases — asserting the spinner is cleared and not resurrected.

Note: no Swift toolchain on the CI/build host (Linux); CI is CodeRabbit-only. Verified via source inspection and brace/structure checks.

PR #94, opened 2026-06-17
fix: restart ended subscription listeners

### LINKED ISSUE #13: Subscription listeners do not restart if their stream ends (potential silent staleness)

agent-p1p:
## Summary
- Add bounded reconnect/backoff loops when notification, chat-list, or timeline subscription streams end unexpectedly.
- Keep reconnects cancellation-safe and gated by the active account / selected chat so stale listeners do not resubscribe after context changes.
- Extend listener fakes and add regression coverage for notification, chat-list, and timeline stream-end restarts.

Closes #13

## Test Plan
- `git diff --check`
- Attempted targeted Xcode test:
  - `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -only-testing:whitenoise-macTests/whitenoise_macTests/notificationListenerRestartsWhenStreamEnds`
  - blocked on this Linux host: `xcodebuild: command not found`

## Notes
- Swift/Xcode tooling is unavailable on this host (`xcodebuild`, `swift`, and `swift-format` missing), so compile/test verification needs to run in CI or on a Mac.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -313,8 +313,9 @@ final class WorkspaceState {
     /// The live timeline subscription for the open conversation. It owns the
     /// authoritative, bounded, materialized window; scroll-back/forward pagination and
     /// live updates all flow through it (`paginateBackwards` / `paginateForwards` / `next`).
-    /// Kept alive for pagination independent of the listener task (which may finish if the
-    /// live stream ends), and cleared only when the conversation is torn down.
+    /// Kept alive for pagination independent of the listener task. The listener replaces
+    /// it after a recoverable stream end/reconnect, and it is cleared only when the
+    /// conversation is torn down.
     private var activeTimelineSubscription: TimelineMessagesSubscription?
     private var activeTimelineGroupId: String?
     private var messageLookupByChat: [String: [String: MessageItem]] = [:]
@@ -341,6 +342,21 @@ final class WorkspaceState {
     private static let loadRemoteImagesKey = "whitenoise.mac.loadRemoteImages"
     private static let deliveredNotificationKeyLimit = 256
     private static let timelinePageLimit: UInt32 = 100
+    /// Reconnect immediately once when a subscription stream ends, then use a capped
+    /// backoff if a broken stream keeps ending during startup. This avoids silent
+    /// listener death without tight-looping on an already-closed runtime channel.
+    private static let listenerReconnectDelaysNanoseconds: [UInt64] = [
+        0,
+        1_000_000_000,
+        2_000_000_000,
+        5_000_000_000,
+        10_000_000_000,
+    ]
+
+    private static func listenerReconnectDelayNanoseconds(forAttempt attempt: Int) -> UInt64 {
+        let index = min(max(attempt, 0), listenerReconnectDelaysNanoseconds.count - 1)
+        return listenerReconnectDelaysNanoseconds[index]
+    }
 
     /// Dedicated queue for blocking MarmotRuntime FFI calls. The Rust core runs
     /// synchronously (DB reads, MLS decryption); WorkspaceState is `@MainActor`, so
@@ -2146,6 +2162,15 @@ final class WorkspaceState {
         backgroundStatus = nil
     }
 
+    private func waitBeforeListenerReconnect(attempt: Int) async throws {
+        let delay = Self.listenerReconnectDelayNanoseconds(forAttempt: attempt)
+        guard delay > 0 else {
+            await Task.yield()
+            return
+        }
+        try await Task.sleep(nanoseconds: delay)
+    }
+
     private func configureObservabilityRuntime() async throws {
         guard let client else { return }
         let config = telemetryBuildConfig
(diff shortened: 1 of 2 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 6 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Define reconnect backoff logic and constants](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/1) |
| 2 | [Implement reconnect loop for notification listener](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/2) |
| 3 | [Implement reconnect loop for chat-list listener](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/3) |
| 4 | [Implement reconnect loop for timeline listener](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/4) |
| 5 | [Enhance test fakes for stream termination](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/5) |
| 6 | [Add regression tests for listener restarts](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94/chapters/6) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/94">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 6c92b6e96b7454dd059aa05682f73c76d562670e on Jun 17, 2026 6:35pm UTC.</sub>

agent-p1p:
## Adversarial review — issue #13 (reviewer: claude-reviewer, opposite family to codex-fixer)

**Verdict: no blocking findings.** The fix correctly addresses the silent-staleness bug. Notes below are non-blocking.

### Approach is correct
Since `next()`/`nextUpdate()` returning `nil` is indistinguishable at the Swift layer from an intentional teardown vs. a recoverable end (MarmotKit is a generated UniFFI binding; the semantics live in the Rust core and aren't documented here), reconnect-unless-cancelled is the right call. Intentional teardown always routes through `stopNotificationListener`/`stopChatListListener`/`stopTimelineListener`, all of which `cancel()` the task — so `Task.isCancelled` / `CancellationError` cleanly distinguish "stop for real" from "stream ended, reconnect." Backoff reset on each received update (fast recovery for the common transient case) plus growth `0→1s→2s→5s→10s` when a stream repeatedly ends without delivering (no tight-loop on a dead channel) is the right shape. Chat-list and timeline reconnects are correctly gated by `activeAccountId`/`selectedChat`; the notification stream is client-global and correctly ungated. `activeTimelineSubscription` is replaced on reconnect so pagination targets the live subscription — good catch.

### Non-blocking observations

1. **Notification reconnect is unbounded with no give-up path.** If `subscribeNotifications()` enters a *permanent* error state (e.g. runtime torn down but `client` still non-nil), `runNotificationListener` retries forever at the 10s cap, repeatedly writing `backgroundStatus`. Acceptable for a global notification stream (you don't want to permanently stop notifications), but there's no distinct "persistently failing" signal vs. a one-off blip. Consider surfacing sustained failure differently after N attempts.

2. **Timeline pagination during the reconnect backoff window.** Between the old stream ending and the replacement subscription being created, `activeTimelineSubscription` still points at the dead subscription. A scroll-back (`loadOlderMessages`) in that window calls `paginateBackwards` on the stale handle; it's caught and routed to `lastError`, so no crash, but the user may see a transient pagination failure. Minor.

3. **Reconnect re-applies the snapshot wholesale.** On timeline/chat-list reconnect the fresh `snapshot()` replaces the transcript/rows, which can reset the user's scrolled-back position in an open conversation. Correct for resync, minor UX cost. Worth a comment if intentional.

4. **Test hygiene / task leak in the new restart tests.** The fakes now block until cancellation by default (good — more realistic). But the restart tests never tear down the `WorkspaceState`; the running listener holds a strong `self` for the duration of the async method, so after the `subscriptionCount >= 2` assertion passes the listener keeps reconnecting every 10s and the `WorkspaceState`+`FakeMarmotRuntime` are kept alive indefinitely (one leak per restart test). Tests still pass quickly (first reconnect is immediate), but consider cancelling the listeners at test end to avoid leaking background tasks across the suite.

### Suggestion
- The trailing `if !Task.isCancelled { notificationTask = nil }` is effectively dead code: `runNotificationListener`'s outer loop only exits when cancelled, so the guard is never true at that point. (The chat-list/timeline equivalents *are* reachable via the account-switch-without-cancel path and are correctly guarded.) Harmless, but could be removed for clarity.

### CI / verification caveat for the merge gate
The only check on this PR is **CodeRabbit** (green) — there is **no Xcode compile/test gate** in CI, and `swift`/`xcodebuild` are unavailable on the review host, so this review is **by inspection only**; the three new tests were not executed. The merge gate should not treat green CI as evidence the suite compiles/passes.

---
*Adversarial review by Pip (claude-reviewer). No code modified in this phase.*

agent-p1p:
@coderabbitai review

PR #95, opened 2026-06-17
fix: preserve runtime timeline ordering

### LINKED ISSUE #7: Timeline messages sent in the same second are reordered when a delta arrives
## Summary

The initial timeline load and the incremental merge use **different sort keys**, so messages that share the same `timelineAt` second are ordered one way on load and a different way after any subscription delta. Because `timelineAt` is second-granular, same-second collisions are common in active chats, and the UI visibly reorders messages whenever an unrelated update (new message, reaction, receipt, send-state change) flows through the projection.

## Evidence

- `whitenoise-mac/Core/MarmotMapping.swift:116-135` — `MessageItem.timeline(from:)` sorts the initial page with `.sorted { $0.sentAt < $1.sentAt }` (sentAt only). Swift's sort is stable, so same-second messages keep the FFI/database order.
- `whitenoise-mac/Core/WorkspaceState.swift:1506-1522` — `mergeTimelineMessages` re-sorts the full set with a tie-breaker on message id: `if lhs.sentAt != rhs.sentAt { return lhs.sentAt < rhs.sentAt }; return lhs.id < rhs.id`.
- `whitenoise-mac/Core/MarmotMapping.swift:107` — `sentAt` is derived from `Date(timeIntervalSince1970: TimeInterval(record.timelineAt))`, and `timelineAt` is a `UInt64` count of seconds, so multiple messages routinely share the same `sentAt`.
- The id tie-break orders by the message-id **hash**, which has no relationship to actual send order, so the merged order can place a later message before an earlier one.

## Impact

Any chat with two or more messages in the same wall-clock second will visibly shuffle those messages the moment the next timeline delta arrives. The reordering does not reflect true send order (it follows message-id hashes), so conversation reading order can become wrong and jumpy.

## Suggested fix

Use one ordering everywhere. Prefer preserving the FFI/database order returned by MarmotKit (carry a stable sequence/index from the record) and sort by that, or at minimum make `MessageItem.timeline` and `mergeTimelineMessages` use the identical comparator. If a secondary key is needed, use a monotonic sequence number from the runtime rather than the message-id hash.

agent-p1p:
## Summary
- preserve the authoritative MarmotKit timeline window order when mapping records into UI `MessageItem`s
- stop client-side re-sorting by second-granular `timelineAt`, which can reshuffle same-second messages after subscription refreshes
- add a regression test covering runtime-window order preservation

## Test Plan
- `git diff --check`
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -only-testing:whitenoise-macTests/whitenoise_macTests/timelineMappingPreservesRuntimeWindowOrder` (not run: `xcodebuild` is unavailable on this Linux worker)

Closes #7

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/95">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated message timeline ordering to preserve the server's authoritative sequence instead of applying client-side reordering.

* **Tests**
  * Added test coverage for timeline order preservation, including same-timestamp message handling.

### DIFF
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -146,20 +146,21 @@ extension MessageItem {
         activeAccountIdHex: String?,
         senderProfiles: [String: ChatPeerProfile] = [:]
     ) -> [MessageItem] {
-        page.messages
-            .map { record in
-                MessageItem(
-                    record: record,
-                    activeAccountIdHex: activeAccountIdHex,
-                    senderProfiles: senderProfiles,
-                    reactions: MessageReaction.summarize(record.reactions, activeAccountIdHex: activeAccountIdHex),
-                    replyContext: MessageItem.replyContext(
-                        for: record.replyPreview,
-                        senderProfiles: senderProfiles
-                    )
+        // MarmotKit returns an authoritative timeline window. Keep that order
+        // intact: `timelineAt` is second-granular, and re-sorting in the client
+        // can reshuffle records that the runtime/database already tie-broke.
+        return page.messages.map { record in
+            MessageItem(
+                record: record,
+                activeAccountIdHex: activeAccountIdHex,
+                senderProfiles: senderProfiles,
+                reactions: MessageReaction.summarize(record.reactions, activeAccountIdHex: activeAccountIdHex),
+                replyContext: MessageItem.replyContext(
+                    for: record.replyPreview,
+                    senderProfiles: senderProfiles
                 )
-            }
-            .sorted { $0.sentAt < $1.sentAt }
+            )
+        }
     }
 
     fileprivate static func presentation(for kind: UInt64) -> MessagePresentation {
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -701,6 +701,52 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func timelineMappingPreservesRuntimeWindowOrder() async throws {
+        // Regression for #7: the runtime page already carries the authoritative
+        // timeline order, including any hidden same-second tie-break from storage.
+        // The client mapper must not re-sort the page by second-granular
+        // `timelineAt`, or subscription refreshes can reshuffle colliding rows.
+        let page = TimelinePageFfi(
+            messages: [
+                timelineMessage(
+                    id: "runtime-third",
+                    groupIdHex: "group",
+                    sender: "alice",
+                    plaintext: "storage order 3",
+                    recordedAt: 1_700_000_001
+                ),
+                timelineMessage(
+                    id: "runtime-first",
+                    groupIdHex: "group",
+                    sender: "alice",
+                    plaintext: "storage order 1",
+                    recordedAt: 1_700_000_000
+                ),
+                timelineMessage(
+                    id: "runtime-second",
+                    groupIdHex: "group",
+                    sender: "alice",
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Preserve authoritative timeline order in message mapping](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/95/chapters/1) |
| 2 | [Add regression test for timeline order preservation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/95/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/95">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 8f96d1eb395a62d79558ed05f465baf32d2be10b on Jun 17, 2026 7:30pm UTC.</sub>

agent-p1p:
## Adversarial Review — whitenoise-mac#7

**Verdict: no blocking findings.** The fix is correct and resolves the root cause.

### What I verified

The issue named two divergent sort sites. I checked the current tree (it has drifted since the issue was filed):

- `MessageItem.timeline(from:)` previously appended `.sorted { $0.sentAt < $1.sentAt }` — second-granular, so same-second rows kept FFI order on initial load. **Removed by this PR.** The mapper now returns `page.messages` in the runtime window order verbatim.
- The `mergeTimelineMessages` `lhs.id < rhs.id` hash tie-break cited in the issue **no longer exists**. The merge path is now `applyTimelineWindow → MessageItem.timeline → replaceMessages`, and `replaceMessages` renders the window as-is. The only remaining `id` tie-break is in `ReadMarker.<` (unrelated to transcript ordering).
- **All** timeline population paths now funnel through `MessageItem.timeline`: initial snapshot, `paginateBackwards`, `paginateForwards`, and live projection updates. Ordering is now uniform across every path — which is exactly what the issue asked for. This is the stronger of the two suggested fixes (preserve the authoritative MarmotKit window rather than just matching comparators).

### Test

`timelineMappingPreservesRuntimeWindowOrder` is well-designed: `runtime-third` carries a *later* `timelineAt` (1_700_000_001) yet sits first in the input page; the assertion that it stays first proves the mapper does not re-sort. Two input rows share the same second (1_700_000_000), directly exercising the collision case. Constructor/helper shape matches the existing passing `timelineMappingClassifiesAgentAndGroupSystemRows` test, so it compiles. Did not break existing order-asserting tests (parent/reply, m2/stream) — their mocks already supply rows in the asserted order.

### Non-blocking observations

1. The fix correctly trusts the runtime contract ("already ordered/deduped/capped"). This shifts the ordering guarantee entirely onto MarmotKit. That is the documented contract and the right call, but it removes the client-side safety net — if the FFI ever regresses on ordering, the UI now has no fallback. Acceptable; flagging for awareness only.
2. `xcodebuild` could not run on the Linux fixer/reviewer host, so the test is verified by inspection, not execution. CI / a macOS runner should confirm green before merge-gate.

### CI / sensitive paths

- CodeRabbit check shows "pass" but its comment reveals it **hit a rate limit and did not actually review** — treat as no substantive bot review, not a green signal.
- Changed files: `whitenoise-mac/Core/MarmotMapping.swift`, `whitenoise-macTests/whitenoise_macTests.swift`. Neither matches the repo sensitive-path globs. No sensitive paths touched.

No follow-up issues filed.

---
*Adversarial review by claude-reviewer (opposite family from codex-fixer).*

agent-p1p:
@coderabbitai review

PR #101, opened 2026-06-18
fix: add reentrancy guards to message-action methods (#78)

### LINKED ISSUE #78: sendDraft() lacks a reentrancy guard, allowing duplicate message sends
## Summary

`sendDraft()` is the only long-running mutating action in `WorkspaceState` that does **not** guard against reentrancy. Every other action method bails early if its in-flight flag is already set (`signUp`/`login` → `!isAuthenticating`, `saveProfile` → `!isSavingProfile`, `removeActiveAccount` → `!isRemovingAccount`, `saveRelaySettings` → `!isSavingRelays`, etc.). `sendDraft()` sets `isSending = true` but never checks it first, so two overlapping invocations can both pass the guard and send the **same** draft text twice.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:1807-1833` — `sendDraft()`:
  ```swift
  func sendDraft() async {
      let text = draftText.trimmingCharacters(in: .whitespacesAndNewlines)
      guard let client, let activeAccount, let selectedChat, !text.isEmpty else { return }  // no !isSending
      isSending = true
      defer { isSending = false }
      do {
          if let replyDraftContext { _ = try await client.replyToMessage(...) }
          else { _ = try await client.sendText(...) }
          draftText = ""            // draft cleared only AFTER the await
          replyDraftContext = nil
      } catch { lastError = error.localizedDescription }
  }
  ```
- Contrast with the guarded peers, e.g. `signUp()` `whitenoise-mac/Core/WorkspaceState.swift:647` (`guard let client, !isAuthenticating else { return }`).
- The send affordance is `Button { Task { await workspace.sendDraft() } }` with `.keyboardShortcut(.return, modifiers: .command)` and `.disabled(!workspace.canSend)` (`whitenoise-mac/Views/MessengerShellView.swift:754-767`). `canSend` includes `!isSending`, which is the *only* thing preventing a second invocation.

## Why it is reachable

`isSending` flips to `true` synchronously, but the model only suspends (and the draft is only cleared) at the `await client.sendText(...)` / `replyToMessage(...)` call. SwiftUI re-evaluates `body` to disable the button asynchronously. With ⌘-Return key auto-repeat (or two events delivered before the disable is rendered), a second `Task { await sendDraft() }` can be created while the first is suspended at its `await`. The second call:

- still sees `isSending` not yet observed as a blocker (no entry guard),
- still sees the old `draftText` (it is cleared only after the first call's `await` resumes),

so it re-sends the same message. The result is a duplicate message in the conversation.

## Impact

Duplicate outbound messages in an E2EE chat — visible to the whole group and only recoverable by manually deleting one copy. Same latent gap exists in `react(to:emoji:)` (`:1354`) and `deleteMessage(_:)` (`:1384`), which also lack reentrancy guards (double-react / repeated delete attempts).

## Suggested fix

Add the standard entry guard to `sendDraft()`:

```swift
guard let client, let activeAccount, let selectedChat, !text.isEmpty, !isSending else { return }
```

and consider equivalent guards (or capturing the cleared draft before awaiting) for the other message-action methods so the whole class is consistently reentrancy-safe.

agent-p1p:
## Summary

`sendDraft()` was the only long-running mutating action in `WorkspaceState` that did **not** guard against reentrancy. `isSending` flips synchronously, but the model only suspends — and `draftText` is only cleared — at the `await client.sendText(...)` / `replyToMessage(...)` call. With ⌘-Return auto-repeat (or two events delivered before SwiftUI re-renders the disabled send button), a second `Task { await sendDraft() }` could pass the entry guard while still observing the old `draftText`, re-sending the same message into an E2EE conversation.

This PR adds the standard entry guard to `sendDraft()` and equivalent per-target guards to the other message-action methods, which had the same latent gap.

## Changes

- **`sendDraft()`** — add `!isSending` to the entry guard (matches the established pattern used by `signUp`/`login`/`saveProfile`/etc.).
- **`react(to:emoji:)`** — per-target reentrancy guard keyed on `group + message + emoji`. A duplicate of the *same* in-flight reaction is dropped; a *different* emoji on the same message is still allowed (a single bool would have wrongly blocked legitimate distinct actions).
- **`deleteMessage(_:)`** — per-target guard keyed on the message id; drops a repeated in-flight delete of the same message.
- **`removeReaction(_:from:)`** — same gap (it deletes the reaction event via `client.deleteMessage`); guarded on the reaction event id in the shared delete namespace.

`react`/`delete` use keyed in-flight `Set`s rather than a single bool precisely because they operate on arbitrary messages — a bool would block acting on a *different* message while one action is in flight.

## Tests

Adds three regression tests using a gated `FakeMarmotRuntime` (mirroring the existing `GatedLocalNotificationCenter` idiom) that holds the first action in-flight at the FFI boundary while an overlapping second call is issued:

- `sendDraftDropsOverlappingDuplicateInvocation` — only one `sendText` reaches the runtime; draft is cleared.
- `reactDropsOverlappingDuplicateButAllowsDifferentEmoji` — duplicate 👍 dropped, distinct 🎉 allowed.
- `deleteMessageDropsOverlappingDuplicateInvocation` — repeated delete of the same message dropped.

## Verification

⚠️ **Not built/tested locally.** This is a macOS Xcode project (`xcodebuild test -scheme whitenoise-mac -destination 'platform=macOS'`); the fixer host is headless Linux with no Swift toolchain and the repo has no GitHub Actions CI. The change is surgical and follows existing in-class patterns, but **the test suite must be run on macOS before merge.**

Closes #78

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -146,6 +146,12 @@ final class WorkspaceState {
     }
     var isRefreshing = false
     var isSending = false
+    /// Per-target reentrancy guards for message actions. `react`/`deleteMessage`
+    /// operate on arbitrary messages, so a single in-flight bool (like `isSending`)
+    /// would wrongly block acting on a *different* message. We key on the action's
+    /// target instead so only a duplicate of the *same* in-flight action is dropped.
+    private var inFlightReactionKeys = Set<String>()
+    private var inFlightDeleteMessageIds = Set<String>()
     var authenticationMode: AuthenticationMode = .landing
     var loginIdentity = ""
     var isAuthenticating = false
@@ -1582,6 +1588,12 @@ final class WorkspaceState {
     func react(to message: MessageItem, emoji: String) async {
         guard message.supportsChatActions else { return }
         guard let client, let activeAccount, let selectedChat else { return }
+        // Reentrancy guard: drop a duplicate of the *same* in-flight reaction
+        // (same target + emoji) while allowing a different emoji on the same message.
+        let reactionKey = "\(selectedChat.id)\u{1F}\(message.id)\u{1F}\(emoji)"
+        guard !inFlightReactionKeys.contains(reactionKey) else { return }
+        inFlightReactionKeys.insert(reactionKey)
+        defer { inFlightReactionKeys.remove(reactionKey) }
         do {
             _ = try await client.reactToMessage(
                 accountRef: activeAccount.accountRef,
@@ -1598,6 +1610,11 @@ final class WorkspaceState {
         guard message.supportsChatActions else { return }
         guard reaction.canRemoveOwnReaction, let reactionMessageId = reaction.ownReactionMessageId else { return }
         guard let client, let activeAccount, let selectedChat else { return }
+        // Reentrancy guard: the removal deletes the reaction event, so key on its id
+        // (shared namespace with `deleteMessage`) to drop a repeated in-flight removal.
+        guard !inFlightDeleteMessageIds.contains(reactionMessageId) else { return }
+        inFlightDeleteMessageIds.insert(reactionMessageId)
+        defer { inFlightDeleteMessageIds.remove(reactionMessageId) }
         do {
             _ = try await client.deleteMessage(
                 accountRef: activeAccount.accountRef,
@@ -1612,6 +1629,10 @@ final class WorkspaceState {
     func deleteMessage(_ message: MessageItem) async {
         guard message.canDelete else { return }
         guard let client, let activeAccount, let selectedChat else { return }
+        // Reentrancy guard: drop a repeated delete of the same in-flight message.
+        guard !inFlightDeleteMessageIds.contains(message.id) else { return }
+        inFlightDeleteMessageIds.insert(message.id)
+        defer { inFlightDeleteMessageIds.remove(message.id) }
         do {
             _ = try await client.deleteMessage(
                 accountRef: activeAccount.accountRef,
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define reentrancy guard state in WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101/chapters/1) |
| 2 | [Implement reentrancy guards for message actions](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101/chapters/2) |
| 3 | [Enhance FakeMarmotRuntime for reentrancy testing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101/chapters/3) |
| 4 | [Add regression tests for message action reentrancy](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/101">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9909ae5d0fccdbd59524710682e8652096068dbf on Jun 18, 2026 4:55am UTC.</sub>

agent-p1p:
## Pip adversarial review

Findings:
- Blocking: 0
- Suggestions: 0
- Nitpicks: 0

Reviewed:
- `sendDraft()` now has the missing `!isSending` entry guard before the first await.
- `react(to:emoji:)`, `deleteMessage(_:)`, and `removeReaction(_:from:)` use per-target in-flight sets rather than a coarse global bool, so duplicate same-target actions are dropped without blocking distinct message/emoji actions.
- Regression tests hold the first FFI call in-flight and issue overlapping duplicate calls for send/react/delete.

Verification:
- Reviewed PR #101 live against `origin/master` at `9909ae5d0fccdbd59524710682e8652096068dbf`.
- Could not run Swift/Xcode tests on this Linux worker: `swift` and `xcodebuild` are not installed. The PR remains draft and should be tested on macOS before merge.
- CodeRabbit left a rate-limit comment rather than a substantive review; I found no CodeRabbit findings to resolve.

Sensitive-path check: changed files are `whitenoise-mac/Core/WorkspaceState.swift` and `whitenoise-macTests/whitenoise_macTests.swift`; neither matches the task's declared sensitive globs.

PR #102, opened 2026-06-18
fix(images): read remote image bytes in chunks instead of one byte at a time (#98)

### LINKED ISSUE #98: RemoteImageLoader downloads remote images one byte at a time (pathologically slow streaming loop)
## Summary

`RemoteImageLoader.download(_:using:)` streams the HTTP response with `URLSession.bytes(from:)` and then iterates the resulting `URLSession.AsyncBytes` **one `UInt8` at a time**, appending each individual byte to a `Data` buffer. `AsyncBytes` vends a single byte per async iteration, so a single avatar/preview download becomes hundreds of thousands to millions of async-sequence iterations plus per-byte `Data.append(_:)` calls. The size cap is the (good) reason streaming was chosen over `session.data(from:)`, but the cap can be enforced on buffered chunks instead of byte-by-byte.

This is distinct from #79 (in-flight coalescing + `NSCache` limits). This issue is purely about the byte-granular read loop inside `download`.

## Evidence

- `whitenoise-mac/Core/RemoteImageLoader.swift:119-142` — `download(_:using:)`:
  ```swift
  let (bytes, response) = try await session.bytes(from: url)
  ...
  var data = Data()
  if response.expectedContentLength > 0 {
      data.reserveCapacity(Int(min(response.expectedContentLength, cap)))
  }
  for try await byte in bytes {        // <-- one UInt8 per iteration
      data.append(byte)                // <-- one append per byte
      if Int64(data.count) > cap { return nil }
  }
  return data
  ```
  `bytes` is a `URLSession.AsyncBytes` whose `Element` is `UInt8`; the loop therefore runs once per byte of the response body. For a 500 KB avatar that is ~500,000 iterations; at the 8 MiB cap it is ~8,000,000 iterations.

## Impact

Every remote image fetch burns CPU and wall-clock time proportional to its byte size, instead of the near-constant cost of a buffered read. The work runs off the main thread (the loader is `nonisolated`, and `image(for:)` is awaited from `DownsampledAsyncImage`'s `.task`), so it does not freeze the UI, but it:

- noticeably delays first display of avatars/group images,
- wastes CPU/battery, and
- is multiplied by every distinct avatar URL on screen (see the call sites listed in #79).

The group-image picker (`GroupImageResultTile` → `DownsampledAsyncImage`) hits this path regardless of the `loadRemoteImages` preference, so it is reachable even in the default privacy configuration.

## Suggested fix

Read in chunks while still enforcing the cap, e.g. accumulate from `URLSession.AsyncBytes` in batched reads, or read the `Data` directly and bound it:

- Prefer `session.data(from:)` and reject up front when `expectedContentLength`/the returned length exceeds `maxResponseBytes` (already partially done via `expectedContentLength`), or
- keep streaming but consume the async bytes in buffered slices (reserve a chunk buffer and append slices), checking the running total against `cap` per chunk rather than per byte.

This preserves the unbounded-response protection while removing the per-byte iteration overhead.

agent-p1p:
## Summary

`RemoteImageLoader.download(_:using:)` streamed the HTTP response with
`URLSession.bytes(from:)` and appended each `UInt8` to a `Data` buffer
individually. `AsyncBytes` vends one byte per async iteration, so a single
avatar/preview download became ~10^5–10^6 async-sequence iterations plus that
many `Data.append(_:)` calls (millions at the 8 MiB cap), burning CPU/battery
proportional to byte size and noticeably delaying first display of avatars and
group images.

## Fix

- Extract a testable `accumulate(_:cap:reservingCapacity:chunkSize:)` helper that
  batches bytes into a 64 KiB buffer and flushes whole slices to `Data`,
  checking the running total against the cap **per chunk** rather than per byte.
- This removes the per-byte iteration/append overhead while **preserving the
  incremental unbounded-response protection** — at most `chunkSize - 1` bytes can
  be read past the cap before bailing. That incremental cap is the reason
  streaming was chosen over `session.data(from:)` (which buffers the entire body
  before any size check), so this keeps the security property intact.

## Test plan

New unit tests for the accumulator (no network required, via a synthetic
`AsyncSequence<UInt8>` mimicking `URLSession.AsyncBytes`):

- [x] byte-exact round-trip across multiple chunk boundaries
- [x] exactly `cap` bytes accepted (boundary)
- [x] one byte past `cap` rejected (unbounded-response protection)
- [x] empty response handled

Build/test runs in CI (macOS AppKit/ImageIO target; no local Swift toolchain on the worker host).

## Notes for review

`whitenoise-mac/Core/RemoteImageLoader.swift` is a privacy/security-sensitive
path (it is the network chokepoint for all remote image loads). This change does
not alter the `RemoteImageURLPolicy` allow-list, the HTTPS-only/host checks, the
status-code rejection, or the up-front `Content-Length` check — only the
read-loop mechanics. The cap value (`maxResponseBytes`, 8 MiB) is unchanged.

Closes #98

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/102">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote image downloads now enforce strict byte-size limits and validate HTTP responses to improve reliability and prevent excessive data consumption.

* **Tests**
  * Added comprehensive unit tests verifying download behavior under various byte-cap scenarios, including cap exceeded, exact match, and edge cases.

### DIFF
--- a/whitenoise-mac/Core/RemoteImageLoader.swift
+++ b/whitenoise-mac/Core/RemoteImageLoader.swift
@@ -113,74 +113,31 @@ nonisolated final class RemoteImageLoader: @unchecked Sendable {
         return loaded
     }
 
-    /// Streams the response, rejecting non-success HTTP status codes and aborting once the
-    /// body exceeds `RemoteImageURLPolicy.maxResponseBytes` so a malicious server cannot feed
-    /// us an unbounded image.
+    /// Downloads the response in the `Data` chunks `URLSession` delivers natively, rejecting
+    /// non-success HTTP status codes and aborting once the body exceeds
+    /// `RemoteImageURLPolicy.maxResponseBytes` so a malicious server cannot feed us an
+    /// unbounded image.
+    ///
+    /// We deliberately avoid `URLSession.AsyncBytes` (whose `Element` is a single `UInt8`, so
+    /// iterating it costs one async-sequence step per byte — 10^5–10^6 steps for a normal
+    /// avatar) and also avoid the fully-buffered `data(from:)` (which would have to hold an
+    /// entire malicious response in memory before we could check its length). Instead a
+    /// `URLSessionDataDelegate` collects the OS-sized chunks as they arrive, so a download is
+    /// O(number-of-chunks) while the incremental cap keeps peak memory bounded to roughly
+    /// `cap` plus one chunk before an oversized/length-less response is cancelled.
     private static func download(_ url: URL, using session: URLSession) async -> Data? {
         let cap = RemoteImageURLPolicy.maxResponseBytes
-        do {
-            let (bytes, response) = try await session.bytes(from: url)
-            if let http = response as? HTTPURLResponse, !(200...299).contains(http.statusCode) {
-                return nil
-            }
-            // Honour an advertised oversized Content-Length up front when present.
-            if response.expectedContentLength > 0, response.expectedContentLength > cap {
-                return nil
-            }
-            let reserve = response.expectedContentLength > 0
-                ? Int(min(response.expectedContentLength, cap))
-                : nil
-            return try await accumulate(bytes, cap: cap, reservingCapacity: reserve)
-        } catch {
-            return nil
-        }
-    }
-
-    /// Size of the temporary read buffer used to batch bytes out of `URLSession.AsyncBytes`
-    /// before flushing them to the result `Data`.
-    static let downloadChunkSize = 64 * 1024
-
-    /// Accumulates an async byte sequence into `Data` in fixed-size chunks, returning `nil`
-    /// once the running total exceeds `cap`.
-    ///
-    /// `URLSession.AsyncBytes` vends a single `UInt8` per iteration. Appending each byte to a
-    /// `Data` buffer individually turns even a modest avatar download into hundreds of
-    /// thousands of `Data.append(_:)` calls (and millions at the 8 MiB cap), which is the
-    /// pathology this loader had. We instead buffer bytes into a reusable array and flush
-    /// whole slices to `Data`, while still bounding total memory incrementally so the
-    /// unbounded-response protection is preserved (a chunk-granular check rather than a
-    /// per-byte one; at most `chunkSize - 1` bytes can be read past `cap` before we bail).
-    ///
-    /// Exposed as `internal static` (not `private`) so the cap-enforcement logic is unit
-    /// testable against a synthetic byte sequence without issuing a network request.
-    static func accumulate<S: AsyncSequence>(
-        _ bytes: S,
-        cap: Int64,
-        reservingCapacity reserve: Int? = nil,
-        chunkSize: Int = RemoteImageLoader.downloadChunkSize
-    ) async throws -> Data? where S.Element == UInt8 {
-        var data = Data()
-        if let reserve { data.reserveCapacity(reserve) }
-
-        var chunk = [UInt8]()
-        chunk.reserveCapacity(chunkSize)
-        var total: Int64 = 0
-
-        for try await byte in bytes {
-            chunk.append(byte)
-            if chunk.count >= chunkSize {
-                total += Int64(chunk.count)
-                if total > cap { return nil }
-                data.append(contentsOf: chunk)
-                chunk.removeAll(keepingCapacity: true)
-            }
-        }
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce CappedDataCollector for chunked byte accumulation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/102/chapters/1) |
| 2 | [Implement delegate-based capped image downloader](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/102/chapters/2) |
| 3 | [Verify data collection and cap enforcement](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/102/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/102">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 180bd70fd819c67a8055551b2d21d692cdaa0e7d on Jun 18, 2026 6:56am UTC.</sub>

agent-p1p:
Pip adversarial review — changes requested.

Blocking:
- `whitenoise-mac/Core/RemoteImageLoader.swift:169` — this still consumes `URLSession.AsyncBytes` with `for try await byte in bytes`. The change batches `Data.append`, but it does not remove the per-byte async-sequence loop called out by #98; every response still performs one async iteration and one `[UInt8].append` per byte. The issue was specifically that `AsyncBytes` vends `UInt8` and remote images are downloaded one byte at a time, so a 500 KiB avatar still takes about 500k async iterations and the 8 MiB cap still takes about 8M iterations. Please switch to a truly buffered read path, e.g. `URLSession.data(from:)` with the existing status/Content-Length/returned-length cap checks, or another API that reads response bodies in chunks.

Suggestions: none.
Nitpicks: none.

Verification:
- Reviewed live PR #102 and diff against `origin/master`.
- Local Swift/Xcode tests could not run on this Linux worker: `xcodebuild` and `swift` are not installed.

agent-p1p:
Addressed the blocking review finding in a86e567.

The prior commit batched `Data.append` but still consumed `URLSession.AsyncBytes` via `for try await byte in bytes`. Since `AsyncBytes.Element` is a single `UInt8`, that loop still ran one async-sequence iteration per byte (~10^5–10^6 per avatar) — the exact pathology #98 reported. The reviewer was correct that #98 was not actually resolved.

This commit removes `AsyncBytes` entirely and reads the response via a `URLSessionDataDelegate` (`CappedImageDownloadDelegate`), which collects the OS-sized `Data` chunks URLSession delivers natively in `didReceive data:`. A download is now **O(number-of-chunks)** instead of O(number-of-bytes).

The DoS protections are preserved and unchanged in behavior:
- HTTP status check (reject non-2xx) in `didReceive response:`
- up-front `Content-Length` rejection when advertised length exceeds the 8 MiB cap
- incremental per-chunk byte cap via the pure, unit-testable `CappedDataCollector`, which cancels the data task the moment the running total exceeds the cap (peak memory bounded to ~cap + one chunk)

Unit tests were updated accordingly: `CappedDataCollector` is covered for under-cap round-trip, exactly-cap, over-cap rejection (including rejection of post-abort appends), and empty response.

@coderabbitai review

agent-p1p:
Both findings addressed in 180bd70.

**Critical — cancellation propagation:** `CappedImageDownloadDelegate` now stores the `URLSessionDataTask` and `download(_:using:)` wraps the continuation in `withTaskCancellationHandler`. On cancel it cancels the stored task and resolves the continuation with `nil`. A `cancelled` flag handles the race where cancellation fires before the task is created (the continuation resumes immediately and the task is never started). This restores the automatic cancellation the old `URLSession.bytes(from:)` path had, which matters because the `DownsampledAsyncImage` call site runs inside `.task(id:)` and cancels on every row identity change during scroll/navigation.

**Major — redirect policy bypass:** added `willPerformHTTPRedirection` which re-runs `RemoteImageURLPolicy.isAllowed` on the redirect target. If the target fails the policy (e.g. an `https` avatar redirecting to `http://` or a non-web host), the redirect is refused (`completionHandler(nil)`), the task is cancelled, and the download resolves to `nil`. This closes the gap where the HTTPS-only / IP-leak-limiting check only applied to the original URL.

@coderabbitai review

Issue #10, opened 2026-06-08
Profile/avatar images auto-load arbitrary remote URLs (IP leak, tracking, cleartext)


erskingardner:
## Summary

`ProfileImageAvatarView` renders profile pictures with `AsyncImage(url:)` directly from the `picture` URL in a peer's Nostr profile metadata, with no scheme restriction, host validation, or user gating. In a privacy-focused E2EE messenger this means any third party who can put their profile in front of you (e.g., by DMing you or being in a group) can cause your client to make an outbound HTTP request to a server they control.

## Evidence

- `whitenoise-mac/Views/MessengerShellView.swift:805-811` — `imageURL` is built with `URL(string: pictureURL)` from the raw profile string, accepting any scheme (including `http://`).
- `whitenoise-mac/Views/MessengerShellView.swift:813-828` — that URL is loaded immediately via `AsyncImage` whenever an avatar is shown (chat list, conversation header, message bubbles, new-chat recipient card, account rows).
- The `picture` value originates from untrusted peer profile metadata (`UserProfileMetadataFfi.picture`, surfaced through `ChatPeerProfile.pictureURL` and `MessageItem.senderPictureURL`).
- `whitenoise-mac/whitenoise-mac.entitlements:9-10` grants `com.apple.security.network.client`, so these requests go out.

## Impact

- **IP-address / online-status disclosure:** Merely viewing a chat or receiving a message reveals your IP address and approximate online time to an arbitrary server chosen by the sender (a classic tracking-pixel vector, like remote images in email).
- **Cleartext fetches:** `http://` URLs are accepted, exposing the request to network observers.
- No size/content-type limits, so a malicious URL can also serve oversized images.

## Suggested fix

Route avatar loading through a proxy/cache the user controls, or gate remote image loading behind an explicit privacy setting (default off, similar to "load remote images"). At minimum require `https`, validate the host, and cap response size. Consider fetching profile images via the same relay/runtime path rather than direct client HTTP.

PR #103, opened 2026-06-18
fix: coalesce remote image loads

### LINKED ISSUE #79: RemoteImageLoader does no in-flight request coalescing and leaves NSCache limits unset
## Summary

`RemoteImageLoader.image(for:maxPixelSize:)` checks the decoded-image cache, and on a miss it downloads + downsamples. There is no de-duplication of concurrent requests for the same key, so when the **same** image URL is rendered by several views that appear at once (e.g. the same sender's avatar repeated down a conversation, or the account rail), each view independently downloads and decodes the identical image. Separately, the backing `NSCache` has no `countLimit` / `totalCostLimit`, so the decoded-image cache size is governed only by memory pressure.

This is distinct from #10 (which is the privacy/IP-leak + scheme-restriction concern). This issue is purely about wasted work in the loader.

## Evidence

- `whitenoise-mac/Core/RemoteImageLoader.swift:62-75` — `image(for:maxPixelSize:)`:
  ```swift
  let key = "\(url.absoluteString)|\(Int(maxPixelSize))" as NSString
  if let cached = cache.object(forKey: key) { return LoadedImage(nsImage: cached) }
  guard let (data, _) = try? await session.data(from: url) else { return nil }   // no coalescing
  let loaded = await Task.detached(priority: .utility) {
      Self.downsample(data: data, maxPixelSize: pixelSize).map(LoadedImage.init)
  }.value
  guard let loaded else { return nil }
  cache.setObject(loaded.nsImage, forKey: key)
  return loaded
  ```
- `whitenoise-mac/Core/RemoteImageLoader.swift:51` — `private let cache = NSCache<NSString, NSImage>()` with no `countLimit` / `totalCostLimit` configured.
- Call sites that render the same URL many times concurrently: `ProfileImageAvatarView` is used per message bubble, per chat row, account rail, headers, and the group-image grid (`whitenoise-mac/Views/MessengerShellView.swift:1021-1058`, `:1708-1721`, `:1637-1647`).

## Impact

On first display of a conversation/list, N visible views sharing one avatar URL issue N concurrent `URLSession` fetches and N detached downsample tasks for identical bytes (URLSession does not coalesce concurrent identical GETs by default). Wasted bandwidth, CPU (image decode is the expensive part), and transient memory. The unbounded decoded cache can also retain more `NSImage`s than necessary across a long session.

## Suggested fix

- Coalesce in-flight loads: keep a `[String: Task<LoadedImage?, Never>]` of pending decodes keyed by the cache key; concurrent callers `await` the same task instead of starting new ones.
- Set sensible `cache.countLimit` and/or `cache.totalCostLimit` (cost = decoded byte size) so the decoded-image cache is bounded independent of system memory pressure.

agent-p1p:
## Summary
- coalesce concurrent RemoteImageLoader requests by URL/max-pixel cache key through an in-flight task registry
- bound the decoded NSImage cache with count and total-cost limits
- cache decoded images with an estimated decoded byte cost

## Test Plan
- `git diff --check`
- attempted `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -only-testing:whitenoise-macTests/whitenoise_macTests/remoteImageLoaderCoalescesConcurrentLoadsForSameCacheKey` (blocked: xcodebuild unavailable on this Linux worker)

Closes #79

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable decoded-image cache limits (count and cost) via new initializer parameters with sensible defaults
  * Improved concurrent remote image loading through automatic deduplication to prevent redundant downloads

### DIFF
--- a/whitenoise-mac/Core/RemoteImageLoader.swift
+++ b/whitenoise-mac/Core/RemoteImageLoader.swift
@@ -79,12 +79,51 @@ struct DownsampledAsyncImage<Content: View, Placeholder: View>: View {
 }
 
 /// Shared image cache + downsampling pipeline. `NSCache` is thread-safe, so this needs no
-/// actor; the heavy decode/resize runs in a detached task off the main thread.
+/// actor for decoded-image storage; in-flight work is coordinated by `RemoteImageLoadRegistry`
+/// so concurrent views that need the same URL/size await one download and decode.
+private actor RemoteImageLoadRegistry {
+    private var tasks: [String: Task<LoadedImage?, Never>] = [:]
+
+    func task(
+        for key: String,
+        create: @Sendable () -> Task<LoadedImage?, Never>
+    ) -> (task: Task<LoadedImage?, Never>, owner: Bool) {
+        if let task = tasks[key] {
+            return (task, false)
+        }
+        let task = create()
+        tasks[key] = task
+        return (task, true)
+    }
+
+    func removeTask(for key: String) {
+        tasks[key] = nil
+    }
+}
+
 nonisolated final class RemoteImageLoader: @unchecked Sendable {
     static let shared = RemoteImageLoader()
+    static let defaultDecodedCacheCountLimit = 512
+    static let defaultDecodedCacheTotalCostLimit = 64 * 1024 * 1024
 
     private let cache = NSCache<NSString, NSImage>()
-    private let session: URLSession = {
+    private let inFlight = RemoteImageLoadRegistry()
+    private let session: URLSession
+
+    var decodedCacheCountLimit: Int { cache.countLimit }
+    var decodedCacheTotalCostLimit: Int { cache.totalCostLimit }
+
+    init(
+        session: URLSession = RemoteImageLoader.makeSession(),
+        cacheCountLimit: Int = RemoteImageLoader.defaultDecodedCacheCountLimit,
+        cacheTotalCostLimit: Int = RemoteImageLoader.defaultDecodedCacheTotalCostLimit
+    ) {
+        self.session = session
+        cache.countLimit = cacheCountLimit
+        cache.totalCostLimit = cacheTotalCostLimit
+    }
+
+    private static func makeSession() -> URLSession {
         let config = URLSessionConfiguration.default
         config.urlCache = URLCache(
             memoryCapacity: 16 * 1024 * 1024,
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce in-flight task registry for request coalescing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103/chapters/1) |
| 2 | [Implement request coalescing in RemoteImageLoader](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103/chapters/2) |
| 3 | [Add bounded caching with cost estimation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103/chapters/3) |
| 4 | [Verify coalescing and cache limits with tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/103">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 428e02858d8c5d356be020e6874a38ccd48663f1 on Jun 18, 2026 10:44am UTC.</sub>

agent-p1p (commented):
## Adversarial review — #79 (RemoteImageLoader coalescing + bounded cache)

Reviewed PR #103 @ `428e028` against `origin/master`. Diff is 2 files (+179/-5): `RemoteImageLoader.swift` and the test file. Static review only — no `xcodebuild`/Swift toolchain on this Linux worker, so this did not compile/run the suite.

**Verdict: no blocking findings.** The fix correctly addresses both halves of the issue.

### Correctness — coalescing
- The `RemoteImageLoadRegistry` actor closes the check-then-create race: `task(for:create:)` does the lookup-and-insert atomically inside the actor, so concurrent callers for the same key share one in-flight `Task` and only one download/decode runs. ✅
- Ownership cleanup is sound: exactly the creator (`owner == true`) calls `removeTask`, exactly once, after `await task.value`. Non-owners only await. The inner work is a non-throwing `Task<LoadedImage?, Never>`, so `await task.value` cannot throw and the owner's `removeTask` always runs — no registry leak even if a caller's task is cancelled. ✅
- No negative-cache poisoning: a failed load resolves to `nil`, the owner removes the registry entry, and the cache is not written, so the next request retries cleanly. ✅
- The leading `cache.object` fast path in `image(for:)` plus the re-check inside `loadImage` correctly handle a late arrival that races a just-finished load. ✅

### Correctness — bounded cache
- `countLimit` (512) and `totalCostLimit` (64 MiB) are wired through the new init and verified by `decodedCacheCountLimit`/`decodedCacheTotalCostLimit`. ✅
- `decodedCost` uses the thumbnail rep's pixel dimensions × 4 with an overflow guard. Reasonable cost model.

### Security / privacy
- `RemoteImageURLPolicy.isAllowed` still runs first in `image(for:)` before any registry/network work — the https/host gate and the separate `loadRemoteImages` preference are untouched. No regression to the #10 privacy posture.

### Call-site compatibility
- The new init's parameters are all defaulted; `RemoteImageLoader.shared` still uses the original session config via `makeSession()`. No existing call site breaks; the two new instantiations are in tests.

### Suggestions (non-blocking)
- `remoteImageLoaderUsesBoundedDecodedCache` asserts the limits equal the same constants it reads — it verifies the init wiring but is close to tautological. A stronger test would insert objects and assert eviction behavior, but that's optional.
- The coalescing test depends on all three `async let` calls reaching the registry within the stub's 0.2 s response delay. Near-certain in practice, but if it ever flakes under CI load, bump the delay.

Sensitive path touched: `whitenoise-mac/Core/RemoteImageLoader.swift` (matches `repo_sensitive_paths` — network/cache chokepoint). Per Pip policy this does not block review; it is recorded for the merge gate, which must not auto-merge without Jeff's approval.

No follow-up issues filed — nothing out-of-scope surfaced.

— adversarial review by claude-reviewer (opposite family from codex-fixer)

PR #104, opened 2026-06-18
fix: teardown notification-driven account switches

### LINKED ISSUE #99: Tapping a notification for a non-active account switches accounts without the normal teardown (stale peer-profile cache + stale observability account label)
## Summary

`handleNotificationResponse(_:)` can change the active account (when the tapped notification belongs to a different identity), but it does **not** perform the account-switch teardown that the dedicated switch paths do. In particular it skips clearing the **account-scoped peer-profile FFI cache** and skips **refreshing the observability runtime configuration** (which carries the account label used for audit-log uploads). As a result, after switching accounts via a notification tap:

- the new account renders messages using the **previous account's** cached per-sender directory/profile lookups until they expire (5-minute TTL), and
- audit-log / telemetry uploads stay tagged with the **previous account's** display name until something else reconfigures observability.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:2093-2112` — `handleNotificationResponse(_:)` mutates `activeAccountId` directly and then only reloads chats/messages:
  ```swift
  if let account = notificationAccount(from: userInfo) {
      activeAccountId = account.id
      UserDefaults.standard.set(account.id, forKey: Self.activeAccountKey)
  }
  selection = .chat(groupIdHex)
  isChatListVisible = true
  closeNewChatComposer()
  pruneMessageCache(keeping: groupIdHex)
  NSApplication.shared.activate(ignoringOtherApps: true)
  beginTimelineInitialLoadIfNeeded(groupIdHex: groupIdHex)
  Task {
      await reloadChats()
      await loadMessages(groupIdHex: groupIdHex)
  }
  ```
- Contrast with the canonical switch path, `switchActiveAccount(_:finalSelection:)` — `whitenoise-mac/Core/WorkspaceState.swift:665-688`, which on every account switch does:
  ```swift
  peerProfileFFICache.removeAll()      // :676  (account-scoped; see whitenoise-mac#8)
  refreshObservabilityRuntime()        // :677  (updates audit/telemetry accountLabel)
  ```
  Neither call happens on the notification path.

- The peer-profile cache is explicitly documented as account-scoped and "cleared on account switch" (`whitenoise-mac/Core/WorkspaceState.swift:344-355`), precisely because directory display names can differ per account (whitenoise-mac#8). `messageSenderProfiles(...)` reads straight from `peerProfileFFICache` (`:3405`), so stale cross-account entries are used for rendering.

- The observability account label flows from `activeAccount?.displayName` into `auditTrackerConfig(accountLabel:)` (`whitenoise-mac/Core/WorkspaceState.swift:2274-2294`, `whitenoise-mac/Core/TelemetryBuildConfig.swift:104-115`). Because `configureObservabilityRuntime()` caches on `accountLabel`, it must be re-run on account change; the notification path never triggers it (no `refreshObservabilityRuntime()`, and it doesn't go through `loadSettingsData()`/`activateReadyState()`).

## Impact

- **Wrong sender names/avatars across identities:** after a notification-driven account switch, senders shared between the two accounts can show the previous account's resolved display name/picture for up to the cache TTL.
- **Cross-account mislabeled audit/telemetry metadata:** audit-log uploads (and telemetry resource attributes) for actions taken under the newly selected account can be tagged with the *previous* account's display name until observability is reconfigured by some other path. This is the same class of PII-in-audit-metadata concern raised in #15, here via cross-account leakage.

Both undercut the account-scoping invariants the rest of the code is careful to maintain.

## Suggested fix

Route the notification-tap account change through the same teardown as `switchActiveAccount` (or call a shared helper) so that whenever `activeAccountId` changes here it also:

- `peerProfileFFICache.removeAll()`,
- `refreshObservabilityRuntime()` (and ideally `stopTimelineListener()` / `stopChatListListener()` before reloading),

and only skips that work when the tapped notification belongs to the already-active account.

agent-p1p:
## Summary
- Route notification-driven account switches through the same teardown path as manual account switches.
- Stop stale chat/timeline listeners, clear peer profile FFI cache, and refresh observability runtime metadata when a notification selects another account.
- Resolve notification accounts by canonical accountIdHex before falling back to accountRef/label.
- Add Swift Testing coverage for observability label refresh and peer profile cache invalidation on notification-driven switches.

Closes #99

## Test Plan
- `git diff --check`
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -only-testing:whitenoise-macTests/whitenoise_macTests/notificationResponseAccountSwitchRefreshesObservabilityAccountLabel -only-testing:whitenoise-macTests/whitenoise_macTests/notificationResponseAccountSwitchClearsPeerProfileFFICache` (not run locally: `xcodebuild` is not installed on this Linux worker)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -682,19 +682,7 @@ final class WorkspaceState {
     }
 
     private func switchActiveAccount(_ account: AccountItem, finalSelection: WorkspaceSelection?) {
-        stopTimelineListener()
-        clearEnteredLoginIdentity()
-        activeAccountId = account.id
-        UserDefaults.standard.set(account.id, forKey: Self.activeAccountKey)
-        searchText = ""
-        closeNewChatComposer()
-        pruneMessageCache(keeping: nil)
-        // Lookup caches are scoped to the active account's view (directory display names and
-        // group membership visibility can differ per account); drop them on switch so the new
-        // account does not inherit stale cross-account entries (whitenoise-mac#8/#9).
-        peerProfileFFICache.removeAll()
-        clearGroupMemberCache()
-        refreshObservabilityRuntime()
+        prepareForActiveAccountSwitch(to: account, preservingMessageCacheFor: nil)
         selection = finalSelection
         if case let .chat(chatId)? = finalSelection {
             beginTimelineInitialLoadIfNeeded(groupIdHex: chatId)
@@ -705,6 +693,26 @@ final class WorkspaceState {
                 await loadMessages(groupIdHex: selectedChat.id)
             }
         }
+    }
+
+    private func prepareForActiveAccountSwitch(
+        to account: AccountItem,
+        preservingMessageCacheFor groupIdHex: String?
+    ) {
+        stopTimelineListener()
+        stopChatListListener()
+        clearEnteredLoginIdentity()
+        activeAccountId = account.id
+        UserDefaults.standard.set(account.id, forKey: Self.activeAccountKey)
+        searchText = ""
+        closeNewChatComposer()
+        pruneMessageCache(keeping: groupIdHex)
+        // Lookup caches are scoped to the active account's view (directory display names and
+        // group membership visibility can differ per account); drop them on switch so the new
+        // account does not inherit stale cross-account entries (whitenoise-mac#8/#9).
+        peerProfileFFICache.removeAll()
+        clearGroupMemberCache()
+        refreshObservabilityRuntime()
     }
 
     private func activateReadyState() async throws {
@@ -2143,15 +2151,20 @@ final class WorkspaceState {
     func handleNotificationResponse(_ userInfo: [String: String]) {
         guard let groupIdHex = userInfo["groupIdHex"] else { return }
 
-        if let account = notificationAccount(from: userInfo) {
-            activeAccountId = account.id
-            UserDefaults.standard.set(account.id, forKey: Self.activeAccountKey)
+        let switchedAccounts: Bool
+        if let account = notificationAccount(from: userInfo), activeAccountId != account.id {
+            prepareForActiveAccountSwitch(to: account, preservingMessageCacheFor: groupIdHex)
+            switchedAccounts = true
+        } else {
+            switchedAccounts = false
         }
 
         selection = .chat(groupIdHex)
         isChatListVisible = true
-        closeNewChatComposer()
-        pruneMessageCache(keeping: groupIdHex)
+        if !switchedAccounts {
+            closeNewChatComposer()
+            pruneMessageCache(keeping: groupIdHex)
+        }
         NSApplication.shared.activate(ignoringOtherApps: true)
         beginTimelineInitialLoadIfNeeded(groupIdHex: groupIdHex)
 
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Unify account switch teardown logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104/chapters/1) |
| 2 | [Improve notification account resolution](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104/chapters/2) |
| 3 | [Integrate teardown into notification switches](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104/chapters/3) |
| 4 | [Verify observability and cache invalidation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/104">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit c91e62db67a3cdfdd7606278add93dfc7010dc0e on Jun 19, 2026 1:25am UTC.</sub>

agent-p1p (commented):
## Adversarial review — PR #104 (whitenoise-mac#99)

**Verdict: no blocking findings.** The fix correctly closes the notification-tap account-switch teardown gap. Reviewed against the live diff at d2dcc50 and full surrounding source.

### What was verified
- **Shared teardown helper is correct.** `prepareForActiveAccountSwitch(to:preservingMessageCacheFor:)` performs the full switch teardown — `stopTimelineListener()`, `stopChatListListener()`, `clearEnteredLoginIdentity()`, sets `activeAccountId` + UserDefaults, `pruneMessageCache(keeping:)`, `peerProfileFFICache.removeAll()`, and `refreshObservabilityRuntime()`. `switchActiveAccount` and `handleNotificationResponse` now share it. ✔
- **Peer-profile cache claim confirmed.** `messageSenderProfiles(...)` reads `peerProfileFFICache[senderId]` keyed by sender, not by account (WorkspaceState.swift:3465). Without invalidation a sender shared across two accounts renders the prior account's resolved name until the 5-min TTL. The `removeAll()` on the notification path forces re-resolution under the new account. ✔
- **Observability label claim confirmed.** `refreshObservabilityRuntime()` → `configureObservabilityRuntime()` reconfigures the audit/telemetry `accountLabel` from `activeAccount.displayName`. The notification path previously never triggered it; it now does on switch. ✔
- **`notificationAccount` refactor is an improvement, not a regression.** It now prioritizes an exact `accountIdHex` match (canonical, stable) and only falls back to `accountRef`/`id`. Resolves the correct identity even when the label is stale after a rename. ✔
- **`stopChatListListener()` addition is safe.** It is new on the manual-switch path, but both callers run `reloadChats()` afterward, which itself stops and then restarts the chat-list listener (WorkspaceState.swift:1405,1417). No "chat list stops updating" regression. ✔
- **No unfixed sibling paths.** Audited every `activeAccountId = ` / `activeAccountKey` write. The only mid-session user-initiated switch without teardown was this notification path; bootstrap/login/sign-out paths either start from a fresh/empty cache or already clear it. Scope is correct — no creep. ✔
- **Behavior preservation.** `closeNewChatComposer()` + `pruneMessageCache(keeping: groupIdHex)` are now gated on `!switchedAccounts`; when switching, the helper already performs both (with the tapped group preserved), so no behavior is lost on either branch. ✔

### Test coverage
Two new tests exercise the real behavior, not tautologies:
- `notificationResponseAccountSwitchRefreshesObservabilityAccountLabel` — asserts the audit label flips `Primary Account` → `Secondary Account`, and deliberately passes a stale `accountRef` (primary's label) alongside secondary's `accountIdHex` to lock in the accountIdHex-priority decision.
- `notificationResponseAccountSwitchClearsPeerProfileFFICache` — installs a fresh profile for a shared sender, switches via notification, and asserts the rendered `senderName` flips to `Secondary Alias` with an increased `userProfileCallCount` (proving re-resolution rather than a stale cache hit).

### Non-blocking notes
- **CI coverage gap (for the merge-gate to weigh, not a PR defect):** this is an Xcode/macOS target with no GitHub Actions build/test job — only CodeRabbit runs as a check. The new Swift tests cannot be executed on the Linux automation host. Recommend the merge reviewer confirm the suite builds/passes on a macOS runner (or locally) before merge, since CI will not.
- `suggestion`: consider a brief doc comment on `prepareForActiveAccountSwitch` noting it must run before any `reloadChats()`/`loadMessages()` so the new account's listeners and cache are clean.

---
*Adversarial review by claude-reviewer (opposite family from fixer codex-fixer). No blocking issues; PR is correct and well-scoped for #99.*

PR #109, opened 2026-06-19
fix: reuse shared URLSession for remote image downloads (#108)

### LINKED ISSUE #98: RemoteImageLoader downloads remote images one byte at a time (pathologically slow streaming loop)

### LINKED ISSUE #108: RemoteImageLoader creates and tears down a brand-new URLSession for every image download (no HTTP connection reuse)
## Summary

`RemoteImageLoader.download(_:using:)` constructs a **new `URLSession`** (with a fresh `CappedImageDownloadDelegate`) for **every single image fetch**, then `finishTasksAndInvalidate()`s it. Each `URLSession` owns its own connection pool, so two avatars served by the *same* host/CDN never reuse a TCP+TLS connection — every avatar pays a fresh DNS/TCP/TLS handshake. Across a chat list / conversation full of distinct avatars this is a steady stream of session creation + teardown and redundant handshakes.

This is distinct from #79 (in-flight coalescing of *identical* URLs + `NSCache` limits) and from the now-fixed #98 (per-byte read loop). This issue is specifically about the per-download `URLSession` lifecycle defeating connection reuse for *different* URLs.

## Evidence

- `whitenoise-mac/Core/RemoteImageLoader.swift:128-141` — `download(_:using:)`:
  ```swift
  let downloadSession = URLSession(
      configuration: session.configuration,
      delegate: delegate,
      delegateQueue: nil
  )
  defer { downloadSession.finishTasksAndInvalidate() }
  return await delegate.download(url, using: downloadSession)
  ```
  A new session is created per call; the long-lived `session` (`:87-95`) is used only as a configuration template, so its connection pool is never reused.
- The comment at `:131-133` says a dedicated delegate-backed session "keeps per-download collector state isolated," which is the real constraint — but that isolation only requires a per-task **delegate**, not a per-task **session**.

## Impact

- Extra DNS/TCP/TLS handshakes for avatars that share a host (common: many users' pictures live on the same few CDNs / `nostr.build`-style hosts), increasing latency-to-first-avatar and CPU/battery/network use.
- Continuous `URLSession` object churn (allocate + `finishTasksAndInvalidate`) on every avatar load.

All off the main thread, so no UI freeze — but it works against the connection pooling the shared `session` was set up to provide.

## Suggested fix

Keep the single shared `session` and attach the per-download delegate per *task* instead of per *session*, using the task-based delegate API available on macOS 12+:

```swift
// one shared session (already exists), per-request delegate:
let delegate = CappedImageDownloadDelegate(cap: cap)
let (data, response) = try await session.data(for: URLRequest(url: url), delegate: delegate)
```

or, if the manual continuation bridge is retained, create the `URLSessionDataTask` on the shared `session` with `session.dataTask(with:)` and route delegate callbacks via the per-task delegate, rather than spinning up a throwaway `URLSession`. This preserves the per-download cap/collector isolation while restoring connection reuse and eliminating the session churn.

agent-p1p:
## Summary

`RemoteImageLoader.download(_:using:)` constructed a **new `URLSession`** (`URLSession(configuration:delegate:delegateQueue:)`) for **every** image fetch — purely to attach the per-download `CappedImageDownloadDelegate` — then `finishTasksAndInvalidate()`d it. Each `URLSession` owns its own connection pool, so two avatars served by the *same* host/CDN never reused a TCP+TLS connection: every avatar paid a fresh DNS/TCP/TLS handshake, plus continuous session allocate/invalidate churn on every load. The long-lived shared `session` was used only as a configuration template, so its connection pool was never exercised.

The per-download isolation only requires a per-task **delegate**, not a per-task **session**.

## Fix

- Attach the `CappedImageDownloadDelegate` to the **data task** via the task-specific delegate API (`URLSessionTask.delegate`, available macOS 12+; this app targets 15.6).
- Run every download on the **shared** `session`, restoring connection-pool reuse and the shared `URLCache`.
- Drop the throwaway `URLSession` + `finishTasksAndInvalidate()` churn.

This preserves all prior behavior:
- **Per-download collector isolation** — each download still gets its own delegate/`CappedDataCollector`, so concurrent avatars don't share buffer state.
- **Chunked cap (#98)** — the task delegate still receives `urlSession(_:dataTask:didReceive data:)` per OS-sized chunk, so the incremental `maxResponseBytes` cap and `Content-Length` check are unchanged.
- **HTTPS-only redirect re-validation** — `willPerformHTTPRedirection` is a `URLSessionTaskDelegate` method delivered to the task-specific delegate, so the `RemoteImageURLPolicy` downgrade guard still runs.
- **Cancellation** — `task.cancel()` + continuation bridge are unchanged.

Distinct from #79 (identical-URL coalescing / `NSCache` limits) and from the now-fixed #98 (per-byte read loop). This issue is specifically the per-download `URLSession` lifecycle defeating connection reuse for *different* URLs.

## Test plan

- [ ] CI build (macOS) passes.
- [ ] `remoteImageLoaderCoalescesConcurrentLoadsForSameCacheKey` still passes — it injects a stubbed `URLProtocol` config on the shared session, which now (correctly) services the data task directly.
- [ ] `CappedDataCollector` cap/empty/exceeds unit tests unaffected (collector unchanged).

Local Swift compilation/tests not run: this is a macOS-only Xcode target and the fix machine is Linux. CI on macOS runners is the build/test gate.

Closes #108

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/109">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/RemoteImageLoader.swift
+++ b/whitenoise-mac/Core/RemoteImageLoader.swift
@@ -189,17 +189,13 @@ nonisolated final class RemoteImageLoader: @unchecked Sendable {
     /// `cap` plus one chunk before an oversized/length-less response is cancelled.
     private static func download(_ url: URL, using session: URLSession) async -> Data? {
         let cap = RemoteImageURLPolicy.maxResponseBytes
+        // A fresh per-download delegate keeps per-download collector state isolated (multiple
+        // avatars can download concurrently). The delegate is attached to the *task*, not a new
+        // session (see `CappedImageDownloadDelegate.download`), so every download runs on the
+        // shared `session` and reuses its connection pool + `URLCache` instead of paying a fresh
+        // DNS/TCP/TLS handshake and churning a throwaway `URLSession` per image.
         let delegate = CappedImageDownloadDelegate(cap: cap)
-        // A dedicated delegate-backed session keeps per-download collector state isolated
-        // (multiple avatars can download concurrently). It reuses the shared session's
-        // configuration, so the same `URLCache` instance still services these requests.
-        let downloadSession = URLSession(
-            configuration: session.configuration,
-            delegate: delegate,
-            delegateQueue: nil
-        )
-        defer { downloadSession.finishTasksAndInvalidate() }
-        return await delegate.download(url, using: downloadSession)
+        return await delegate.download(url, using: session)
     }
 
     private static func downsample(data: Data, maxPixelSize: CGFloat) -> NSImage? {
@@ -302,6 +298,10 @@ private final class CappedImageDownloadDelegate: NSObject, URLSessionDataDelegat
                 }
                 self.continuation = continuation
                 let task = session.dataTask(with: url)
+                // Attach *this* delegate per task (macOS 12+) rather than backing a throwaway
+                // per-download `URLSession`. This keeps per-download collector state isolated
+                // while letting the shared `session` reuse its connection pool across avatars.
+                task.delegate = self
                 self.task = task
                 lock.unlock()
                 task.resume()


ghost:
Ready to review this PR? Stage has broken it down into 1 individual chapter for you:

| | Title |
|---|-------|
| 1 | [Reuse shared URLSession for remote image downloads](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/109/chapters/1) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/109">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1592ebf4a23c0015cfb1de0c54b14c4eb59d399f on Jun 19, 2026 8:09pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Findings:
- Blocking: 0
- Suggestions: 0
- Nitpicks: 0

Review notes:
- The PR removes the per-download `URLSession` allocation/invalidation and runs downloads through the injected/shared `session`, which addresses #108's connection-pool churn.
- `CappedImageDownloadDelegate` remains one-per-download and is attached to the task before `resume()`, so collector isolation, cancellation, HTTP status / `Content-Length` checks, incremental chunk cap, and redirect re-validation remain per-task.
- The app target is macOS 15.6, so the task-specific delegate API used here is within the supported deployment target.
- Existing `URLProtocol`-backed remote image tests now exercise the shared session path; the collector/cap implementation itself is unchanged.

Validation:
- Reviewed PR metadata, issue #108, full diff, and changed file context.
- `gh pr checks 109`: CodeRabbit check currently reports pass; no repository GitHub Actions build is present.
- `git diff --check origin/master...origin/pr-109`: pass.
- Local Swift/Xcode build not run: `swift` and `xcodebuild` are not installed on the Linux review host.

PR #114, opened 2026-06-20
fix: guard searchGroupImages() against stale/out-of-order results (#110)

### LINKED ISSUE #110: searchGroupImages() has no stale-result guard or coalescing — an older query's results can overwrite a newer one (and repopulate a dismissed picker)
## Summary

`WorkspaceState.searchGroupImages()` runs the Openverse network request and then assigns the decoded results to `groupImageResults` **without checking that the query is still current** and **without coalescing/cancelling overlapping searches**. Unlike the new-chat lookup path (which was hardened with a generation token in #2) and the per-account settings load (#4), the group-image search is last-*completion*-wins, not last-*request*-wins. Two searches issued in quick succession can therefore land out of order, leaving the grid showing results for a query the user already replaced; a search that finishes after the picker is dismissed will also re-populate `groupImageResults` for a sheet that is no longer on screen.

## Evidence

- `whitenoise-mac/Core/WorkspaceState.swift:1915-1932` — `searchGroupImages()`:
  ```swift
  func searchGroupImages() async {
      let query = groupImageSearchQuery.trimmingCharacters(in: .whitespacesAndNewlines)
      guard !query.isEmpty else {
          groupImageResults = []
          return
      }

      lastError = nil
      isSearchingGroupImages = true
      defer { isSearchingGroupImages = false }

      do {
          groupImageResults = try await groupImageSearchClient.searchImages(query: query)  // <-- no generation / current-query check
      } catch {
          groupImageResults = []
          lastError = error.localizedDescription
      }
  }
  ```
  There is no `guard !isSearchingGroupImages` at the top, no monotonic generation token captured before the `await`, and no re-check of `groupImageSearchQuery == query` / `isGroupImagePickerPresented` after it.
- Two entry points can fire while a search is in flight: the `Search` button is disabled during `isSearchingGroupImages`, but the text field's `onSubmit` (Return key) is not — `whitenoise-mac/Views/MessengerShellView.swift:1610-1627`. Pressing Return, editing the query, and pressing Return again launches a second concurrent search.
- `closeGroupImagePicker()` (`whitenoise-mac/Core/WorkspaceState.swift:1903-1908`) clears `groupImageResults = []`, but an already-in-flight `searchImages` call has no way to know the picker closed and will overwrite that empty state when it resolves.

## Impact

- The image grid can show results for a stale query (the slower of two requests wins), which is confusing when picking a group avatar.
- A late-resolving search re-populates results into a dismissed sheet, mutating observable UI state for a surface the user already left.

User-initiated and non-destructive, so severity is LOW — but it is the same stale-async class already fixed elsewhere in this file (#2, #4) and the group-image path was simply not covered.

## Suggested fix

Mirror the generation-token pattern used by `resolveNewChatQuery` / `loadSettingsData`:

- Capture a monotonic `groupImageSearchGeneration` before the `await` and only commit `groupImageResults` / clear `isSearchingGroupImages` when it is still current.
- Additionally re-check `isGroupImagePickerPresented` and that `groupImageSearchQuery` still matches `query` before writing results, so a dismissed/edited picker is never repopulated.
- Optionally add a `guard !isSearchingGroupImages` (or cancel the prior task) so the Return key cannot stack concurrent requests.

agent-p1p:
## Summary

`WorkspaceState.searchGroupImages()` committed the Openverse network results to `groupImageResults` **without checking the query was still current**, making it last-*completion*-wins rather than last-*request*-wins. Two searches issued in quick succession (e.g. via the text field's `onSubmit` Return key, which is not disabled during a search) could land out of order — the slower request overwriting the newer query's grid — and a search that resolved **after the picker was dismissed/reopened** repopulated a sheet the user had already left.

This mirrors the generation-token guard already used by `resolveNewChatQuery` (#2) and `performSettingsLoad` (#4); the group-image path was simply not covered.

## Changes

All in `whitenoise-mac/Core/WorkspaceState.swift`:

- Add a monotonic `groupImageSearchGeneration` and `beginGroupImageSearch()` / `invalidateGroupImageSearch()` / `isCurrentGroupImageSearch(generation:query:)` helpers. `isCurrentGroupImageSearch` additionally requires the picker still be presented and the live trimmed query still equal the one the search was issued for, so a dismissed or edited picker is never repopulated.
- `searchGroupImages()` captures the generation before the `await` and only commits `groupImageResults` (and only clears `isSearchingGroupImages` in its `defer`) while still current. A stale completion can no longer (a) overwrite a newer search's results, (b) repopulate a dismissed sheet, or (c) clear a spinner a newer search owns.
- Invalidate the in-flight search from `showGroupImagePicker()`, `closeGroupImagePicker()`, the empty-query early return, and the full state-reset teardown.

## Edge cases verified (by hand — Swift, no CI on this runner)

`WorkspaceState` is `@MainActor`-isolated, so all generation reads/writes around the `await` are race-free (the suspension resumes on the main actor) — identical safety to the existing `resolveNewChatQuery` guard.

1. **Out-of-order overwrite:** A (gen 1) → edit query → B (gen 2); A resolves → gen mismatch → dropped. B wins. ✓
2. **Dismissed-picker repopulation:** A in flight → `closeGroupImagePicker()` bumps gen + flips `isGroupImagePickerPresented` false → A dropped. ✓
3. **Reopened picker:** A in flight → close → `showGroupImagePicker()` invalidates again → A cannot land. ✓
4. **Empty-query clear:** clearing the field invalidates a prior in-flight non-empty search. ✓
5. **Spinner ownership:** A's stale `defer` does not clear the spinner a newer B owns; B's `defer` clears it when current — no stuck spinner, no premature clear. ✓

## Test plan

- [ ] Builds in Xcode (macOS — cannot build on the Linux CI runner)
- [ ] Manual: issue two rapid searches with different queries; confirm the grid shows the newer query's results regardless of completion order
- [ ] Manual: start a search, dismiss the picker before it resolves; confirm the grid stays empty

Closes #110

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/114">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -346,6 +346,15 @@ final class WorkspaceState {
     private var deliveredNotificationKeys = Set<String>()
     private var deliveredNotificationKeyOrder: [String] = []
     private var newChatLookupGeneration = 0
+    /// Monotonic token identifying the most recently started group-image (Openverse) search.
+    /// `searchGroupImages` captures the value before its `await` and only commits results /
+    /// clears `isSearchingGroupImages` while it is still current — i.e. no newer search has
+    /// superseded it and the picker is still on screen for the same query. This makes the
+    /// search last-request-wins (a slow earlier search cannot overwrite a newer one) and
+    /// prevents a search resolving after the picker is dismissed/reopened from repopulating
+    /// `groupImageResults`. Mirrors the new-chat lookup / settings-load generation guards
+    /// (issues #2, #4). See `searchGroupImages` / issue #110.
+    private var groupImageSearchGeneration = 0
     /// Raw per-sender FFI lookups (userProfile + directory displayName), cached so that
     /// scrolling back through history does not re-resolve the same senders from Rust on
     /// every page. Keyed by sender accountIdHex.
@@ -1895,6 +1904,7 @@ final class WorkspaceState {
         guard !chat.isDirect else { return }
         lastError = nil
         closeGroupDetails()
+        invalidateGroupImageSearch()
         groupImageSearchQuery = ""
         groupImageResults = []
         isGroupImagePickerPresented = true
@@ -1902,8 +1912,8 @@ final class WorkspaceState {
 
     func closeGroupImagePicker() {
         isGroupImagePickerPresented = false
+        invalidateGroupImageSearch()
         groupImageResults = []
-        isSearchingGroupImages = false
         isSavingGroupImage = false
     }
 
@@ -1915,17 +1925,31 @@ final class WorkspaceState {
     func searchGroupImages() async {
         let query = groupImageSearchQuery.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !query.isEmpty else {
+            invalidateGroupImageSearch()
             groupImageResults = []
             return
         }
 
         lastError = nil
+        let searchGeneration = beginGroupImageSearch()
         isSearchingGroupImages = true
-        defer { isSearchingGroupImages = false }
+        defer {
+            // Only clear the spinner if this search is still the current one. A newer search
+            // (which bumped the generation) owns `isSearchingGroupImages` and must not have it
+            // cleared out from under it by this stale completion.
+            if isCurrentGroupImageSearch(generation: searchGeneration, query: query) {
+                isSearchingGroupImages = false
+            }
+        }
 
         do {
-            groupImageResults = try await groupImageSearchClient.searchImages(query: query)
+            let results = try await groupImageSearchClient.searchImages(query: query)
+            // Drop results if a newer search superseded this one, the query was edited, or the
+            // picker was dismissed/reopened while the request was in flight.
+            guard isCurrentGroupImageSearch(generation: searchGeneration, query: query) else { return }
+            groupImageResults = results
         } catch {
+            guard isCurrentGroupImageSearch(generation: searchGeneration, query: query) else { return }
             groupImageResults = []
             lastError = error.localizedDescription
         }
@@ -2267,7 +2291,7 @@ final class WorkspaceState {
         isGroupImagePickerPresented = false
         groupImageSearchQuery = ""
         groupImageResults = []
-        isSearchingGroupImages = false
+        invalidateGroupImageSearch()
         isSavingGroupImage = false
         isGroupDetailsPresented = false
         groupDetailsSnapshot = nil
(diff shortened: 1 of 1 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define group-image search generation tracking](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/114/chapters/1) |
| 2 | [Guard search results against stale completions](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/114/chapters/2) |
| 3 | [Invalidate searches on picker lifecycle changes](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/114/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/114">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit d7f6045ed6f37aaa5de80efb7dbefff0df8dba0a on Jun 20, 2026 2:26pm UTC.</sub>

agent-p1p:
## Adversarial review: changes requested

Blocking: 1
Suggestions: 0
Nitpicks: 0

### Blocking

- `whitenoise-mac/Core/WorkspaceState.swift:1936-1942` — `defer` uses `isCurrentGroupImageSearch(...)` to decide whether to clear `isSearchingGroupImages`, but that predicate also requires the live trimmed query to still equal the captured query. Repro: start a search for `cat`, edit the text field to `dog` while the Openverse request is still in flight, then wait without submitting the second query. When the `cat` request finishes, `isCurrentGroupImageSearch` is false because the query changed, so the result is dropped and the `defer` also refuses to clear the spinner. Since the Search button is disabled while `isSearchingGroupImages` is true, the picker can remain stuck in `Searching` until the user presses Return or closes the sheet. This is the same stale-async path the PR is trying to harden.

  Suggested fix: separate spinner ownership from result-commit validity. For example, clear the spinner in `defer` when `groupImageSearchGeneration == searchGeneration` (or an `ownsCurrentGroupImageSearch(generation:)` helper) while keeping the stricter `isCurrentGroupImageSearch(generation:query:)` guard for writing `groupImageResults` / `lastError`. Dismiss/reopen paths already bump the generation and clear the spinner, so this still avoids a stale completion clearing a newer search's spinner.

### Verified

- The generation guard does prevent older completions from overwriting newer submitted searches.
- `closeGroupImagePicker()` and `showGroupImagePicker()` now invalidate in-flight searches, so post-dismissal completions are dropped.
- Changed file is outside the configured sensitive globs for this repo.

### Checks

- `gh pr checks 114`: CodeRabbit pass.
- Local build/tests not run: this Linux runner has neither `swift` nor `xcodebuild` installed.

agent-p1p:
Addressed the blocking review finding in d7f6045.

The spinner-clear in `searchGroupImages()`'s `defer` previously went through `isCurrentGroupImageSearch(generation:query:)`, which also requires the live query to match. Editing the query mid-flight without resubmitting therefore dropped the result *and* left `isSearchingGroupImages` stuck `true`, disabling the Search button indefinitely.

Fix: spinner ownership is now keyed on the generation alone via a new `ownsGroupImageSearch(generation:)` helper — only a newer search or an `invalidateGroupImageSearch` (both of which set the spinner state themselves) transfers ownership. The stricter picker/query guard (`isCurrentGroupImageSearch`) still gates writes to `groupImageResults`/`lastError`, so out-of-order or post-dismissal results can never repopulate current UI state.

Edge cases verified:
- out-of-order overwrite: stale search drops its results, does not touch spinner (newer search owns it)
- edit query mid-flight, no resubmit: result dropped, spinner cleared (no longer stuck)
- picker dismissed/reopened mid-flight: result dropped, spinner already cleared by invalidate
- empty-query clear: early return, no spinner left set

PR #115, opened 2026-06-20
fix: reject private/loopback/link-local hosts in RemoteImageURLPolicy (SSRF) (#107)

### LINKED ISSUE #107: RemoteImageURLPolicy allows private/loopback/link-local hosts — peer-controlled avatar URLs can probe the victim's internal network (SSRF)
## Summary

`RemoteImageURLPolicy` is documented as "the single place that decides whether a remote image URL is allowed to be fetched at all" and "the defense-in-depth check that runs even once the user has opted in" (`whitenoise-mac/Core/RemoteImageLoader.swift:10-19`). But `isAllowed(_:)` only enforces **`https` scheme + non-empty host** — it does **not** restrict the *destination address*. So once a user enables "Load Remote Profile Images" (`loadRemoteImages`), an arbitrary peer can put a URL like `https://192.168.1.1/x.png`, `https://10.0.0.5/x.png`, `https://127.0.0.1:8080/x.png`, or `https://[::1]/x.png` in their (untrusted) Nostr `picture` metadata and the client will issue an outbound request to that internal/loopback host.

This is the residual of #10. #10's own suggested fix called for "validate the host," but the shipped policy validates only that a host *string exists*, not that it is a public address. The same gap also re-appears on the redirect path.

## Evidence

- `whitenoise-mac/Core/RemoteImageLoader.swift:26-30` — the entire allow-decision:
  ```swift
  static func isAllowed(_ url: URL) -> Bool {
      guard let scheme = url.scheme?.lowercased(), scheme == "https" else { return false }
      guard let host = url.host, !host.isEmpty else { return false }
      return true   // no check that `host` is a public address
  }
  ```
- `whitenoise-mac/Core/RemoteImageLoader.swift:97-114` — `image(for:maxPixelSize:)` gates every fetch on `isAllowed` only.
- `whitenoise-mac/Core/RemoteImageLoader.swift:269-283` — the redirect handler re-checks `isAllowed`, so a public `https://` avatar could also **3xx-redirect to an internal host** and pass this same insufficient check.
- The `picture` value is untrusted peer metadata (`UserProfileMetadataFfi.picture` → `ChatPeerProfile.pictureURL` / `MessageItem.senderPictureURL`), and the avatar view fetches it once the user opts in (`whitenoise-mac/Views/MessengerShellView.swift:1058-1066`).
- `com.apple.security.network.client` is granted, and the macOS App Sandbox does **not** block LAN/loopback destinations, so these requests do go out.

## Impact

A privacy-focused E2EE messenger should not let a remote correspondent steer the client's HTTP requests at the user's *internal* network. Once remote images are enabled:

- **Internal host / port discovery (SSRF oracle):** load success vs. failure (and timing) tells the sender-controlled server whether `https://<internal-ip>:<port>/` is reachable from the victim — i.e. it can map which LAN hosts/services the victim can see and confirm the victim is on a particular network.
- **Loopback / localhost service probing:** local dev servers, admin panels, or other apps listening on `127.0.0.1` become reachable targets.
- This is reachable purely by being in a DM/group with the victim; the victim only has to view the chat.

Severity is MEDIUM rather than HIGH because it requires the user to have opted into remote images (default off) and the exfiltration channel is a reachability/timing oracle (the attacker generally can't read non-image response bodies).

## Suggested fix

In `RemoteImageURLPolicy.isAllowed` (applied to both the initial URL and every redirect target), additionally reject URLs whose host resolves to a non-public address:

- Reject literal IPs in private/loopback/link-local/ULA ranges: `127.0.0.0/8`, `10/8`, `172.16/12`, `192.168/16`, `169.254/16`, `::1`, `fc00::/7`, `fe80::/10`, and `0.0.0.0`.
- Reject obvious local hostnames (`localhost`, `*.local`).
- For non-literal hostnames, ideally resolve and re-validate the resolved address before/at connection time (or route image loads through a fixed proxy/CDN the app controls), since DNS can otherwise point a public-looking host at an internal IP (DNS rebinding).

Reusing the loopback/range logic already present in `RelayURLValidator.isLoopbackHost` (`whitenoise-mac/Core/RelayURLValidator.swift:102-125`) is a good starting point, extended to the full set of private ranges.

agent-p1p:
## Summary

`RemoteImageURLPolicy.isAllowed` is documented as the single chokepoint deciding whether a remote avatar URL may be fetched, but it only enforced **`https` scheme + non-empty host**. It did **not** restrict the destination address. So once a user enabled "Load Remote Profile Images", an arbitrary peer could put a URL like `https://192.168.1.1/x.png`, `https://127.0.0.1:8080/x.png`, or `https://[::1]/x.png` in their untrusted Nostr `picture` metadata and the client would issue an outbound request to that internal/loopback host — an SSRF reachability/timing oracle for internal host and port discovery (and the same gap re-appeared on the redirect path).

This is the residual of #10: #10 validated only that a host *string exists*, not that it is a public address.

## Fix

Add an address check to `isAllowed` (which is already re-run on every redirect target in `CappedImageDownloadDelegate.willPerformHTTPRedirection`, so this closes both the initial-URL and redirect paths):

- **IPv4 literals** in `127.0.0.0/8`, `0.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16` are rejected — including obfuscated decimal/hex/octal/shorthand forms (`2130706433`, `0x7f000001`, `0177.0.0.1`, `127.1`) so a naive dotted-quad check cannot be bypassed.
- **IPv6 literals** `::`, `::1`, `fc00::/7` (ULA), `fe80::/10` (link-local), and IPv4-mapped/compatible addresses whose embedded IPv4 is private are rejected.
- **Local hostnames** (`localhost`, `*.local`) are rejected.

A small dependency-free `IPAddress` literal parser backs the check. Parsing literals ourselves (instead of `inet_pton`) keeps the accepted grammar explicit and lets us catch the obfuscated IPv4 forms that hide a private address.

### Documented limitation

A *public-looking* DNS hostname can still resolve to a private IP (DNS rebinding), because `URLSession` performs its own resolution after this check and we do not pin the resolved address at connection time. This is documented on `isAllowed`; fully closing it requires connection-time address re-validation (custom resolver), a much larger change. This PR deterministically closes the directly-exploitable literal-IP and local-hostname vectors — what the attacker can set without controlling DNS.

## Test plan

New tests in `whitenoise-macTests`:
- `remoteImagePolicyRejectsPrivateAndLoopbackHosts` — blocked/allowed host matrix end-to-end through `isAllowed` (IPv4, obfuscated IPv4, IPv6, local hostnames; public hosts and public IP literals still allowed).
- `remoteImageSanitizedURLRejectsPrivateHosts` — the string entry point used by the UI also rejects internal destinations.
- `ipAddressParserHandlesLiteralForms` — direct coverage of the literal parser (dotted-quad, decimal, hex, octal, shorthand IPv4; `::` compression, IPv4-mapped tail IPv6).

> Note: whitenoise-mac is a macOS-only Xcode project; it cannot be built or tested on the Linux CI host. Verification relies on Mac CI + CodeRabbit.

Closes #107

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Substantially strengthened remote image URL validation to prevent unauthorized access attempts targeting private networks, loopback addresses, local system resources, and various obfuscated IP address formats.
  * Significantly enhanced security controls to effectively defend against server-side request forgery and other unauthorized network access threats that impact overall application safety and security.


ghost:

| | Title |
|---|-------|
| 1 | [Implement IP address literal parser](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115/chapters/1) |
| 2 | [Enforce SSRF protection in RemoteImageURLPolicy](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115/chapters/2) |
| 3 | [Verify SSRF protections with unit tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115/chapters/3) |
| 4 | [Update project configuration and versioning](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/115">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2f10f5f280d894fbf61ba25fff82881c806d113e on Jun 21, 2026 4:23pm UTC.</sub>

whitenoise-mac/Core/RemoteImageLoader.swift
```diff
@@ -42,6 +143,112 @@ enum RemoteImageURLPolicy {
...
+        default: maxLast = 0xFF
+        }
+        guard let last = nums.last, last <= maxLast else { return nil }
+
+        var addr: UInt32 = 0
+        for n in nums.dropLast() {
+            addr = (addr << 8) | UInt32(n)
+        }
+        // Shift the leading octets up to make room for the final (multi-byte) part.
+        addr <<= UInt32((4 - nums.count) * 8)
```
agent-p1p: 🔴 **Blocking — off-by-one shift corrupts every multi-part IPv4 literal.**

The exponent here is wrong. For an `N`-part literal the `N-1` leading octets occupy the high `(N-1)` bytes and `last` fills the remaining `(5 - N)` bytes, so the leading group must be shifted left by `(5 - nums.count) * 8`, not `(4 - nums.count) * 8`. As written, `count == 4` shifts by `0`, so the first octet is dropped and `last` is OR'd into the wrong byte:

```
8.8.8.8       -> (0, 8, 8, 8)        (should be 8.8.8.8)
1.1.1.1       -> (0, 1, 1, 1)        (should be 1.1.1.1)
192.168.1.1   -> (0, 192, 168, 1)    (should be 192.168.1.1)
203.0.113.5   -> (0, 203, 0, 117)    (last OR'd into a non-empty byte: 0x71|0x05)
```

Two consequences:

1. **Public IPv4 literals are wrongly rejected.** `8.8.8.8` parses to `(0,8,8,8)`, whose first octet `0` hits the `0.0.0.0/8` branch in `isPrivateIPv4`, so `isAllowed` returns `false`. The PR's own positive assertions fail: `remoteImagePolicyRejectsPrivateAndLoopbackHosts` asserts `https://8.8.8.8/x.png`, `https://1.1.1.1/x.png`, and `https://[::ffff:8.8.8.8]/x.png` (the IPv4-mapped path reuses `parseIPv4`) are **allowed** — all three `#expect` lines will fail on a real Swift build. (CodeRabbit is static and does not compile/run, which is why CI is green.)

2. **Private-range matching for dotted-quad is correct only by accident.** Loopback/private dotted-quad literals still get blocked, but only because the corrupted first octet is always `0` (∈ `0.0.0.0/8`) — not because the range logic matched the real address. The SSRF guard's dotted-quad correctness is illusory and brittle; a later "fix" that restores the first octet without fixing this shift would silently reopen the hole.

Single-part decimal/hex forms (`2130706433`, `0x7f000001`) parse correctly (no leading bytes), and the IPv6 group logic (ULA/link-local/loopback) is sound — only the `parseIPv4`-dependent paths are affected.

Fix:

```swift
addr <<= UInt32((5 - nums.count) * 8)
```

Verified across all forms with this: `8.8.8.8`→`(8,8,8,8)`, `127.1`→`(127,0,0,1)`, `2130706433`→`(127,0,0,1)`, `203.0.113.5`→`(203,0,113,5)`, `0.0.0.0`→`(0,0,0,0)`. (For `count == 1` the exponent is 32 but the leading group is empty so `addr` is `0`; Swift's `<<` returns 0 for over-width shifts — safe.)

agent-p1p (commented):
## Adversarial review — #115 (SSRF host policy, fixes #107)

**Verdict: Changes Requested (1 blocking).** The SSRF intent, scope, and structure are right — surgical change at the single `isAllowed` chokepoint, redirect path already covered by the existing `willPerformHTTPRedirection` re-check, documented DNS-rebinding limitation, and a good blocked/allowed test matrix. But the IPv4 literal parser has a byte-placement bug that breaks the dotted-quad path and fails the PR's own positive tests.

### 🔴 Blocking

**`whitenoise-mac/Core/RemoteImageLoader.swift:186` — off-by-one shift corrupts every multi-part IPv4 literal.**
`addr <<= UInt32((4 - nums.count) * 8)` should be `(5 - nums.count) * 8`. For a 4-part literal the shift is `0`, so the first octet is dropped and `last` lands in the wrong byte:
- `8.8.8.8` → `(0,8,8,8)`, `192.168.1.1` → `(0,192,168,1)`, `203.0.113.5` → `(0,203,0,117)`.

Effects:
1. Public IPv4 literals are wrongly **blocked** (`8.8.8.8`→first octet `0`→`0.0.0.0/8` branch). The PR's own `allowed` assertions for `https://8.8.8.8/...`, `https://1.1.1.1/...`, and `https://[::ffff:8.8.8.8]/...` will **fail** on a real `swift test` run — CodeRabbit is static and doesn't compile, which is why CI shows green.
2. Private dotted-quad literals stay blocked only by the accident that the corrupted first octet is always `0` (∈ `0.0.0.0/8`), not because the range logic matched — the dotted-quad SSRF guard is correct by coincidence and fragile.

One-line fix: `addr <<= UInt32((5 - nums.count) * 8)`. Full inline detail with the verification matrix is on the diff.

### 💡 Suggestions

- **Octal edge case (non-blocking):** `parseUInt` rejects parts like `08`/`09` (invalid octal) → `08.0.0.1` is treated as a hostname and allowed. `inet_aton` rejects it too and most resolvers won't resolve it, so the practical risk is low, but a test asserting the behavior either way would lock it in.
- After fixing line 186, add a positive parse assertion for a full dotted-quad in `ipAddressParserHandlesLiteralForms` (e.g. `parseIPv4("8.8.8.8") == [8,8,8,8]`) — the current parser tests only exercise the single-number/shorthand/octet-overflow forms, which is exactly why this slipped through.

### ✅ Looks good

- Correct chokepoint: extending `isAllowed` covers both the initial URL and every redirect (the delegate re-runs `isAllowed`); no new attack surface.
- IPv6 group logic (`::`, `::1`, `fc00::/7`, `fe80::/10`) is correct; only the `parseIPv4`-dependent IPv4-mapped tail inherits the bug above.
- 169.254.169.254 (cloud metadata) explicitly covered; DNS-rebinding residual honestly documented rather than silently ignored.

Sensitive path touched: `whitenoise-mac/Core/RemoteImageLoader.swift` (matches `**/RemoteImageLoader.swift`). Per policy this does not block the fix, but it does mean the merge gate requires Jeff's explicit approval and must not auto-merge.

---
*Adversarial review by Pip (agent-p1p), reviewer profile claude-reviewer (opposite family from the fixer).*

whitenoise-mac/Core/RemoteImageLoader.swift
agent-p1p: Fixed in 4741c888c61f949e79c42839283909495d7f6b50: changed the leading-octet shift to `(5 - nums.count) * 8` and added direct dotted-quad parser assertions for `8.8.8.8` and `203.0.113.5` to lock the byte placement.

PR #123, opened 2026-06-21
fix: bound media JSON attachment traversal

### LINKED ISSUE #120: Hardening: media-JSON attachment parser recurses without a depth bound on untrusted message content (potential stack-overflow DoS)
## Summary

`MessageMediaParser` parses a message's `mediaJson` — decrypted inner content authored by another group member, i.e. untrusted — with `JSONSerialization` and then walks the resulting object graph with **unbounded mutual recursion**. A peer who crafts a deeply nested `media` payload (nested objects or nested arrays) can drive recursion as deep as the JSON nesting, which risks a stack overflow → app crash when the message is mapped for display. In a messenger this is a remote DoS reachable by any group member; the victim only has to open/scroll the conversation.

## Evidence

`whitenoise-mac/Core/MarmotMapping.swift:371-410`:

```swift
private static func references(fromMediaJson mediaJson: String?) -> [MediaAttachmentReferenceFfi] {
    guard let mediaJson,
          let data = mediaJson.data(using: .utf8),
          let root = try? JSONSerialization.jsonObject(with: data)   // untrusted, no options/limits
    else { return [] }
    return references(fromJSONObject: root)
}

private static func references(fromJSONObject value: Any) -> [MediaAttachmentReferenceFfi] {
    if let dictionary = value as? [String: Any] {
        var output: [MediaAttachmentReferenceFfi] = []
        if let imeta = dictionary["imeta"] { /* … */ }
        if let media = dictionary["media"] {
            output.append(contentsOf: references(fromJSONObject: media))   // ← recursion per "media" nesting level
        }
        if let direct = reference(fromJSONObject: dictionary) { output.append(direct) }
        return output
    }
    if let array = value as? [Any] {
        if let tagReferences = references(fromIMetaArray: array, sourceEpoch: nil), !tagReferences.isEmpty {
            return tagReferences
        }
        return array.flatMap(references(fromJSONObject:))                   // ← recursion per array nesting level
    }
    return []
}
```

Recursion depth tracks the input nesting depth via either `{"media": {"media": {"media": … }}}` or `[[[[ … ]]]]`. `mediaJson` originates from `TimelineMessageRecordFfi.mediaJson` and is parsed during `MessageItem` construction (`MarmotMapping.swift:118-122`, `:273-277`), which runs on the timeline mapping path for received messages. There is no app-side cap on inner message size or JSON nesting depth before this walk.

## Impact / severity

Marked **LOW** pending verification, but with remote-DoS potential:
- If `JSONSerialization.jsonObject(with:)` itself parses deeply nested input (it has no documented nesting limit), the subsequent recursive walk can overflow the stack and crash the app on a crafted message.
- If `JSONSerialization` rejects sufficiently deep nesting first, the practical risk is reduced to that parser's own limit — which is why this is filed as hardening rather than confirmed HIGH. **This should be verified** (feed `references(fromMediaJson:)` a synthetic deeply-nested payload, e.g. tens of thousands of `[` levels, and confirm whether it crashes).

## Suggested fix

- Add an explicit depth limit to `references(fromJSONObject:)` (pass and decrement a `maxDepth`, returning `[]` past the bound) — a small, safe number (e.g. 16–32) is far more than any legitimate `imeta`/`media` structure needs.
- Optionally cap `mediaJson` length and/or `JSONSerialization` output size before walking.
- Consider decoding into a fixed `Decodable` shape instead of an open-ended `[String: Any]`/`[Any]` walk, which removes the unbounded recursion entirely.

agent-p1p:
## Summary
- bounds media JSON attachment traversal to depth 32 before and after JSONSerialization
- rejects overly nested object/array payloads before walking peer-controlled mediaJson
- adds regression coverage for deeply nested media object and array payloads

## Test plan
- `git diff --check`
- `python3` smoke test of the JSON nesting scanner logic (shallow payload false; object depth 40 true; array depth 40 true; brackets inside strings false)
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -destination 'platform=macOS' -only-testing:whitenoise-macTests/whitenoise_macTests/deeplyNestedMediaJSONDoesNotProduceAttachments` (blocked locally: `xcodebuild: command not found` on this Linux worker)

Closes #120

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards to limit how deeply nested media JSON is traversed during timeline mapping, preventing excessive recursion on attacker-controlled payloads.
  * Deeply nested/invalid media JSON now results in a safe fallback body and no unintended media attachments.
  * Moderately nested media JSON still extracts expected attachments.

* **Tests**
  * Added regression tests covering deep-nesting behavior (ensuring no attachments and correct fallback) and near-limit behavior (ensuring attachments are still produced).
  * Enhanced test media JSON generation helpers to support controlled nesting depths.

### DIFF
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -390,6 +390,8 @@ private nonisolated enum MessageMediaParser {
             var output: [MediaAttachmentReferenceFfi] = []
 
             if let imeta = dictionary["imeta"] {
+                // Safe without its own depth counter because the raw JSON pre-scan rejects
+                // any object graph deeper than maxMediaJSONTraversalDepth before parsing.
                 output.append(
                     contentsOf: references(
                         fromIMetaValue: imeta,
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -349,6 +349,8 @@ private enum MarmotTimelineKind {
 }
 
 private nonisolated enum MessageMediaParser {
+    private static let maxMediaJSONTraversalDepth = 32
+
     static func attachments(
         mediaJson: String?,
         tags: [MessageTagFfi],
@@ -370,14 +372,20 @@ private nonisolated enum MessageMediaParser {
 
     private static func references(fromMediaJson mediaJson: String?) -> [MediaAttachmentReferenceFfi] {
         guard let mediaJson,
+              !mediaJSONNestingExceedsLimit(mediaJson, maxDepth: maxMediaJSONTraversalDepth),
               let data = mediaJson.data(using: .utf8),
               let root = try? JSONSerialization.jsonObject(with: data)
         else { return [] }
 
-        return references(fromJSONObject: root)
+        return references(fromJSONObject: root, remainingDepth: maxMediaJSONTraversalDepth)
     }
 
-    private static func references(fromJSONObject value: Any) -> [MediaAttachmentReferenceFfi] {
+    private static func references(
+        fromJSONObject value: Any,
+        remainingDepth: Int
+    ) -> [MediaAttachmentReferenceFfi] {
+        guard remainingDepth >= 0 else { return [] }
+
         if let dictionary = value as? [String: Any] {
             var output: [MediaAttachmentReferenceFfi] = []
 
@@ -390,7 +398,7 @@ private nonisolated enum MessageMediaParser {
                 )
             }
             if let media = dictionary["media"] {
-                output.append(contentsOf: references(fromJSONObject: media))
+                output.append(contentsOf: references(fromJSONObject: media, remainingDepth: remainingDepth - 1))
             }
             if let direct = reference(fromJSONObject: dictionary) {
                 output.append(direct)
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define media JSON traversal depth limit](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123/chapters/1) |
| 2 | [Implement JSON nesting depth scanner](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123/chapters/2) |
| 3 | [Enforce depth limits during attachment parsing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123/chapters/3) |
| 4 | [Add regression tests for nested media JSON](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/123">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit cb61e682b0b1e20da3971c4c51ab5e96aeb5e5cd on Jun 21, 2026 9:02am UTC.</sub>

agent-p1p (commented):
## Pip adversarial review — PR #123 (whitenoise-mac#120)

Reviewed the depth-bound fix for unbounded recursion in `MessageMediaParser`. Verified the pre-scan logic against boundary and bypass cases (object/array nesting, brackets inside strings, escaped quotes, the depth-32 boundary).

**BLOCKING:**
- none

**SUGGESTION:**
- `MarmotMapping.swift:392-399` — the `imeta` path (`references(fromIMetaValue:)` → `references(fromIMetaArray:)`) is not depth-tracked, unlike the `media`/array paths. It's safe today only because `mediaJSONNestingExceedsLimit` caps total structural nesting at 32 *before* `JSONSerialization` parses, so the object graph that reaches the walker can't exceed depth 32 on any path. That makes the pre-scan load-bearing for the `imeta` path's safety. Consider a one-line comment noting this so a future refactor that weakens/removes the pre-scan doesn't silently reopen the recursion on the `imeta` branch.
- The issue also suggested capping `mediaJson` length. Not done. Non-blocking — nesting depth is the actual stack-overflow vector and that is now bounded; a length cap would only add defense against allocation-size abuse, which is a separate (and lower) concern.

**NITPICK:**
- `mediaJSONNestingExceedsLimit` always scans the entire string. Fine for the bounded `mediaJson` sizes on this path; no change needed.

**SUMMARY:** The fix correctly closes the unbounded-recursion DoS. The primary guard is a string-level, string/escape-aware nesting scanner that rejects any payload exceeding depth 32 *before* `JSONSerialization.jsonObject(with:)` runs — so the question of whether Apple's parser itself would crash on deep input (the issue's open verification item) is moot: such input never reaches it. The in-walk `remainingDepth` is a correct belt-and-suspenders bound. I verified the scanner ignores brackets inside JSON strings and after escaped quotes (no false positives), passes a legitimate depth-3 `imeta` payload, passes at depth 32, and rejects at depth 33+. The regression test wraps a genuinely valid `imeta` reference in 40 layers of object and array nesting and asserts both messages map to zero attachments with a graceful "Unsupported message" body rather than crashing — a real regression test, not a tautology. No scope creep, no sensitive paths touched.

**VERDICT:** APPROVE
**SENSITIVE_PATHS:** none

agent-p1p:
@coderabbitai review

agent-p1p:
@coderabbitai review

PR #146, opened 2026-06-24
fix: avoid dangling timeline subscription

### LINKED ISSUE #145: loadMessages unconditionally records activeTimelineSubscription even when startTimelineListener bails, leaking a dangling subscription with no draining task
## Summary

In `loadMessages(groupIdHex:)`, the live timeline subscription is recorded into `activeTimelineSubscription` / `activeTimelineGroupId` **unconditionally**, immediately after calling `startTimelineListener`. But `startTimelineListener` can decide not to start a listener at all (its internal guard fails after a selection change across the preceding `await`), in which case no draining `Task` is created and `timelineTaskGroupId` stays `nil`. The result is a live `TimelineMessagesSubscription` retained with nothing consuming it, and internally inconsistent state (`activeTimelineGroupId` set while `timelineTaskGroupId` is nil).

## Where

```swift
// loadMessages — 1524-1529
await applyTimelineWindow(page, groupIdHex: groupIdHex, account: activeAccount, client: client)  // suspension point
// Start the listener first (it tears down any prior listener, which would clear
// these), then record the subscription so scroll-back pagination can reach it.
startTimelineListener(groupIdHex: groupIdHex, account: activeAccount, subscription: subscription)
activeTimelineSubscription = subscription   // 1528 — unconditional
activeTimelineGroupId = groupIdHex          // 1529 — unconditional
```

```swift
// startTimelineListener — 3130-3146
private func startTimelineListener(groupIdHex: String, account: AccountItem, subscription: ...) {
    guard client != nil else { return }
    stopTimelineListener()                                  // nils activeTimelineSubscription/GroupId
    guard activeAccountId == account.id, selectedChat?.id == groupIdHex else { return }  // may bail here
    timelineTaskGroupId = groupIdHex
    timelineTask = Task { [weak self] in
        await self?.runTimelineListener(groupIdHex: groupIdHex, account: account, existingSubscription: subscription)
    }
}
```

## The problem

`applyTimelineWindow` (1524) is awaited, and because `WorkspaceState` is `@MainActor`, the main actor can run other work (e.g. the user selecting a different chat or a settings page) during that suspension. If selection changes there, `startTimelineListener`'s guard at the third line fails: it has already run `stopTimelineListener()` (clearing the fields) and now returns **without** creating the task or setting `timelineTaskGroupId`. Control returns to `loadMessages`, which then sets `activeTimelineSubscription = subscription` / `activeTimelineGroupId = groupIdHex` anyway.

Net effect: a live UniFFI subscription handle (an open relay subscription on the Rust side) is held with no `next()` loop draining it and with `timelineTask == nil`. In the common case it is reclaimed by the next `stopTimelineListener()` (e.g. the subsequent `loadMessages` for the new chat), so it self-heals quickly; but if the selection changed to a non-chat destination (a settings page) that does not spawn a follow-up `loadMessages`, the dangling subscription persists until the next chat selection.

## Suggested fix

Only record the subscription when the listener actually started — e.g. re-check the guard before the assignment, or have `startTimelineListener` own the assignment / return whether it started:

```swift
startTimelineListener(groupIdHex: groupIdHex, account: activeAccount, subscription: subscription)
guard timelineTaskGroupId == groupIdHex else { return }   // listener actually started
activeTimelineSubscription = subscription
activeTimelineGroupId = groupIdHex
```

## Severity

LOW — narrow timing window and self-heals on the next chat switch in the common case; the pagination guards (`selectedChat?.id == groupIdHex`) prevent the stale subscription from being misused. It is a real state-consistency / resource-retention gap rather than a data-correctness bug.

agent-p1p:
Closes #145

## Summary
- Guard `loadMessages(groupIdHex:)` after `startTimelineListener(...)` so `activeTimelineSubscription` / `activeTimelineGroupId` are recorded only when the listener actually starts.
- This prevents retaining a live `TimelineMessagesSubscription` with no draining task when selection changes during the preceding `applyTimelineWindow` await.
- Kept the change to the single affected state transition in `WorkspaceState`.

## Verification
- `git diff --check` — passed.
- Static guard/order check — passed: `guard timelineTaskGroupId == groupIdHex else { return }` now sits between `startTimelineListener(...)` and the active subscription assignment.
- `bash -n scripts/ci/macos-sanity-checks.sh` — passed.
- macOS/Xcode CI commands were not runnable on this Linux host: `xcrun`, `swift-format`, `swift`, `xcodebuild`, `plutil`, and `/usr/libexec/PlistBuddy` are unavailable.

## Sensitive paths
none

## Notes
No regression test was added. The affected state (`activeTimelineSubscription`, `activeTimelineGroupId`, `timelineTaskGroupId`) is private, and deterministically reproducing the race would require test-only timing hooks around `applyTimelineWindow` / `messageSenderProfiles`; that is disproportionate for this one-line guard.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/146">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -1504,7 +1504,10 @@ final class WorkspaceState {
             if let row = try await runOffMain({
                 try client.initializeChatReadState(accountRef: accountRef, groupIdHex: groupIdHex)
             }) {
-                await applyChatRow(row, account: activeAccount)
+                // `initializeChatReadState` may race a live chat-list delta. Do not let an
+                // older read-state row roll back a newer preview/timestamp already applied
+                // by the subscription listener while the FFI call was in flight.
+                await applyChatRow(row, account: activeAccount, skippingStaleRow: true)
             }
 
             let subscription = try await client.subscribeTimelineMessages(
@@ -3292,7 +3295,11 @@ final class WorkspaceState {
         }
     }
 
-    private func applyChatRow(_ row: ChatListRowFfi, account: AccountItem) async {
+    private func applyChatRow(
+        _ row: ChatListRowFfi,
+        account: AccountItem,
+        skippingStaleRow: Bool = false
+    ) async {
         guard activeAccountId == account.id else { return }
 
         var chats = chatsByAccount[account.id] ?? []
@@ -3302,6 +3309,12 @@ final class WorkspaceState {
         }
 
         let chat = baseChatItem(from: row, account: account)
+        if skippingStaleRow,
+            let current = chats.first(where: { $0.id == chat.id }),
+            isOlderChatRow(chat, than: current)
+        {
+            return
+        }
         if let index = chats.firstIndex(where: { $0.id == chat.id }) {
             chats[index] = chat
         } else {
@@ -3309,6 +3322,12 @@ final class WorkspaceState {
         }
         chatsByAccount[account.id] = sortedChatItems(chats)
         startChatListEnrichment(rows: [row], account: account, replacingCurrent: false)
+    }
+
+    private func isOlderChatRow(_ candidate: ChatItem, than current: ChatItem) -> Bool {
+        guard let currentUpdatedAt = current.updatedAt else { return false }
+        guard let candidateUpdatedAt = candidate.updatedAt else { return true }
+        return candidateUpdatedAt < currentUpdatedAt
     }
 
     private func removeChat(groupIdHex: String, account: AccountItem) {
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -1524,7 +1524,11 @@ final class WorkspaceState {
             await applyTimelineWindow(page, groupIdHex: groupIdHex, account: activeAccount, client: client)
             // Start the listener first (it tears down any prior listener, which would clear
             // these), then record the subscription so scroll-back pagination can reach it.
+            // `startTimelineListener` can bail without starting a task (e.g. selection changed
+            // while we awaited above); only record the subscription when it actually started,
+            // otherwise we leak a live handle with no `next()` loop draining it.
             startTimelineListener(groupIdHex: groupIdHex, account: activeAccount, subscription: subscription)
+            guard timelineTaskGroupId == groupIdHex else { return }
             activeTimelineSubscription = subscription
             activeTimelineGroupId = groupIdHex
         } catch {


ghost:

| | Title |
|---|-------|
| 1 | [Add staleness check to chat row updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/146/chapters/1) |
| 2 | [Prevent race conditions during timeline loading](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/146/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/146">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1e4e9cf874e6fa317489e8a51be95cd2bbad757c on Jun 24, 2026 4:52pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Verdict: blocked until CI is green.

### Blocking
- **PR Checks / Unit tests** — the current PR head (`d5b0c15`) has red CI. The failing test is `whitenoise_macTests.incrementalChatRowUpdateEnrichesDirectChatTitle()` with failures at `whitenoise_macTests.swift:3400` and `whitenoise_macTests.swift:3409`:
  - `Expected incremental direct-chat enrichment. title=Alice Actual preview=No messages yet pictureURL=https://example.com/alice.png detailsCalls=1`
  - `Expectation failed: didEnrichIncrementally`

  Check run: https://github.com/marmot-protocol/whitenoise-mac/actions/runs/28112651481/job/83243848934

### Suggestions
- None.

### Nitpicks
- None.

### Looks good
- The actual diff is scoped to `WorkspaceState.loadMessages(groupIdHex:)`.
- The new guard is placed after `startTimelineListener(...)` and before `activeTimelineSubscription` / `activeTimelineGroupId` are recorded, so the dangling-subscription path from #145 is closed when `startTimelineListener` bails after a selection/account change.
- `defer { finishTimelineInitialLoad(...) }` still runs on the early return.
- Local checks available on this Linux host passed: `git diff --check origin/master...HEAD` and `bash -n scripts/ci/macos-sanity-checks.sh`.

No sensitive-path changes found in this PR.

agent-p1p:
## Pip address-review update

Pushed `1e4e9cf` to address the red `incrementalChatRowUpdateEnrichesDirectChatTitle()` CI failure.

Root cause: `loadMessages` could apply the `initializeChatReadState` row after a newer chat-list subscription delta, rolling the row preview/timestamp back to the stale snapshot while keeping the enriched direct-chat title/avatar. The update now skips only older read-state rows, preserving newer live chat-list data while keeping equal/newer rows authoritative.

Verification:
- `git diff --check origin/master...HEAD` passed
- `bash -n scripts/ci/macos-sanity-checks.sh` passed
- GitHub PR Checks passed on head `1e4e9cf`
- GitHub Static Analysis passed on head `1e4e9cf`
- CodeRabbit status is green; no inline review comments

PR marked ready for review.

PR #148, opened 2026-06-24
fix: isolate media download state observation

### LINKED ISSUE #144: mediaDownloadState reads the whole mediaDownloads dictionary, so every visible media tile re-renders whenever any single attachment finishes downloading
## Summary

Every media tile / attachment view in a conversation reads its download state through `workspace.mediaDownloadState(...)`, which accesses the single `@Observable` `mediaDownloads` dictionary. Under Swift Observation, reading a stored property (even via a subscript) subscribes the view to **any** mutation of that property. So when one attachment finishes downloading (one key changes), SwiftUI invalidates **every** visible media tile in the conversation, not just the one whose state changed — O(N) re-renders per download completion, and the re-renders re-run the in-`body` `NSImage(data:)` decodes (see #132).

## Where

```swift
// 118
private(set) var mediaDownloads: [String: MediaDownloadState] = [:]

// 1646-1648
func mediaDownloadState(for message: MessageItem, attachment: MessageMediaAttachment) -> MediaDownloadState {
    mediaDownloads[mediaDownloadKey(message: message, attachment: attachment)] ?? .idle
}
```

Read sites, all inside per-attachment views in the message `ForEach`:

- `MessageVisualMediaTile.content` (`MessengerShellView.swift:~2586`)
- `MessageMediaAttachmentView.body` (`~2640`)
- `MessageImageGalleryOverlay.imageContent` (`~3179`)

## The problem

Swift's `@Observable` macro tracks dependencies at **stored-property** granularity. A read of `mediaDownloads[someKey]` registers a dependency on the whole `mediaDownloads` property; a write to *any* key calls `withMutation(keyPath: \.mediaDownloads)`, invalidating every observer that read it. There is no per-key isolation. In a transcript with many images downloading concurrently, each completion mutates `mediaDownloads` once and forces every visible tile/attachment/gallery `body` to re-evaluate — producing O(N²) re-render churn across a screen of N loading attachments, compounded by the full-resolution decodes those bodies perform on each pass (#132).

## Suggested fix

Stop having each tile read the shared dictionary. Resolve each attachment's `MediaDownloadState` once in the parent (e.g. in `ConversationMessageRow`, which already passes a resolved `MessageItem` down) and hand the per-attachment value into the tile by value; or store download state on a per-attachment observable sub-object so a key change only invalidates the one matching row.

## Severity

MEDIUM — SwiftUI re-render fan-out on the primary media path; amplifies the in-body decode cost from #132.

agent-p1p:
## Summary
- Replace the shared observed `mediaDownloads` value dictionary with ignored per-attachment `MediaDownloadStateStore` objects.
- Pass each attachment's store into visual tiles, file/audio rows, and the gallery overlay so only the matching attachment view observes state changes.
- Add a regression test that verifies unrelated media state changes do not invalidate WorkspaceState observation.

## Test Plan
- `git diff --check`
- `xcodebuild -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -destination 'platform=macOS' test` (not run: `xcodebuild` is not installed on this Linux worker)

Closes #144

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -1,5 +1,6 @@
 import AVFoundation
 import AppKit
+import Combine
 import Foundation
 import MarmotKit
 import Observation
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -85,6 +85,16 @@ struct ChatListRowEnrichmentTracker {
 }
 
 @MainActor
+final class MediaDownloadStateStore: ObservableObject {
+    @Published private(set) var state: MediaDownloadState = .idle
+
+    func update(_ newState: MediaDownloadState) {
+        guard state != newState else { return }
+        state = newState
+    }
+}
+
+@MainActor
 @Observable
 final class WorkspaceState {
     enum Phase: Equatable {
@@ -115,7 +125,7 @@ final class WorkspaceState {
     private(set) var accounts: [AccountItem]
     private(set) var chatsByAccount: [String: [ChatItem]]
     private(set) var messagesByChat: [String: [MessageItem]]
-    private(set) var mediaDownloads: [String: MediaDownloadState] = [:]
+    @ObservationIgnored private var mediaDownloads: [String: MediaDownloadStateStore] = [:]
     /// Error for the user-initiated action on the *current* screen. Rendered by form
     /// surfaces (login, settings, new-chat composer). Must never be written by
     /// background tasks — see `backgroundStatus`.
@@ -1686,27 +1696,35 @@ final class WorkspaceState {
     }
 
     func mediaDownloadState(for message: MessageItem, attachment: MessageMediaAttachment) -> MediaDownloadState {
-        mediaDownloads[mediaDownloadKey(message: message, attachment: attachment)] ?? .idle
+        mediaDownloadStateStore(for: message, attachment: attachment).state
+    }
+
+    func mediaDownloadStateStore(
+        for message: MessageItem,
+        attachment: MessageMediaAttachment
+    ) -> MediaDownloadStateStore {
+        mediaDownloadStateStore(forKey: mediaDownloadKey(message: message, attachment: attachment))
     }
 
     func loadMediaAttachment(_ attachment: MessageMediaAttachment, for message: MessageItem) async {
         let key = mediaDownloadKey(message: message, attachment: attachment)
-        if case .loaded = mediaDownloads[key] {
+        let stateStore = mediaDownloadStateStore(forKey: key)
+        if case .loaded = stateStore.state {
             return
         }
-        if case .loading = mediaDownloads[key] {
+        if case .loading = stateStore.state {
             return
         }
 
         guard let client, let activeAccount, !message.groupIdHex.isEmpty else {
-            mediaDownloads[key] = .failed(L10n.string("Attachment unavailable"))
+            stateStore.update(.failed(L10n.string("Attachment unavailable")))
             return
         }
 
         let accountId = activeAccount.id
         let accountRef = activeAccount.accountRef
         let groupIdHex = message.groupIdHex
-        mediaDownloads[key] = .loading
+        stateStore.update(.loading)
 
         do {
             let reference = try await resolvedMediaReference(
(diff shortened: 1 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce MediaDownloadStateStore for isolated state observation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148/chapters/1) |
| 2 | [Update WorkspaceState to manage per-attachment stores](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148/chapters/2) |
| 3 | [Wire isolated download state into UI components](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148/chapters/3) |
| 4 | [Add regression tests for observation isolation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/148">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit d6641bc23914713405f811b1e68cf3280609bd79 on Jun 24, 2026 7:21pm UTC.</sub>

agent-p1p:
## Pip adversarial review

VERDICT: REQUEST_CHANGES

BLOCKING:
- **Red CI — swift-format lint failure**: `whitenoise-mac/Core/WorkspaceState.swift:1700-1707` — CI reports `[AddLines]` / `[Indentation]` failures on the nested `stateStore.update(.loaded(...))` block. This is in a touched file and must be fixed before merge. `git diff --check` passes locally, but this Linux worker does not have `swift-format`/`xcodebuild`; the GitHub check is the authoritative signal here. Static Analysis is also red; its visible actor-isolation errors appear to be in untouched files, but the PR remains non-mergeable while checks are red.

SUGGESTION:
- **Strengthen the regression test to exercise a real download state transition**: `whitenoise-macTests/whitenoise_macTests.swift:1117-1124` — the test creates `WorkspaceState` without bootstrapping/logging in, so `client == nil` / `activeAccount == nil` and `loadMediaAttachment(secondAttachment, ...)` only hits the early `Attachment unavailable` failure path at `WorkspaceState.swift:1677-1679`. The observation assertion is still useful, but it does not cover `.loading -> .loaded`, `client.downloadMedia`, or the `activeAccountId == accountId` guard. Prefer a bootstrapped fake account/runtime and assert that the second store changes while the first store does not publish.
- **Residual coarse observation via `activeAccountId`**: `whitenoise-mac/Core/WorkspaceState.swift:1660-1664` and `3460-3462` — parent views resolve stores in `body` through `mediaDownloadStateStore(...)`, which calls `mediaDownloadKey(...)` and reads `activeAccountId`, an observed `WorkspaceState` property. This does not reintroduce the reported per-download fan-out, because download completions mutate only the per-attachment `ObservableObject` store, but an account switch still invalidates all parent readers. Acceptable, but worth documenting or avoiding if you want the store lookup path fully observation-free.
- **Store creation now happens from view body evaluation**: `whitenoise-mac/Core/WorkspaceState.swift:1714-1721` — `mediaDownloadStateStore(forKey:)` lazily inserts a store when a tile renders, before any load starts. `@ObservationIgnored` prevents an invalidation loop and pruning bounds it to the active conversation, so this is not blocking. Still, side-effecting state allocation from `body` is a SwiftUI smell; a short comment explaining the intentional lazy cache would help.

NITPICK:
- **Test fixtures use identical media hashes for different attachments**: `whitenoise-macTests/whitenoise_macTests.swift:1071-1072` — attachment IDs differ, so the current keying works, but distinct test hashes would make the fixture less fragile if the keying changes later.
- **`update(.idle)` immediately before removing stores is mostly defensive noise**: `whitenoise-mac/Core/WorkspaceState.swift:3473`, `3480` — harmless, but the store is then dropped from the dictionary. Keep it only if the intent is to notify any lingering `@ObservedObject` subscribers during teardown.
- **Consider colocating `MediaDownloadStateStore` with the media state models**: `whitenoise-mac/Core/WorkspaceState.swift:87-95` — minor organization nit; `MediaDownloadState` / `MessageMediaDownload` live in `Models/MessengerModels.swift`.

SUMMARY: The architectural fix is sound: moving the mutable download state out of the single observed `WorkspaceState.mediaDownloads` dictionary and into per-attachment `MediaDownloadStateStore` objects means a completion publishes only to the tile/attachment/gallery view observing that store. The changed per-tile bodies now read `downloadState.state` instead of the shared dictionary, so the reported O(N²) re-render fan-out is actually addressed. No security/privacy regression found; plaintext media handling remains unchanged and main-actor-confined. The PR is blocked on red CI formatting, and the regression test should be strengthened before this is considered done.

SENSITIVE_PATHS: none

agent-p1p:
Pip address-review update:

- Fixed the swift-format blocking finding around `stateStore.update(.loaded(...))`.
- Added the missing `Combine` import required for `ObservableObject` / `@Published` in `WorkspaceState.swift`.
- Strengthened the media-download observation regression test to use a bootstrapped fake runtime, distinct attachment hashes, a real download success path, and a first-store `objectWillChange` guard.
- Added comments documenting lazy per-attachment store allocation and teardown notifications.

Current blocker: GitHub `PR Checks` is still red in `Unit tests`, but not on the media test. It repeatedly fails existing `incrementalChatRowUpdateEnrichesDirectChatTitle()` with:

`title=Alice Actual preview=No messages yet pictureURL=https://example.com/alice.png detailsCalls=1`

Evidence:
- Static Analysis: pass
- swift-format lint: pass
- CodeRabbit: pass / rate-limit notice only
- Failing runs after the review fix: 28114863453, 28115416895, 28115671816, 28115902928

I am not marking this ready. Needs a decision: fix the unrelated chat-list enrichment test/behavior in this PR, or handle it as a separate master fix and rebase this PR afterwards.

PR #149, opened 2026-06-24
fix: clear media downloads on account removal

agent-p1p:
Closes #141

## Summary
- Clear `mediaDownloads` when `removeAccount(_:)` resets the active account state.
- This wipes decrypted attachment plaintext cached for a removed account before the last-account early return or surviving-account reselection path.
- Matches the surrounding cache-clearing behavior and the full-wipe `resetToNewInstallState` path.

## Verification
- `git diff --check` — passed.
- Inspected `.github/workflows/mac-ci.yml`; CI requires macOS/Xcode (`xcrun`, `xcodebuild`).
- Confirmed this Linux host lacks `xcrun`, `xcodebuild`, `swift`, and `swift-format`, so macOS CI commands could not be run locally.

## Sensitive paths
- `whitenoise-mac/Core/WorkspaceState.swift` — account-removal state cleanup for decrypted media plaintext cache.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/149">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearing an account now also removes leftover media download state when the active session is reset, helping prevent stale attachment data from appearing after account removal.

PR #150, opened 2026-06-24
fix: fail transcript export on empty history page

### LINKED ISSUE #139: ConversationTranscriptExport silently truncates history if the FFI returns an empty page with hasMoreBefore == true
## Summary

The transcript‑export pagination loop advances its `before` cursor from the oldest message of each page. If a page ever comes back **empty while `hasMoreBefore` is still true**, `orderedPage.first` is `nil`, the loop `break`s, and the "full" transcript is silently truncated — older history that exists is dropped with no error.

## Where

`whitenoise-mac/Core/ConversationTranscriptExport.swift:92-99`

```swift
let orderedPage = sortChronologically(page.messages)
guard page.hasMoreBefore, let oldest = orderedPage.first else { break }   // empty page + hasMoreBefore => break
let nextBefore = oldest.timelineAt
let nextBeforeMessageId = oldest.messageIdHex
```

## The problem

The loop conflates two distinct conditions into one `break`:
1. `hasMoreBefore == false` → genuinely done.
2. `page.messages.isEmpty && hasMoreBefore == true` → cursor cannot advance, but more history exists.

Case (2) (a page fully filtered server‑side, or a transient empty result) is treated as "done", truncating the export. Because the export is presented to the user as the complete conversation dump (debugging/audit), a silent truncation is misleading and hard to notice.

## Suggested fix

Distinguish the two cases: if `page.messages.isEmpty && page.hasMoreBefore`, treat it as a terminal/error condition explicitly (throw or surface a warning) rather than a normal `break`, or paginate by count/offset rather than by the oldest‑message cursor so an empty page can't stall advancement.

## Severity

LOW — depends on the FFI returning an empty‑but‑more page, but it silently produces an incomplete transcript when it happens.

agent-p1p:
Closes #139

## Summary
- Split the transcript-export pagination stop condition so `hasMoreBefore == false` remains the normal completion path.
- Added `ConversationTranscriptExport.ExportError.emptyPageWithMoreHistory` and throw it when the FFI returns an empty page while still reporting older history, preventing silent partial exports.
- Added focused Swift Testing coverage for the empty-but-more regression and the empty-no-more completion case.

## Verification
- `git diff --check` ✅
- working-tree conflict-marker sweep on touched files ✅
- changed-file line-length check against `.swift-format` 120-char limit ✅
- GitHub `Static Analysis` ✅
- GitHub `PR Checks` ❌, rerun once and failed both times on `whitenoise_macTests.incrementalChatRowUpdateEnrichesDirectChatTitle()`.
  - The same failure is present on current `master` run 28115979319 and sibling PR #149, so this appears to be unrelated master CI drift rather than this transcript-export change.
  - The new transcript-export tests passed in GitHub CI before the unrelated failure.
- Local macOS checks could not run on this Linux worker: `xcrun`, `xcodebuild`, `swift`, and `swift-format` are unavailable. `scripts/ci/macos-sanity-checks.sh` was attempted and stopped at missing `xcodebuild`.

## Sensitive paths
None.

### DIFF
--- a/whitenoise-mac/Core/ConversationTranscriptExport.swift
+++ b/whitenoise-mac/Core/ConversationTranscriptExport.swift
@@ -7,6 +7,22 @@ import MarmotKit
 nonisolated enum ConversationTranscriptExport {
     static let pageLimit: UInt32 = 200
 
+    enum ExportError: LocalizedError {
+        /// The FFI reported more history exists (`hasMoreBefore == true`) but returned an
+        /// empty page, so the `before` cursor cannot advance. Surfacing this prevents
+        /// silently truncating the transcript (issue #139).
+        case emptyPageWithMoreHistory
+
+        var errorDescription: String? {
+            switch self {
+            case .emptyPageWithMoreHistory:
+                return
+                    "Transcript export stopped early: the timeline reported more history but returned an empty page, "
+                    + "so older messages could not be loaded."
+            }
+        }
+    }
+
     struct Document: Encodable {
         var v: Int = 1
         var exportedAt: String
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define ExportError for empty history pages](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/150/chapters/1) |
| 2 | [Fail transcript export on empty history pages](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/150/chapters/2) |
| 3 | [Verify transcript export failure and completion](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/150/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/150">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit ecbc7874da4cb88d42c742e23d4709799be209e4 on Jun 24, 2026 7:31pm UTC.</sub>

PR #159, opened 2026-06-24
fix: guard concurrent group avatar updates

### LINKED ISSUE #134: updateSelectedGroupImage lacks a reentrancy guard, allowing duplicate concurrent group-avatar updates
## Summary

`updateSelectedGroupImage(url:dim:)` sets `isSavingGroupImage = true` but never checks it in its entry `guard`, so a second invocation can re‑enter across the `await` and publish a second `updateGroupAvatarUrl` for the same group. This is the same class of bug fixed for `sendDraft` in #78, but this method was missed.

## Where

`whitenoise-mac/Core/WorkspaceState.swift:2078-2096`

```swift
private func updateSelectedGroupImage(url: String?, dim: String?) async {
    guard let client, let activeAccount, let selectedChat, !selectedChat.isDirect else { return }
    isSavingGroupImage = true                 // set, but NOT part of the guard
    defer { isSavingGroupImage = false }
    do {
        _ = try await client.updateGroupAvatarUrl(   // suspension point
            accountRef: activeAccount.accountRef, groupIdHex: selectedChat.id,
            url: url, dim: dim, thumbhash: nil)
        await reloadChats()
        closeGroupImagePicker()
    } catch { lastError = error.localizedDescription }
}
```

Contrast the sibling methods, which all guard their in‑flight flag:

```swift
acceptGroupInvite:  guard ..., !isAcceptingGroupInvite, !isDecliningGroupInvite else { return }   // 2099
declineGroupInvite: guard ..., !isDecliningGroupInvite, !isAcceptingGroupInvite else { return }   // 2119
```

## The problem

Both public entry points (`setGroupImage` at 2070, `clearGroupImage` at 2074) funnel here, and they are driven by **different** controls (tapping a search‑result tile vs. a separate "Clear" button). `updateGroupAvatarUrl` is an async FFI, so the actor suspends; a second action fired before SwiftUI propagates `isSavingGroupImage` (e.g. set‑then‑immediately‑clear, or a double‑tap) re‑enters and publishes a second conflicting avatar mutation to relays and races two `reloadChats()`.

The UI does apply `.disabled(workspace.isSavingGroupImage)` (MessengerShellView.swift:1673/2119/2144), which narrows the window, but the disable only takes effect after the first `await` and does not cover the set‑vs‑clear cross‑control race — exactly the residual gap #78 closed with an explicit guard.

## Suggested fix

Add the flag to the entry guard:

```swift
guard let client, let activeAccount, let selectedChat, !selectedChat.isDirect, !isSavingGroupImage else { return }
```

## Severity

MEDIUM — duplicate group‑avatar mutation events published to relays (mitigated by UI `.disabled`, but the guard is the actual fix and matches #78).

agent-p1p:
## Summary
- Guard `updateSelectedGroupImage(url:dim:)` with `!isSavingGroupImage` so overlapping set/clear avatar actions return before publishing a second mutation.
- Add a regression test that holds the first fake group-avatar update in-flight, fires an overlapping clear, and asserts only one FFI update is issued.

## Test Plan
- `git diff --check`
- committed-tree conflict-marker sweep for changed Swift files
- line-length scan for changed Swift files
- `xcrun swift-format lint ...` skipped: `xcrun` is not available on this Linux worker
- `xcodebuild test ...` skipped: `xcodebuild` is not available on this Linux worker

Closes #134

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/159">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -2099,7 +2099,6 @@ final class WorkspaceState {
         isGroupImagePickerPresented = false
         invalidateGroupImageSearch()
         groupImageResults = []
-        isSavingGroupImage = false
     }
 
     private func dismissGroupImagePickerIfSelectedChatUnavailable() {
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -2153,7 +2153,12 @@ final class WorkspaceState {
     }
 
     private func updateSelectedGroupImage(url: String?, dim: String?) async {
-        guard let client, let activeAccount, let selectedChat, !selectedChat.isDirect else { return }
+        guard let client,
+            let activeAccount,
+            let selectedChat,
+            !selectedChat.isDirect,
+            !isSavingGroupImage
+        else { return }
         isSavingGroupImage = true
         defer { isSavingGroupImage = false }
 
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -3905,6 +3905,67 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func groupImageUpdateDropsOverlappingDuplicateInvocation() async throws {
+        // Issue #134: set/clear group image both funnel through an async avatar update.
+        // The in-flight flag must guard the entry point itself so a second control action
+        // delivered before SwiftUI disables the UI does not publish a conflicting update.
+        let account = AccountSummaryFfi(
+            label: "Desktop Account",
+            accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
+            localSigning: true,
+            running: true
+        )
+        let runtime = FakeMarmotRuntime(accounts: [account])
+        runtime.installGroup(messageGroup())
+        let state = WorkspaceState(clientFactory: { runtime })
+
+        await state.bootstrap()
+        guard let groupChat = state.activeChats.first else {
+            Issue.record("Expected a group chat")
+            return
+        }
+
+        let result = GroupImageSearchResult(
+            id: "image-1",
+            title: "Aurora",
+            imageURL: "https://example.com/aurora.jpg",
+            thumbnailURL: "https://example.com/aurora-thumb.jpg",
+            creator: "Open Photographer",
+            license: "by",
+            attribution: nil,
+            sourceURL: "https://example.com/aurora",
+            width: 1024,
+            height: 680
+        )
+
+        state.showGroupImagePicker(for: groupChat)
+        runtime.groupAvatarUpdateGateEnabled = true
+        async let firstUpdate: Void = state.setGroupImage(result)
+
+        while !(state.isSavingGroupImage && runtime.didReachGroupAvatarUpdateGate) {
+            await Task.yield()
+        }
+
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Guard concurrent group avatar updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/159/chapters/1) |
| 2 | [Enhance test runtime with avatar update gating](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/159/chapters/2) |
| 3 | [Add regression test for overlapping avatar updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/159/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/159">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 4cc6d898d7ce8ae2e0631219b6449796af99b203 on Jun 24, 2026 8:00pm UTC.</sub>

agent-p1p:
## Pip adversarial review

VERDICT: REQUEST_CHANGES

BLOCKING:
- Residual reentrancy remains if the picker is closed mid-save: `whitenoise-mac/Core/WorkspaceState.swift:2057-2061`, `whitenoise-mac/Core/WorkspaceState.swift:2114-2125`. The new guard depends on `isSavingGroupImage` being a true in-flight flag, but `closeGroupImagePicker()` still clears it while `updateGroupAvatarUrl(...)` can be suspended. The close button is not disabled (`MessengerShellView.swift:2062-2070`), and the header can reopen the picker (`MessengerShellView.swift:1529-1540`). Sequence: tap an image result; first update sets `isSavingGroupImage = true` and suspends in the FFI; close the picker, which sets the flag back to false; reopen/clear or otherwise invoke `clearGroupImage()` while the first update is still in flight; the new guard now passes and publishes a second concurrent avatar mutation. Keep the in-flight state owned by `updateSelectedGroupImage` (for example, do not clear it from normal picker close; let the `defer`/reset path own it, or use a separate private in-flight token) and add a regression for close-then-second-update.

SUGGESTION:
- none

NITPICK:
- `whitenoise-macTests/whitenoise_macTests.swift:3688` — The spin loop waits for both `state.isSavingGroupImage` and `runtime.didReachGroupAvatarUpdateGate`; the gate flag should already imply the first condition in this actor-isolated path. Harmless, just slightly noisier than the sibling #78 tests.
- `whitenoise-macTests/whitenoise_macTests.swift:7216` — `updateGroupAvatarUrlCallCount` counts calls that reached the FFI, not completed calls. That is the right assertion for this regression, but a short comment would prevent future misuse if the helper is reused.

SUMMARY: The direct set-vs-clear duplicate path is guarded and the new regression covers it, but the same in-flight flag can still be cleared by the enabled close control before the FFI returns. That leaves a user-reachable bypass of the guard and the issue is not fully fixed yet.

SENSITIVE_PATHS: none

agent-p1p:
Addressed the adversarial review finding in c1372ec / rebased head 4cc6d89.

- `closeGroupImagePicker()` no longer clears `isSavingGroupImage`; the in-flight flag is now owned by `updateSelectedGroupImage`'s `defer`.
- The regression test now covers close + reopen while the first avatar update is suspended, then verifies an overlapping clear is still dropped.
- Rebased onto current `origin/master` after the first CI run hit the unrelated `closingGroupDetailsInvalidatesInFlightLoad` timeout twice. Fresh Mac CI is green: PR Checks + Static Analysis.

PR #165, opened 2026-06-24
fix: preserve newer read marker on mark-read failure (#128)

### LINKED ISSUE #128: Bug: read-marker rollback on FFI failure can clobber a concurrently-advanced newer marker
## Summary

`markLatestVisibleMessageRead(_:)` optimistically advances the in-memory read marker, then awaits an FFI call, and on failure rolls the marker back **unconditionally**. Because the method can be re-entered (across its `await`) by a later invocation that successfully advances the marker further, the failing invocation's rollback can rewind the marker below the newer value that a concurrent invocation already committed.

## Where

`whitenoise-mac/Core/WorkspaceState.swift:3661-3682`:

```swift
let marker = ReadMarker(sentAt: latest.sentAt, messageId: latest.id)
let previousMarker = lastMarkedReadMarkers[groupIdHex]
guard previousMarker != marker else { return }
guard previousMarker.map({ $0 < marker }) ?? true else { return }
lastMarkedReadMarkers[groupIdHex] = marker          // A: sets M1

do {
    ...
    if let row = try await runOffMain({               // suspension point
        try client.markTimelineMessageRead(...)
    }) {
        await applyChatRow(row, account: account)
    }
} catch {
    lastMarkedReadMarkers[groupIdHex] = previousMarker  // A: unconditional rollback to A's snapshot
    setBackgroundStatus(error.localizedDescription)
}
```

## The problem

The method is `@MainActor` but has `await` points, so two invocations for the same group can interleave. Sequence:

1. Invocation **A**: `previousMarker = M0`, latest = M1; sets slot = **M1**; suspends at `runOffMain` (line 3670).
2. Invocation **B** runs while A is suspended: reads `previousMarker = M1`, latest = M2 (a newer message); sets slot = **M2**; FFI succeeds; slot stays **M2**.
3. A resumes and its FFI throws. The `catch` runs `lastMarkedReadMarkers[groupIdHex] = previousMarker`, i.e. **M0**, clobbering B's committed **M2**.

The rollback assumes the slot still holds A's own `marker`, but never checks that before overwriting.

## Impact

- The marker silently rewinds below an already-committed newer value.
- The next live timeline delta then re-issues `markTimelineMessageRead` for an already-read message (redundant FFI + potential duplicate read receipt), and unread state can transiently resurface.

Narrow timing window and benign-ish effect, hence LOW, but it is a real correctness bug in read-state bookkeeping.

## Suggested fix

Only roll back if the slot still holds this invocation's own marker:

```swift
} catch {
    if lastMarkedReadMarkers[groupIdHex] == marker {
        lastMarkedReadMarkers[groupIdHex] = previousMarker
    }
    setBackgroundStatus(error.localizedDescription)
}
```

## Severity

LOW.

agent-p1p:
## Summary
- preserve the newest in-memory read marker when an older optimistic `markTimelineMessageRead` call fails after a concurrent newer advance
- centralize the failed-mark rollback as compare-and-set semantics on `ReadMarker`
- add a regression test for the failed rollback helper preserving newer/nil slots while rolling back only the marker it installed

## Test Plan
- `git diff --check`
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -testPlan PR -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` (not run locally: `xcodebuild` is unavailable on the Linux fixer host)
- `scripts/ci/macos-sanity-checks.sh` (blocked locally at `xcodebuild: command not found` on the Linux fixer host)
- `xcrun swift-format lint --configuration .swift-format --recursive --parallel --strict whitenoise-mac whitenoise-macTests whitenoise-macUITests` (not run locally: `xcrun` is unavailable on the Linux fixer host)

Closes #128

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/165">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -4405,6 +4405,11 @@ struct ReadMarker: Equatable, Comparable {
         return lhs.messageId < rhs.messageId
     }
 
+    /// Returns the read marker to keep after a failed optimistic advance.
+    ///
+    /// `previous` is the caller's snapshot from before it wrote `attempted`; it
+    /// is not proof of the last committed marker if another mark-read call
+    /// advanced this slot while the caller was suspended.
     static func afterFailedOptimisticAdvance(
         current: ReadMarker?,
         attempted: ReadMarker,
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -3795,7 +3795,11 @@ final class WorkspaceState {
                 await applyChatRow(row, account: account)
             }
         } catch {
-            lastMarkedReadMarkers[groupIdHex] = previousMarker
+            lastMarkedReadMarkers[groupIdHex] = ReadMarker.afterFailedOptimisticAdvance(
+                current: lastMarkedReadMarkers[groupIdHex],
+                attempted: marker,
+                previous: previousMarker
+            )
             setBackgroundStatus(error.localizedDescription)
         }
     }
@@ -4392,13 +4396,21 @@ private enum GroupMemberMutationAction {
     case remove
 }
 
-private struct ReadMarker: Equatable, Comparable {
+struct ReadMarker: Equatable, Comparable {
     let sentAt: Date
     let messageId: String
 
     static func < (lhs: ReadMarker, rhs: ReadMarker) -> Bool {
         if lhs.sentAt != rhs.sentAt { return lhs.sentAt < rhs.sentAt }
         return lhs.messageId < rhs.messageId
+    }
+
+    static func afterFailedOptimisticAdvance(
+        current: ReadMarker?,
+        attempted: ReadMarker,
+        previous: ReadMarker?
+    ) -> ReadMarker? {
+        current == attempted ? previous : current
     }
 }
 
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -1717,6 +1717,43 @@ struct whitenoise_macTests {
         #expect(state.selectedMessageIDs == ["m2", "stream"])
     }
 
+    @Test func failedReadMarkerRollbackPreservesConcurrentNewerAdvance() {
+        let previous = ReadMarker(
+            sentAt: Date(timeIntervalSince1970: 1_700_000_000),
+            messageId: "m0"
+        )
+        let attempted = ReadMarker(
+            sentAt: Date(timeIntervalSince1970: 1_700_000_010),
+            messageId: "m1"
+        )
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Expose and enhance ReadMarker with rollback logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/165/chapters/1) |
| 2 | [Apply safe rollback logic to mark-read failures](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/165/chapters/2) |
| 3 | [Test ReadMarker rollback preservation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/165/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/165">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 47b50de4646a90670f63dc980b51effdbd8b2048 on Jun 24, 2026 9:40pm UTC.</sub>

agent-p1p:
Pip adversarial review for marmot-protocol/whitenoise-mac#165.

BLOCKING:
- none

SUGGESTION:
- Reversed overlapping-failure interleaving can still retain an uncommitted marker: `whitenoise-mac/Core/WorkspaceState.swift:3797-3804`. This PR fixes the issue's primary sequence: an older failure no longer overwrites a newer successfully-advanced marker. There is still a narrower adjacent case: A sets M1, B sets M2, A fails first and preserves M2, then B fails and rolls back to its captured `previousMarker` M1. Final cache value is M1 even though both FFI calls failed and only M0 is known committed. I filed follow-up #166 for this instead of blocking this PR.

NITPICK:
- `whitenoise-macTests/whitenoise_macTests.swift:1720-1756`: the regression test is meaningful for the pure helper, but the production bug is async re-entrancy. An integration-style test with two overlapping `markLatestVisibleMessageRead` calls would guard the real path more directly.
- `whitenoise-mac/Core/WorkspaceState.swift:4408-4414`: consider a short comment noting that `previous` is the caller's optimistic snapshot, not necessarily committed state.

SUMMARY: The patch is minimal and correctly preserves a newer concurrently-advanced marker when an older optimistic mark-read call fails. CI is green (`PR Checks`, `Static Analysis`, CodeRabbit status). No sensitive paths touched. The remaining overlapping-failure edge is narrower and tracked separately in #166, so it does not block this PR.

VERDICT: APPROVE
SENSITIVE_PATHS: none

agent-p1p:
Address-review update: pushed 47b50de to document that the rollback helper's `previous` marker is this call's optimistic snapshot, not proof of committed state. The remaining integration-test nit is non-blocking; the adjacent overlapping-failure edge is already tracked in #166. Current PR checks are green.

PR #167, opened 2026-06-24
fix: cancel unused coalesced remote image loads

### LINKED ISSUE #126: RemoteImageLoader: coalescing wraps the download in an unstructured Task, defeating the documented scroll/navigation cancellation
## Summary

The in-flight coalescing layer added to `RemoteImageLoader` runs each download inside an **unstructured `Task {}`**, which does not inherit cancellation from the caller. As a result the per-download cancellation that `CappedImageDownloadDelegate` documents (and that the old non-coalesced path actually had) no longer works: scrolling past an avatar, or navigating away, no longer cancels its in-flight download. Off-screen image downloads run to completion regardless.

## Where

`whitenoise-mac/Core/RemoteImageLoader.swift:346-366`:

```swift
func image(for url: URL, maxPixelSize: CGFloat) async -> LoadedImage? {
    ...
    let (task, owner) = await inFlight.task(for: key) { [self] in
        Task { [self] in                                  // <-- unstructured top-level Task
            await loadImage(for: url, cacheKey: key, maxPixelSize: maxPixelSize)
        }
    }
    let loaded = await task.value                          // awaiting does NOT cancel `task`
    if owner {
        await inFlight.removeTask(for: key)
    }
    return loaded
}
```

The download itself installs a cancellation handler and relies on the surrounding task being cancelled (`RemoteImageLoader.swift:496-523`):

```swift
func download(_ url: URL, using session: URLSession) async -> Data? {
    // "...runs this inside a `.task(id:)`, which cancels the awaiting task whenever the
    //  row's URL/size identity changes (scrolling, navigation); without this the request
    //  would keep running and buffering until it completed or timed out."
    await withTaskCancellationHandler {
        ...
    } onCancel: {
        cancel()
    }
}
```

## The problem

`DownsampledAsyncImage` calls `image(for:)` from a SwiftUI `.task(id:)` (`RemoteImageLoader.swift:276-282`), which cancels its task when the row's URL/size identity changes. That cancellation propagates down structured `await`s — **but** `Task { ... }` created inside `RemoteImageLoadRegistry.task(for:create:)` is a new top-level task and is **not** a child of the caller. Swift cancellation does not propagate into it.

Concretely:
- The view's `.task` is cancelled on scroll/navigation.
- `image(for:)`'s `await task.value` is non-throwing (`Task<LoadedImage?, Never>`), so cancelling the awaiting task does not interrupt it and does not cancel the unstructured download task.
- The unstructured task — the one that actually owns the `withTaskCancellationHandler` in `download(_:using:)` — is never cancelled, so `onCancel` / `cancel()` never fires.

So the explicit guarantee in the `CappedImageDownloadDelegate.download` comment ("without this the request would keep running and buffering until it completed or timed out") is exactly what now happens, for every image, because all loads go through the coalescing registry.

## Impact

- Fast-scrolling the chat list or settings avatars keeps downloading every avatar that scrolled past, until each completes (or hits the 8 MiB cap / timeout) — wasted bandwidth, sockets, CPU, and battery.
- Regression vs. the pre-coalescing behavior (#79 fix) and vs. the in-code documentation, which both promise cancellation on scroll.

It is bounded (8 MiB cap, results are cached), so it is not a leak, but it negates a documented optimization on the hottest image path.

## Suggested fix

Coalescing and cancellation are not mutually exclusive, but they need care (cancelling because *one* waiter left would harm other waiters). Options:
- Track waiter count per key and cancel the underlying download only when the last waiter's `await` is cancelled (use `withTaskCancellationHandler` around `await task.value` to decrement / trigger cancel).
- Or make the registry's task a structured child via a `TaskGroup`/`async let` owned by a long-lived loader actor with explicit cancellation when the in-flight set has no remaining waiters.

At minimum, update the `CappedImageDownloadDelegate.download` comment if cancellation is intentionally dropped, so the code no longer claims a guarantee it does not provide.

## Severity

MEDIUM — performance/resource regression on the primary image path; defeats documented behavior.

(Distinct from #79/#108/#98, which are about coalescing existing/absent, connection reuse, and per-byte streaming respectively — none cover the lost cancellation introduced by the coalescing wrapper.)

agent-p1p:
## Summary
- track waiters for coalesced remote image loads
- cancel the shared download task when the last waiter is cancelled or completes
- keep cancelled waiters from applying a coalesced image result after cancellation
- add a regression test covering last-waiter cancellation while preserving coalescing

## Test Plan
- `git diff --check`
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -destination 'platform=macOS' -only-testing:whitenoise-macTests/whitenoise_macTests/remoteImageLoaderCancelsCoalescedDownloadAfterLastWaiterCancels` (not run locally: `xcodebuild` is not installed on this Linux worker)

Closes #126

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/RemoteImageLoader.swift
+++ b/whitenoise-mac/Core/RemoteImageLoader.swift
@@ -291,23 +291,71 @@ struct DownsampledAsyncImage<Content: View, Placeholder: View>: View {
 /// Shared image cache + downsampling pipeline. `NSCache` is thread-safe, so this needs no
 /// actor for decoded-image storage; in-flight work is coordinated by `RemoteImageLoadRegistry`
 /// so concurrent views that need the same URL/size await one download and decode.
-private actor RemoteImageLoadRegistry {
-    private var tasks: [String: Task<LoadedImage?, Never>] = [:]
+private final class RemoteImageLoadRegistry: @unchecked Sendable {
+    private struct Entry {
+        let task: Task<LoadedImage?, Never>
+        var waiters: Int
+    }
+
+    private let lock = NSLock()
+    private var tasks: [String: Entry] = [:]
 
     func task(
         for key: String,
         create: @Sendable () -> Task<LoadedImage?, Never>
-    ) -> (task: Task<LoadedImage?, Never>, owner: Bool) {
-        if let task = tasks[key] {
-            return (task, false)
+    ) -> Task<LoadedImage?, Never> {
+        lock.lock()
+        defer { lock.unlock() }
+
+        if var entry = tasks[key] {
+            entry.waiters += 1
+            tasks[key] = entry
+            return entry.task
         }
+
         let task = create()
-        tasks[key] = task
-        return (task, true)
+        tasks[key] = Entry(task: task, waiters: 1)
+        return task
     }
 
-    func removeTask(for key: String) {
-        tasks[key] = nil
+    func releaseWaiter(for key: String) {
+        var taskToCancel: Task<LoadedImage?, Never>?
+
+        lock.lock()
+        if var entry = tasks[key] {
+            entry.waiters -= 1
+            if entry.waiters <= 0 {
+                tasks[key] = nil
+                taskToCancel = entry.task
+            } else {
+                tasks[key] = entry
+            }
+        }
+        lock.unlock()
+
+        taskToCancel?.cancel()
+    }
+}
+
+private final class RemoteImageLoadWaiter: @unchecked Sendable {
+    private let lock = NSLock()
+    private var released = false
+    private let onRelease: @Sendable () -> Void
+
+    init(onRelease: @escaping @Sendable () -> Void) {
+        self.onRelease = onRelease
+    }
+
+    func release() {
+        lock.lock()
+        guard !released else {
+            lock.unlock()
+            return
+        }
+        released = true
+        lock.unlock()
+
+        onRelease()
     }
 }
 
@@ -364,16 +412,22 @@ nonisolated final class RemoteImageLoader: @unchecked Sendable {
             return LoadedImage(nsImage: cached)
         }
 
-        let (task, owner) = await inFlight.task(for: key) { [self] in
+        let task = inFlight.task(for: key) { [self] in
             Task { [self] in
                 await loadImage(for: url, cacheKey: key, maxPixelSize: maxPixelSize)
             }
         }
-        let loaded = await task.value
-        if owner {
-            await inFlight.removeTask(for: key)
+        let waiter = RemoteImageLoadWaiter { [inFlight] in
+            inFlight.releaseWaiter(for: key)
         }
-        return loaded
+
+        return await withTaskCancellationHandler {
+            let loaded = await task.value
+            waiter.release()
+            return Task.isCancelled ? nil : loaded
+        } onCancel: {
+            waiter.release()
+        }
     }
 
     private static func cacheKey(for url: URL, maxPixelSize: CGFloat) -> String {
--- a/whitenoise-mac/Core/RemoteImageLoader.swift
+++ b/whitenoise-mac/Core/RemoteImageLoader.swift
@@ -288,9 +288,9 @@ struct DownsampledAsyncImage<Content: View, Placeholder: View>: View {
     }
 }
 
-/// Shared image cache + downsampling pipeline. `NSCache` is thread-safe, so this needs no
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Track and manage image load waiters](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167/chapters/1) |
| 2 | [Wire cancellation handlers to image loading](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167/chapters/2) |
| 3 | [Enhance test stub for cancellation tracking](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167/chapters/3) |
| 4 | [Verify coalesced cancellation with regression test](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/167">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 66be87bcd1c0b5f0a8f0b9cec542a5fbc79befda on Jun 24, 2026 9:51pm UTC.</sub>

agent-p1p:
## Pip adversarial review

BLOCKING:
- Regression test has a coalescing race and can flake: `whitenoise-macTests/whitenoise_macTests.swift:4761-4769` — The test only waits for `requestCount() == 1` before cancelling `first`, but that only proves one load reached `URLProtocol.startLoading()`. It does not prove `second` has registered with `RemoteImageLoadRegistry` and incremented the waiter count. If `first` owns the download and `second` has not entered `image(for:)` yet, `first.cancel()` drops the waiter count to zero, removes the entry, cancels the underlying request, and makes the `stopLoadingCount() == 0` assertion fail. This is the invariant the regression test is meant to protect, so it needs a deterministic synchronization point before the intermediate-waiter cancellation assertion.

SUGGESTION:
- `create()` runs while the registry lock is held: `whitenoise-mac/Core/RemoteImageLoader.swift:316` — Today the closure only allocates `Task { ... }`, so this is not a deadlock, but the lock now covers an arbitrary `@Sendable () -> Task` closure. A comment constraining `create` to cheap/non-reentrant work, or restructuring so only the dictionary mutation is under lock, would reduce future foot-guns.

NITPICK:
- Registry doc wording is slightly stale: `whitenoise-mac/Core/RemoteImageLoader.swift:291-293` — The `NSCache`/"needs no actor" wording now sits above a lock-based registry class, so it reads oddly. Minor wording refresh.
- `RemoteImageLoadWaiter` could use a short comment: `whitenoise-mac/Core/RemoteImageLoader.swift:340-360` — The one-shot latch is functionally needed because `onCancel` and the operation body can both call `release()`. A note would make the concurrency intent easier to audit.
- `RemoteImageURLProtocolStub.stopLoading()` takes `Self.lock` twice in sequence: `whitenoise-macTests/whitenoise_macTests.swift:8566-8571` — Harmless, but one critical section would be cleaner.

SUMMARY: The production fix appears to restore issue #126's intended behavior: each waiter releases exactly once, the shared task is cancelled only when the last waiter leaves, and cancellation then propagates into `CappedImageDownloadDelegate` and the `URLSessionDataTask`. Coalescing remains intact for active waiters, cancelled waiters discard their eventual result, and the remote-image security/privacy policy, redirect validation, byte cap, and ephemeral cache configuration are untouched. The merge blocker is test determinism: the new regression test can cancel the owner before the second waiter has actually joined, producing a false failure and weakening confidence in the guard.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: `whitenoise-mac/Core/RemoteImageLoader.swift` (repo-sensitive remote image/network loading path; cancellation/resource behavior changed)

agent-p1p:
## Pip address-review update

Addressed the adversarial review finding in `66be87b`:
- Made the cancellation regression test wait for the second coalesced waiter to register before cancelling the first waiter.
- Added a DEBUG-only waiter-count observation helper for deterministic synchronization.
- Cleaned up the registry/one-shot waiter comments and collapsed the URLProtocol stub stop counter into one lock section.

Verification:
- `xcrun swift-format lint`, `xcodebuild` targeted test, and `scripts/ci/macos-sanity-checks.sh` are not runnable on this Linux worker (`xcrun`/`xcodebuild` missing).
- GitHub Mac CI is green: PR Checks and Static Analysis passed on run 28131951535.
- CodeRabbit status is green, but the bot comment is still a rate-limit notice, not a substantive review.

PR #168, opened 2026-06-24
fix: track confirmed read markers for rollback (#166)

### LINKED ISSUE #166: Bug: overlapping read-marker failures can retain an uncommitted marker
## Summary

The compare-and-set rollback in `markLatestVisibleMessageRead(_:)` prevents an older failed mark-read call from overwriting a newer concurrently-advanced marker. However, overlapping failures can still leave the in-memory read marker at a value that was never committed by FFI.

## Sequence

Starting from committed marker `M0`:

1. Invocation A sets the optimistic slot to `M1` and suspends in `markTimelineMessageRead`.
2. Invocation B observes `previousMarker = M1`, sets the optimistic slot to `M2`, and suspends.
3. A's FFI call fails first. Because the current slot is `M2`, A does not roll back.
4. B's FFI call also fails. Because the current slot is `M2`, B rolls back to its captured `previousMarker`, which is `M1`.

Final slot: `M1`, even though both FFI calls failed and the only known committed marker is `M0`.

## Impact

This is narrower than #128: it requires two overlapping failures, not a newer success. But the cache can still suppress a future retry for `M1` if `M1` remains the latest visible message, because `previousMarker == marker` causes an early return.

## Suggested fix

Track committed/read-confirmed marker state separately from optimistic in-flight markers, or make rollback restore only to a value known to have been confirmed rather than each invocation's optimistic snapshot.

## Related

Found during adversarial review of #165.

agent-p1p:
## Summary
- track confirmed read markers separately from optimistic in-flight read markers
- roll failed optimistic read-marker advances back only to the last FFI-confirmed marker
- keep the optimistic gate monotonic when older successes race with newer in-flight marks
- update read-marker regression coverage for overlapping failure/success ordering

## Test Plan
- `git diff --check`
- `scripts/ci/macos-sanity-checks.sh` (blocked locally: `xcodebuild: command not found` on the Linux fixer host)
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -testPlan PR -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` (not run locally: `xcodebuild` is unavailable on the Linux fixer host)
- `xcrun swift-format lint --configuration .swift-format --recursive --parallel --strict whitenoise-mac whitenoise-macTests whitenoise-macUITests` (not run locally: `xcrun` is unavailable on the Linux fixer host)

Closes #166

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -379,6 +379,7 @@ final class WorkspaceState {
     /// underlying messages change.
     private var messageIDsByChat: [String: [String]] = [:]
     private var lastMarkedReadMarkers: [String: ReadMarker] = [:]
+    private var lastConfirmedReadMarkers: [String: ReadMarker] = [:]
     private var deliveredNotificationKeys = Set<String>()
     private var deliveredNotificationKeyOrder: [String] = []
     private var newChatLookupGeneration = 0
@@ -2807,6 +2808,7 @@ final class WorkspaceState {
         timelinePagingByChat = [:]
         timelineInitialLoadGroupId = nil
         lastMarkedReadMarkers = [:]
+        lastConfirmedReadMarkers = [:]
         deliveredNotificationKeys = []
         deliveredNotificationKeyOrder = []
         UserDefaults.standard.removeObject(forKey: Self.activeAccountKey)
@@ -3430,6 +3432,7 @@ final class WorkspaceState {
             timelineInitialLoadGroupId = nil
         }
         lastMarkedReadMarkers[groupIdHex] = nil
+        lastConfirmedReadMarkers[groupIdHex] = nil
 
         guard case .chat(let selectedGroupId) = selection,
             selectedGroupId == groupIdHex
@@ -3777,28 +3780,36 @@ final class WorkspaceState {
             return
         }
         let marker = ReadMarker(sentAt: latest.sentAt, messageId: latest.id)
-        let previousMarker = lastMarkedReadMarkers[groupIdHex]
-        guard previousMarker != marker else { return }
-        guard previousMarker.map({ $0 < marker }) ?? true else { return }
+        let currentMarker = lastMarkedReadMarkers[groupIdHex]
+        guard currentMarker != marker else { return }
+        guard currentMarker.map({ $0 < marker }) ?? true else { return }
         lastMarkedReadMarkers[groupIdHex] = marker
 
         do {
             let accountRef = account.accountRef
             let messageId = latest.id
-            if let row = try await runOffMain({
+            let row = try await runOffMain({
                 try client.markTimelineMessageRead(
                     accountRef: accountRef,
                     groupIdHex: groupIdHex,
                     messageIdHex: messageId
                 )
-            }) {
+            })
+            let committedState = ReadMarker.afterSuccessfulCommit(
+                current: lastMarkedReadMarkers[groupIdHex],
+                confirmed: lastConfirmedReadMarkers[groupIdHex],
+                attempted: marker
+            )
+            lastMarkedReadMarkers[groupIdHex] = committedState.current
+            lastConfirmedReadMarkers[groupIdHex] = committedState.confirmed
+            if let row {
                 await applyChatRow(row, account: account)
             }
         } catch {
             lastMarkedReadMarkers[groupIdHex] = ReadMarker.afterFailedOptimisticAdvance(
                 current: lastMarkedReadMarkers[groupIdHex],
                 attempted: marker,
-                previous: previousMarker
+                confirmed: lastConfirmedReadMarkers[groupIdHex]
             )
             setBackgroundStatus(error.localizedDescription)
         }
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Update ReadMarker logic for confirmed state](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168/chapters/1) |
| 2 | [Track confirmed read markers in WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168/chapters/2) |
| 3 | [Implement rollback and commit logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168/chapters/3) |
| 4 | [Verify read marker rollback and commit behavior](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/168">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 7b96822d71f3efc174dd9e985db9633277e0187e on Jun 24, 2026 10:59pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Reviewed PR #168 against issue #166.

Checks observed:
- `git diff --check origin/master...HEAD` passed locally.
- Local macOS tests were not runnable on this Linux host (`swift`, `xcodebuild`, and `xcrun` are unavailable).
- GitHub PR Checks and Static Analysis are currently passing.

BLOCKING:
- none

SUGGESTION:
- Post-await teardown can resurrect cleared marker slots: `whitenoise-mac/Core/WorkspaceState.swift:3798-3804` — `markLatestVisibleMessageRead` guards `activeAccountId`/`selectedChat` only at the top (line 3766), not after `await runOffMain`. If `removeChat` (3434-3435), an account switch, or `reset` (2810-2811) clears `lastMarkedReadMarkers`/`lastConfirmedReadMarkers` while the FFI call is suspended, the success path unconditionally writes both slots back via `afterSuccessfulCommit` (and the failure path writes via `afterFailedOptimisticAdvance`). After a reset, `current`/`confirmed` are both nil, so `afterSuccessfulCommit` returns `(marker, marker)` — re-populating a stale entry for a chat that is no longer selected. This is pre-existing behavior (the old code also did `lastMarkedReadMarkers[groupIdHex] = marker`), but the PR now writes a second dictionary and so slightly widens the stale-write surface. A re-guard after the await (mirroring the `guard activeAccountId == account.id, selectedChat?.id == groupIdHex` checks used in `applyTimelinePageProjection` at 1655/1662) would close it. Worth addressing given the issue is explicitly about cache entries suppressing future retries.

- Test coverage is unit-only at the `ReadMarker` helper level: `whitenoise-macTests/whitenoise_macTests.swift:1750-1831` — the two new tests exercise `afterFailedOptimisticAdvance`/`afterSuccessfulCommit` in isolation, which is good, but there is no integration-level test that drives two genuinely overlapping `markLatestVisibleMessageRead` invocations through the mock `MarmotRuntime` (with one or both failing) to assert the end-to-end slot state described in #166. The existing `olderTimelineProjectionDeltaDoesNotMoveReadMarkerBackward` test (1834) shows the harness can do this. Adding the overlapping-failure sequence as an actor-level test would protect against regressions in the call-site wiring (e.g. a future refactor reordering the `current`/`confirmed` reads at 3799-3800), which the helper-level tests cannot catch.

NITPICK:
- `afterSuccessfulCommit` return type uses a bare tuple `(current:, confirmed:)`: `whitenoise-mac/Core/WorkspaceState.swift:4438-4447` — fine and clear at this scale, but the parallel naming with the stored `lastMarkedReadMarkers`/`lastConfirmedReadMarkers` is slightly indirect ("current" maps to "lastMarked"). Not worth changing, but a one-line doc note tying tuple field names to the storage slots would aid future readers.

- `latest(_:_:)` helper reimplements a max-with-optional: `whitenoise-mac/Core/WorkspaceState.swift:4449-4452` — correct, and the tie/equality case (`marker == candidate` returns `candidate`) is harmless since they're equal. No change needed; just noting it duplicates logic that `Comparable` could express, but the explicit form is arguably clearer here.

SUMMARY: The fix is correct and directly resolves #166. Separating an FFI-confirmed marker (`lastConfirmedReadMarkers`) from the optimistic in-flight gate (`lastMarkedReadMarkers`), and rolling failed advances back to the confirmed value rather than each caller's optimistic snapshot, eliminates the "rollback to an uncommitted M1" path. Tracing the exact #166 sequence (two overlapping failures from committed M0) now yields a final slot of M0, not M1, so future retries are no longer suppressed. The monotonic `afterSuccessfulCommit` correctly preserves a newer in-flight optimistic marker while still recording the committed value, and handles older-success-after-newer-commit. Since `WorkspaceState` is `@MainActor`, the only interleaving points are the `runOffMain` awaits, and the synchronous read/modify of both dictionaries around those awaits is sound. The changes are tightly scoped with no scope creep, naming follows existing conventions, and teardown paths clear the new dictionary. The main residual concern is the pre-existing lack of a post-await active-account/selected-chat re-guard, which the new second dictionary write marginally widens; and that the regression coverage stays at the helper level rather than asserting the end-to-end overlapping-failure sequence. Neither is blocking.
VERDICT: APPROVE
SENSITIVE_PATHS: none

---
Reviewed by Pip / cursor-reviewer. Cursor usage: inputTokens=14, outputTokens=4994, cacheReadTokens=244181, cacheWriteTokens=32009.

agent-p1p:
## Pip address-review update

Addressed the non-blocking post-await stale-write suggestion in 7b96822:
- `markLatestVisibleMessageRead` now re-checks `activeAccountId` and `selectedChat` after `runOffMain` returns before writing read-marker cache state on either success or failure.
- Added a doc note tying `afterSuccessfulCommit` tuple fields back to the two marker dictionaries.

Checks:
- `scripts/ci/macos-sanity-checks.sh` cannot run on this Linux host: `xcodebuild: command not found`.
- GitHub Mac CI is green: PR Checks, Static Analysis, CodeRabbit.

I left the integration-test suggestion as follow-up-level only; the existing helper-level regression remains, and the PR is now marked ready for review.

PR #172, opened 2026-06-25
fix: avoid chat-list resort for row updates

### LINKED ISSUE #170: Performance: applyChatRow re-sorts the whole chat list and spawns an enrichment task on every delta, including read-state-only updates
## Summary

`applyChatRow(_:account:)` is the funnel for every chat-list mutation (subscription deltas *and* read-marker commits). On each call it (a) re-sorts the **entire** chat list with `sortedChatItems` and reassigns `chatsByAccount[...]`, and (b) unconditionally spawns a chat-list **enrichment task** via `startChatListEnrichment`. Both run on the main actor, per event — even when the update changes only unread/preview state and the enriched metadata (title/avatar/isDirect) is unchanged.

## Where

```swift
// applyChatRow(_:account:)  (3384-3411)
if let index = chats.firstIndex(where: { $0.id == chat.id }) {
    chats[index] = chat
} else {
    chats.append(chat)
}
chatsByAccount[account.id] = sortedChatItems(chats)          // 3409 — full O(n log n) re-sort per delta
startChatListEnrichment(rows: [row], account: account, replacingCurrent: false)   // 3410 — task spawn per delta
```

The read-marker path funnels straight into it on every advance:

```swift
// markLatestVisibleMessageRead(...)  (3806-3808)
if let row {
    await applyChatRow(row, account: account)               // every kind-9 read advance → applyChatRow
}
```

## The problem

While a conversation is open and active, **every** incoming kind-9 message advances the read marker (`markLatestVisibleMessageRead`), and each advance calls `markTimelineMessageRead` and then `applyChatRow` with the returned row. For a busy chat that means, per message:

- a full `sortedChatItems(chats)` re-sort of the whole list (O(n log n)) plus a full reassignment of `chatsByAccount[account.id]` (triggering a SwiftUI list diff), for a single-row change; and
- an enrichment ownership token + a spawned `Task` (`startChatListEnrichment` → `chatListRowEnrichment.register`/`finishTask` bookkeeping), even though a read-state-only update never changes the fields enrichment resolves (title, avatar, isDirect).

The same full-resort-per-event also applies to ordinary chat-list subscription deltas (`unreadChanged`, `newLastMessage`, etc.): a one-row change re-sorts and re-diffs the whole list. For an account with many conversations receiving frequent deltas this is repeated O(n log n) main-actor work.

This is distinct from the already-filed redundant-FFI work (#9) and the untracked/uncancellable enrichment tasks on account switch (#40): those concern the *FFI fetches* and *cancellation*; this is about the unconditional full re-sort and the enrichment-task spawn for updates that don't need either.

## Suggested fix

- Skip enrichment when the applied row came from a read-state-only update (e.g. add a flag to `applyChatRow` so the `markTimelineMessageRead` / `initializeChatReadState` callers suppress `startChatListEnrichment`, since they never change enriched metadata).
- When the changed row's sort key (`updatedAt`) is unchanged, update in place without a full re-sort; otherwise insert at the correct position via binary search rather than re-sorting the whole array.

## Severity

LOW (performance) — repeated O(n log n) main-actor work plus per-event task churn on the read-marking and chat-list-delta hot paths; functionally correct today.

agent-p1p:
## Summary
- Replace per-row full chat-list sorting with a single-row upsert that updates in place when ordering is unchanged and uses binary insertion when `updatedAt`/title ordering changes.
- Suppress chat-list enrichment for read-state-only rows returned by `initializeChatReadState` and `markTimelineMessageRead`.
- Preserve already-resolved direct-chat metadata when applying read-state-only rows so skipping enrichment does not regress titles/avatars.
- Add regression coverage for single-row ordering updates, binary insertion, tie-breaker moves, and metadata preservation.

## Verification
- `git diff --check`
- changed Swift line-length scan (<=120 chars)
- conflict-marker scan

Not run locally: `xcrun swift-format`, `xcodebuild test`, `xcodebuild build`, and `xcodebuild analyze` are unavailable on this Linux worker. GitHub macOS CI should run the repo's PR checks.

Closes #170

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -503,6 +503,7 @@ final class WorkspaceState {
     /// that membership slice and invalidate it on membership-changing subscription events.
     private var groupMemberDetailsCache: [String: [GroupMemberDetailsFfi]] = [:]
     private var groupMemberDetailsLookups: [String: GroupMemberDetailsLookup] = [:]
+    private var readStateMetadataEnrichmentAttempts = Set<String>()
     private var nextGroupMemberDetailsLookupToken: UInt64 = 0
 
     /// How long a *complete* peer-profile lookup is trusted before it is re-resolved
@@ -3473,7 +3474,7 @@ final class WorkspaceState {
     ) async {
         guard activeAccountId == account.id else { return }
 
-        var chats = chatsByAccount[account.id] ?? []
+        let chats = chatsByAccount[account.id] ?? []
         if row.archived {
             removeChat(groupIdHex: row.groupIdHex, account: account)
             return
@@ -3494,7 +3495,7 @@ final class WorkspaceState {
             // direct chats to flicker back to their raw group-id fallback.
             chat = ChatListOrdering.preservingResolvedMetadata(in: chat, from: current)
         }
-        let needsInitialMetadata = readStateRowNeedsMetadataEnrichment(row, current: current)
+        let needsInitialMetadata = !shouldEnrich && readStateRowNeedsMetadataEnrichment(row, current: current)
 
         chatsByAccount[account.id] = ChatListOrdering.upserting(chat, into: chats)
         if shouldEnrich {
@@ -3502,9 +3503,9 @@ final class WorkspaceState {
         } else if needsInitialMetadata {
             // A read-state row normally does not need enrichment, but the first selected-chat
             // read-state row can arrive after reload wiring cancels the snapshot enrichment.
-            // Resolve that one missing membership cache synchronously so direct-chat metadata
-            // and timeline sender-name fallbacks are populated without leaving background work
-            // that can race later group-details tests or UI actions.
+            // Resolve at most one missing membership cache per invalidation; repeated failures
+            // must not put groupDetails/userProfile work back on every read-marker advance.
+            readStateMetadataEnrichmentAttempts.insert(row.groupIdHex)
             await enrichChatRows([row], account: account)
         }
     }
@@ -3514,11 +3515,15 @@ final class WorkspaceState {
     }
 
     private func readStateRowNeedsMetadataEnrichment(_ row: ChatListRowFfi, current: ChatItem?) -> Bool {
+        if readStateMetadataEnrichmentAttempts.contains(row.groupIdHex) { return false }
         if groupMemberDetailsCache[row.groupIdHex] == nil { return true }
         guard let current else { return true }
         guard !current.isDirect else { return false }
         let rawTitle = row.title.trimmingCharacters(in: .whitespacesAndNewlines)
         let groupNameIsBlank = row.groupName.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        // Blank-name group rows use the raw row title until one enrichment pass can confirm
+        // whether this is a direct chat. Retry once after invalidation, then leave later
+        // read-state-only rows on the hot path without another synchronous metadata lookup.
         return groupNameIsBlank && current.title == rawTitle
     }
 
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce ChatListOrdering utility for efficient updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172/chapters/1) |
| 2 | [Implement incremental chat list upserting](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172/chapters/2) |
| 3 | [Suppress enrichment for read-state updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172/chapters/3) |
| 4 | [Add regression tests for ordering and enrichment](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/172">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit af7412f92561e92ab3beac394d98d37d10f6b49b on Jun 25, 2026 5:13am UTC.</sub>

agent-p1p:
Pip adversarial review for PR #172.

BLOCKING:
- Read-state-only rows can still perform inline enrichment on every read-marker advance when metadata resolution does not cache successfully: `whitenoise-mac/Core/WorkspaceState.swift:3497-3508` and `3516-3522`. `readStateRowNeedsMetadataEnrichment` returns true whenever `groupMemberDetailsCache[row.groupIdHex] == nil`; `cachedGroupMembers` only populates that cache when `groupDetails` succeeds with non-nil members. If the group is not yet synced or `groupDetails` keeps failing/returning nil, every kind-9 read advance re-enters `await enrichChatRows([row], ...)`, which can do `groupDetails` and `userProfile` inline on the read-marker path. That partially reintroduces the hot-path work #170 is trying to remove, and makes it synchronous instead of background/coalesced. Please add an attempted/negative-cache/backoff guard, or keep this fallback off the read-marker continuation.
- `ChatListOrdering.upserting` now assumes the existing `chatsByAccount` array is sorted, but `insertCreatedChatIfNeeded` still violates that invariant by inserting a new `ChatItem(updatedAt: nil)` at index 0: `whitenoise-mac/Core/WorkspaceState.swift:4149-4168`. Per the comparator, nil `updatedAt` sorts after dated chats. Previously the next `applyChatRow` full sort repaired that array; now binary insertion and neighbor checks operate on the unsorted array, so unrelated chat-list deltas arriving before the created chat's own row can be placed incorrectly. Either insert the optimistic chat with the same ordering helper, explicitly preserve/document a pinned optimistic-row invariant, or restore sortedness before using incremental upserts.

SUGGESTION:
- Add integration coverage for the new runtime behavior, not only the pure `ChatListOrdering` helpers. In particular: repeated read-marker advances with missing group metadata should keep `groupDetails` calls bounded, and a direct-chat read-state update through `applyChatRow(shouldEnrich: false)` should preserve resolved title/avatar/isDirect while updating unread/preview state.

NITPICK:
- `whitenoise-mac/Core/WorkspaceState.swift:3476` — `var chats` can be `let chats` after the refactor.
- `readStateRowNeedsMetadataEnrichment` is computed even when `shouldEnrich == true`, where the result is discarded. Move it inside the `else if !shouldEnrich` path.
- The `groupNameIsBlank && current.title == rawTitle` heuristic is a fragile proxy for “unresolved title”; if it stays, add a short comment explaining why one extra resolve is acceptable.

SUMMARY: The single-row ordering helper looks logically sound for sorted inputs, and preserving resolved direct-chat metadata avoids the obvious flicker regression when read-state rows skip enrichment. The remaining issue is that the fallback metadata path can still put repeated FFI work on the read-marker hot path in failure/missing-cache cases, and the new incremental ordering path makes an existing unsorted optimistic insertion load-bearing.

VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: none

Local/CI checks observed during review:
- `git diff --check origin/master...HEAD` passed
- changed Swift line-length scan passed
- `bash -n scripts/ci/macos-sanity-checks.sh` passed
- GitHub checks are green: CodeRabbit, PR Checks, Static Analysis

agent-p1p:
Pip address-review update for PR #172.

Addressed the adversarial review blockers:
- bounded read-state metadata fallback with a per-group attempt guard, reset on membership-cache invalidation; repeated read-marker advances with failed/missing group metadata no longer re-run groupDetails/userProfile work on the hot path.
- changed optimistic created-chat insertion to use ChatListOrdering.upserting, preserving the sorted-array invariant that incremental binary insertion depends on.

Also covered the requested behavior with tests:
- nil-updatedAt optimistic rows sort after dated chats.
- read-marker rows preserve resolved direct-chat title/avatar/isDirect while updating preview/read state.
- repeated read-marker rows with failing group metadata keep groupDetails calls bounded.

Checks:
- local: git diff --check ✅
- local: changed Swift line-length scan ✅
- local: bash -n scripts/ci/macos-sanity-checks.sh ✅
- CI: CodeRabbit ✅
- CI: PR Checks ✅ https://github.com/marmot-protocol/whitenoise-mac/actions/runs/28148429287/job/83360471867
- CI: Static Analysis ✅ https://github.com/marmot-protocol/whitenoise-mac/actions/runs/28148429287/job/83360471939

Head: af7412f92561e92ab3beac394d98d37d10f6b49b
Sensitive paths touched: none

PR #189, opened 2026-06-26
fix: bound raw outgoing attachment reads

### LINKED ISSUE #169: Outgoing image attachment bypasses the raw-size guard, reading an unbounded file into memory when fileSize is unknown
## Summary

When preparing an outgoing image attachment, the raw input bytes are never bounded. The only pre-read size check is on `URLResourceValues.fileSize`, which is **optional** — when it resolves to `nil`, the guard is skipped and the entire file is read into memory and handed to `CGImageSourceCreateWithData`, with no cap on the raw input. For the image path there is no second `data.count` guard to catch it (unlike the generic/document path), so an arbitrarily large image is fully buffered + decoded.

## Where

`whitenoise-mac/Models/MessengerModels.swift`

The single pre-read guard only fires when `fileSize` is present:

```swift
// preparedAttachment(fromFileURL:)  (615-618)
if let fileSize = resourceValues.fileSize, fileSize > maxAttachmentBytes {
    throw Failure.attachmentTooLarge(fileSize)
}
let data = try Data(contentsOf: url)   // 618 — whole file into memory, no fallback cap
```

`preparedAttachmentValue` then branches to the image path **before** any `data.count` check:

```swift
// preparedAttachmentValue(...)  (658-674)
let kind = kind(for: typeIdentifier, fileName: fileName)
if kind == .image {
    return try imageAttachment(from: data, fileName: fileName)   // 660 — early return
}
...
guard data.count <= maxAttachmentBytes else {                    // 672 — only the NON-image path
    throw Failure.attachmentTooLarge(data.count)
}
```

`imageAttachment` only validates the *encoded output*, never the raw input:

```swift
// imageAttachment(from:fileName:)  (727-742)
guard let source = CGImageSourceCreateWithData(data as CFData, nil) else { ... }   // 728 — full raw bytes
...
let encoded = try encodeJPEG(image)
guard encoded.count <= maxImageAttachmentBytes else {                              // 740 — output only
    throw Failure.attachmentTooLarge(encoded.count)
}
```

(`maxAttachmentBytes = 50 MiB`, `maxImageAttachmentBytes = 10 MiB`, `maxLongEdge = 2048` — `MessengerModels.swift:579-581`.)

## The problem

`URLResourceValues.fileSize` is documented as optional and is `nil` for a number of provider/URL types (some drag-and-drop promised items, certain file-coordination URLs, non-regular files). Whenever it is absent:

1. The `fileSize > maxAttachmentBytes` guard is skipped (the `if let` simply doesn't fire).
2. `Data(contentsOf: url)` reads the whole file into memory regardless of size.
3. For `kind == .image`, control returns at line 660 before the `data.count <= maxAttachmentBytes` guard at 672, and `imageAttachment` only ever bounds the *re-encoded* JPEG — never the raw `data` it decodes.

So a multi-hundred-MB image whose URL doesn't report a file size is fully buffered and decoded with no limit. Even when `fileSize` *is* reported, the image path permits up to the full 50 MiB raw (the generic 50 MiB cap), then decodes it; the missing-fileSize case removes the cap entirely. This is a self-inflicted memory-spike / potential OOM on a common user action (attaching a picture), and a defense-in-depth gap in the one place outgoing attachment size is supposed to be enforced.

## Suggested fix

Apply a raw-size guard to **every** path, immediately after the read, before kind branching:

```swift
let data = try Data(contentsOf: url)
guard data.count <= maxAttachmentBytes else {
    throw Failure.attachmentTooLarge(data.count)
}
```

(Optionally bound images by a tighter raw limit, since they are re-encoded down to `maxImageAttachmentBytes` anyway.) This makes the cap independent of whether `fileSizeKey` happens to be populated, and removes the image-path bypass.

## Severity

LOW — bounded by the user attaching their own (large) file rather than attacker-controlled input, but it is a real unbounded read + decode on a hot user action and defeats the intended size guard when `fileSize` is `nil`.

agent-p1p:
Closes #169

## Summary
- Replaced the unbounded outgoing file read with a bounded FileHandle read that stops at `maxAttachmentBytes + 1`, so URLs with missing `fileSize` cannot force arbitrary buffering before validation.
- Added a shared raw-size guard and invoke it in the common attachment-value path before image/non-image branching, preventing images from reaching decode/re-encode without the same raw cap as other attachments.
- Removed the now-redundant non-image-only `data.count` guard.

## Verification
- `git diff --check` — passed.
- macOS CI commands were inspected in `.github/workflows/mac-ci.yml` but not run locally: this host is Linux and lacks `xcrun`, `xcodebuild`, and `swift`.

## Sensitive paths
None. This touches only outgoing attachment preparation in `whitenoise-mac/Models/MessengerModels.swift`.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/189">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Models/MessengerModels.swift
+++ b/whitenoise-mac/Models/MessengerModels.swift
@@ -615,7 +615,7 @@ nonisolated enum OutgoingMediaDraftProcessor {
             if let fileSize = resourceValues.fileSize, fileSize > maxAttachmentBytes {
                 throw Failure.attachmentTooLarge(fileSize)
             }
-            let data = try Data(contentsOf: url)
+            let data = try readAttachmentData(from: url)
             return try await SendableAttachment(
                 attachment: preparedAttachmentValue(
                     from: data,
@@ -624,6 +624,16 @@ nonisolated enum OutgoingMediaDraftProcessor {
                 ))
         }.value
         return prepared.attachment
+    }
+
+    private static func readAttachmentData(from url: URL) throws -> Data {
+        let handle = try FileHandle(forReadingFrom: url)
+        defer { try? handle.close() }
+        // Read at most one byte past the cap so URLs without fileSize cannot force
+        // unbounded buffering before the raw-size guard runs.
+        let data = try handle.read(upToCount: maxAttachmentBytes + 1) ?? Data()
+        try enforceRawAttachmentSize(data)
+        return data
     }
 
     static func preparedVoiceAttachment(from recording: VoiceRecordingResult) async throws -> PendingMediaAttachment {
@@ -655,6 +665,9 @@ nonisolated enum OutgoingMediaDraftProcessor {
         fileName: String?,
         typeIdentifier: String?
     ) async throws -> PendingMediaAttachment {
+        // Keep the raw-size invariant in the common funnel too, so no caller can
+        // branch into image decoding with unbounded data.
+        try enforceRawAttachmentSize(data)
         let kind = kind(for: typeIdentifier, fileName: fileName)
         if kind == .image {
             return try imageAttachment(from: data, fileName: fileName)
@@ -669,9 +682,6 @@ nonisolated enum OutgoingMediaDraftProcessor {
         else {
             throw Failure.unsupportedAttachment
         }
-        guard data.count <= maxAttachmentBytes else {
-            throw Failure.attachmentTooLarge(data.count)
-        }
         let videoDim = kind == .video ? await MediaVideoMetadata.dim(from: data, mediaType: mediaType) : nil
         return try genericAttachment(
             from: data,
@@ -680,6 +690,12 @@ nonisolated enum OutgoingMediaDraftProcessor {
             kind: kind,
             videoDim: videoDim
         )
+    }
+
+    private static func enforceRawAttachmentSize(_ data: Data) throws {
+        guard data.count <= maxAttachmentBytes else {
+            throw Failure.attachmentTooLarge(data.count)
+        }
     }
 
     private static func kind(for typeIdentifier: String?, fileName: String?) -> MessageMediaKind? {


ghost:

| | Title |
|---|-------|
| 1 | [Introduce shared raw attachment size validation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/189/chapters/1) |
| 2 | [Implement bounded file reading for attachments](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/189/chapters/2) |
| 3 | [Enforce size guards in the attachment funnel](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/189/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/189">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit b436852092ce1120870f818f96d80466b812cf2a on Jun 26, 2026 2:01pm UTC.</sub>

agent-p1p:
## Pip adversarial review

Reviewed issue #169 against PR #189. The diff replaces the unbounded `Data(contentsOf:)` attachment read with a bounded `FileHandle` read of `maxAttachmentBytes + 1`, then enforces the raw-size invariant before the image/non-image branch. That closes the `fileSize == nil` bypass and preserves the exact-cap behavior: `maxAttachmentBytes` is accepted, `maxAttachmentBytes + 1` is rejected before image decoding.

Checks:
- Local diff scans found no conflict markers, debug artifacts, or credential-looking additions.
- GitHub Mac CI completed successfully: `PR Checks`, `Static Analysis`, and the `CodeRabbit` status context are green.
- CodeRabbit's visible PR comment is a rate-limit notice only; it left no actionable findings.

Findings:
- blocking: 0
- suggestions: 0
- nitpicks: 0

Sensitive paths touched: none observed.

PR #219, opened 2026-06-27
fix: refresh audit log files after delete failure

### LINKED ISSUE #214: Bug: deleteAllAuditLogFiles() leaves the file list stale after a mid-loop delete failure
## Summary

`deleteAllAuditLogFiles()` deletes files in a loop and only calls `loadAuditLogFiles()` on the success path. If one `deleteAuditLogFile` throws partway through, the loop aborts: some files are already deleted on disk, but the UI list is never refreshed, so already-deleted files keep showing as present.

## Location

`whitenoise-mac/Core/WorkspaceState+Settings.swift:525-541`

```swift
do {
    for file in auditLogFiles {
        _ = try await client.deleteAuditLogFile(path: file.path)
    }
    await loadAuditLogFiles()   // only reached on full success
} catch {
    lastError = error.localizedDescription   // no reload
}
```

## Details

On a mid-loop failure the on-disk state and the displayed `auditLogFiles` diverge: files before the failing one are gone but still listed. Because these are audit logs that can contain sensitive metadata (cf. #15), an inaccurate "still present" indication is misleading and may prompt redundant delete attempts.

## Suggested fix

Reload regardless of outcome:

```swift
defer { Task { await loadAuditLogFiles() } }
```

or call `await loadAuditLogFiles()` in the `catch` as well as the success path.

## Review notes

Verified directly.

agent-p1p:
## Summary
- refresh the audit log file list after `deleteAllAuditLogFiles()` finishes, even when a delete throws partway through
- add a regression test covering a mid-loop delete failure so already-deleted files disappear from state

## Test Plan
- `git diff --check`
- `git grep -n -E '^(<<<<<<<|=======|>>>>>>>)' -- . ':!Vendored'` (no conflict markers)
- GitHub Actions: `PR Checks` passed
- GitHub Actions: `Static Analysis` passed
- Not run locally: `xcodebuild` / Swift tests unavailable on this Linux worker (`xcodebuild`, `xcrun`, and `swift` not installed)

Closes #214

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Settings.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Settings.swift
@@ -534,10 +534,11 @@ extension WorkspaceState {
             for file in auditLogFiles {
                 _ = try await client.deleteAuditLogFile(path: file.path)
             }
-            await loadAuditLogFiles()
         } catch {
             lastError = error.localizedDescription
         }
+
+        await loadAuditLogFiles()
     }
 
     func uploadAuditLogFiles() async {
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -7469,6 +7469,44 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func deleteAllAuditLogFilesRefreshesListAfterMidLoopFailure() async throws {
+        let account = AccountSummaryFfi(
+            label: "Desktop Account",
+            accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
+            localSigning: true,
+            signedOut: false,
+            running: true
+        )
+        let firstFile = AuditLogFileFfi(
+            accountRef: account.label,
+            path: "/tmp/audit-1.jsonl",
+            fileName: "audit-1.jsonl",
+            sizeBytes: 123,
+            modifiedAtMs: nil
+        )
+        let secondFile = AuditLogFileFfi(
+            accountRef: account.label,
+            path: "/tmp/audit-2.jsonl",
+            fileName: "audit-2.jsonl",
+            sizeBytes: 456,
+            modifiedAtMs: nil
+        )
+        let runtime = FakeMarmotRuntime(accounts: [account])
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Refresh audit logs after deletion failure](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/219/chapters/1) |
| 2 | [Enhance test runtime to simulate deletion errors](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/219/chapters/2) |
| 3 | [Add regression test for mid-loop deletion failure](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/219/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/219">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit b461f2d5f5a7da6d0840eeb51ce3d4bbdfe89dc5 on Jun 27, 2026 3:46am UTC.</sub>

agent-p1p:
Pip adversarial review for PR #219.

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- Misleading reused error case: `whitenoise-macTests/whitenoise_macTests.swift:8580` simulates the mid-loop delete failure with `FakeMarmotRuntimeError.unused`. That case is now used, so a dedicated name like `auditLogDeleteFailed` would read more clearly. Cosmetic only.
- Test could assert the specific delete error string: `whitenoise-macTests/whitenoise_macTests.swift:7506` currently checks `state.lastError != nil`. An exact assertion would better lock in that the unconditional reload does not clobber the delete error. Optional hardening.

SUMMARY: The fix is correct, minimal, and targeted. Moving `await loadAuditLogFiles()` outside the `do` block reloads the displayed audit-log list after both full success and mid-loop delete failure, which directly fixes issue #214. The regression test is meaningful: the fake removes the first file, throws on the second, and then verifies the refreshed UI state only lists the remaining file. CI is green; local Swift/Xcode tests are unavailable on this Linux worker.

VERDICT: APPROVE

SENSITIVE_PATHS: `whitenoise-mac/Core/WorkspaceState+Settings.swift` handles audit-log file state; audit logs may contain sensitive metadata. No data-exposure regression found.

PR #222, opened 2026-06-27
fix: clamp disappearing-message custom labels

### LINKED ISSUE #212: Hardening: DisappearingMessageOption.custom label uses trapping Int(UInt64) — crashes on values > Int.max
## Summary

`DisappearingMessageOption.custom(value).label` formats a `UInt64` with `Int(value)`, which **traps** for any value greater than `Int.max`. The value originates from the Marmot core (`GroupDetailsSnapshot.disappearingMessageSecs: UInt64`), so a garbage/hostile retention epoch would crash the group-details UI when the picker is shown.

## Location

`whitenoise-mac/Views/GroupViews.swift:759`

```swift
case .custom(let value): return String(format: L10n.string("%d seconds"), Int(value))
```

`value` is `UInt64` (`seconds: UInt64`, line 741-750; `option(for: UInt64)`, line 764-765).

## Details

`Int(UInt64)` is a trapping conversion that aborts the process for values above `Int.max`. This is the same class of issue the codebase already hardened elsewhere — `unsignedInteger()` rejecting bad JSON numbers (#179) and the generation counters switching from trapping `+= 1` to `&+=` (#182) — so this trapping conversion is inconsistent with the established hardening posture for core/peer-influenced numeric values.

## Suggested fix

Use a non-trapping conversion or format the `UInt64` directly:

```swift
case .custom(let value): return String(format: L10n.string("%d seconds"), Int(clamping: value))
// or interpolate the UInt64 without %d
```

## Review notes

Verified directly. Same hardening family as #179 / #182; not otherwise filed.

agent-p1p:
## Summary
- Clamp `DisappearingMessageOption.custom` values before formatting the label so hostile/garbage `UInt64` retention values cannot trap the UI.
- Add a regression test for an oversized core-provided disappearing-message value.

## Test Plan
- `git diff --check`
- `xcodebuild` unavailable on this Linux host; Swift/macOS unit tests not run locally.

Closes #212

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/222">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed how custom disappearing-message durations are displayed so large second values show correctly, including values above the usual integer range.
  * Added support for a new “seconds” localization label for these duration values.

* **Tests**
  * Added coverage to verify custom duration labels format full 64-bit values accurately.


ghost:

| | Title |
|---|-------|
| 1 | [Add localizable string for 64-bit seconds](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/222/chapters/1) |
| 2 | [Fix custom disappearing message label formatting](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/222/chapters/2) |
| 3 | [Verify formatting with regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/222/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/222">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 7e0ab65128f8216b3969414b2f1dc7a497d755ee on Jun 27, 2026 4:21am UTC.</sub>

agent-p1p:
Pip adversarial review: REQUEST_CHANGES

BLOCKING:
- `whitenoise-mac/Views/GroupViews.swift:760` — The trapping conversion is gone, but the label can still render garbage for the hostile/oversized values this issue is about. The code now clamps to a 64-bit `Int`, then feeds that value to a `%d` format. `%d` is a 32-bit signed integer conversion under printf/Foundation formatting semantics, so values above `Int32.max` are truncated when rendered. For this PR's own regression input, `Int(clamping: UInt64(Int.max) + 1)` becomes `Int.max`; formatted with `%d`, that is the low 32 bits and renders as `-1 seconds` on printf-style varargs. That converts the crash into a misleading negative label. Use a width-correct format instead, e.g. `%llu seconds` with the `UInt64`/`CUnsignedLongLong` value, or direct interpolation if localization is not needed.

SUGGESTION:
- `whitenoise-macTests/PureValueTests.swift:27-30` — The regression test only proves the property does not trap. It computes the expected label with the same expression used by the implementation, so it cannot catch the `%d` truncation/wrong-label defect. Pin the actual rendered text instead, e.g. with the oversized `UInt64` literal expected from a `%llu`/interpolated implementation, and consider adding a large mid-range value above `Int32.max` as well.

NITPICK:
- `whitenoise-mac/Views/GroupViews.swift:760` — `%d seconds` is not present in `Localizable.xcstrings` (`%d members`, `%lld members`, and `%llu bytes` are). This line was already unlocalized before the PR, but if it changes to a width-correct format string, adding the matching localization key would avoid leaving another fallback-only UI label.

SUMMARY: This PR removes the direct `Int(UInt64)` trap, so it addresses the crash path in part. It needs one more formatting fix before merge: do not pass a 64-bit/clamped value through `%d`, and make the test assert the real output rather than reusing the implementation expression.

VERDICT: REQUEST_CHANGES

SENSITIVE_PATHS: none

agent-p1p:
Addressed adversarial review findings in 7e0ab65:

- Replaced `%d`/`Int(clamping:)` with `%llu` and `CUnsignedLongLong(value)` for the custom disappearing-message label.
- Pinned regression expectations for `Int32.max + 1` and `Int.max + 1` values.
- Added the `%llu seconds` key to `Localizable.xcstrings`.

CI is green. CodeRabbit re-review completed with no actionable comments.

PR #226, opened 2026-06-27
fix: ignore unsupported imeta blurhash field

### LINKED ISSUE #208: Bug: imeta `blurhash` field makes the parser discard the entire media attachment
## Summary

When parsing a NIP-92 `imeta` tag, encountering a `blurhash <value>` field causes an immediate `return nil`, which discards the **entire** media attachment rather than just ignoring the unsupported field. `blurhash` is a standard, very common optional field in `imeta` tags, so attachments from spec-compliant / other White Noise clients silently vanish on macOS.

## Location

`whitenoise-mac/Core/MarmotMapping.swift:571-574` (`reference(fromIMetaTag:sourceEpoch:)`)

```swift
for field in tag.dropFirst() {
    if field.hasPrefix("blurhash ") {
        return nil          // <-- aborts the WHOLE attachment
    }
    ...
}
```

## Details

- The supported placeholder field is `thumbhash` (see `MediaAttachmentReferenceFfi`); there is no `blurhash` field, so this special-case exists purely to reject-if-present.
- Every *other* unrecognized field falls through harmlessly into the generic collector at lines 584-588 (`fields[key] = value`) and is simply ignored. `blurhash` is uniquely fatal — an inconsistency with no apparent justification (the generic parser would handle `blurhash <hash>` fine; the hash value contains no spaces that would break the split).
- This is the local-parse fallback path (records with empty `record.media` + populated `imeta` tags). For those records, any image carrying a blurhash placeholder renders with **no media at all**.

Strictly worse than the already-fixed #137 (which dropped only a single field, `sourceEpoch`); here the whole attachment is lost.

## Impact

User-visible media loss / interop break: messages with image attachments that include a `blurhash` (common from mobile clients and other Nostr clients) appear to have no attachment on macOS.

## Suggested fix

Treat `blurhash` like any other unrecognized field — skip and continue instead of aborting:

```swift
for field in tag.dropFirst() {
    if field.hasPrefix("blurhash ") {
        continue   // unsupported placeholder; ignore, do not drop the attachment
    }
    ...
}
```

(Optionally capture it into `fields["blurhash"]` for future use.)

## Review notes

Verified directly. Checked against existing media-parsing issues (#137, #179, #185, #120); none cover the blurhash full-attachment-drop behavior.

agent-p1p:
## Summary
- Treat NIP-92 `blurhash` imeta fields as unsupported optional metadata instead of dropping the whole attachment.
- Add a regression test covering local media JSON parsing with a blurhash field.

## Verification
- `git diff --check`
- `git grep -n -E '^(<<<<<<<|=======|>>>>>>>)' -- . ':!*.xcstrings'` (no matches)
- Added-line length check against 120-char Swift format limit
- `xcodebuild`, `xcrun`, `swift`, and `swift-format` are not installed on this Linux worker, so macOS build/test/format checks need GitHub CI/macOS.

## Sensitive paths
None by repo policy. This changes local NIP-92 media metadata parsing only; it does not touch auth, key handling, crypto/MLS/CGKA, trust anchors, group-state core, entitlements, or app secrets.

Closes #208

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/226">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -570,7 +570,7 @@ private nonisolated enum MessageMediaParser {
 
         for field in tag.dropFirst() {
             if field.hasPrefix("blurhash ") {
-                return nil
+                continue
             }
             if let locator = field.dropPrefix("locator "),
                 let split = locator.firstIndex(of: " ")
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -570,6 +570,7 @@ private nonisolated enum MessageMediaParser {
 
         for field in tag.dropFirst() {
             if field.hasPrefix("blurhash ") {
+                // Recognized NIP-92 placeholder; unsupported locally, but keep the attachment.
                 continue
             }
             if let locator = field.dropPrefix("locator "),
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -1362,6 +1362,38 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func imetaBlurhashFieldDoesNotDropAttachment() async throws {
+        // Regression for whitenoise-mac#208: `blurhash` is a standard optional NIP-92
+        // imeta field. The macOS client does not consume it, but its presence must not
+        // make the local-parse fallback discard the whole media attachment.
+        let reference = mediaAttachmentReference(mediaType: "image/png", fileName: "photo.png")
+        let page = TimelinePageFfi(
+            messages: [
+                timelineMessage(
+                    id: "blurhash-imeta",
+                    groupIdHex: "group",
+                    sender: "alice",
+                    plaintext: "",
+                    recordedAt: 1_700_000_000,
+                    mediaJson: mediaJson(
+                        for: reference,
+                        appendingIMetaField: "blurhash LEHV6nWB2yk8pyo0adR*.7kCMdnj"
+                    )
+                )
+            ],
+            hasMoreBefore: false,
+            hasMoreAfter: false
+        )
+
+        let messages = MessageItem.timeline(from: page, activeAccountIdHex: "self")
+        let message = try #require(messages.first)
+
+        #expect(message.mediaAttachments.count == 1)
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Ignore blurhash field in NIP-92 imeta tags](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/226/chapters/1) |
| 2 | [Add regression test for blurhash field handling](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/226/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/226">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 961c90a63f7cd572d3d2de3a0e95e085fc5799a3 on Jun 27, 2026 5:18am UTC.</sub>

agent-p1p (commented):
## Pip adversarial review

Verdict: APPROVE-equivalent (comment review; no blocking findings)

### BLOCKING
none

### SUGGESTION
- `whitenoise-mac/Core/MarmotMapping.swift:572-574` — The retained `if field.hasPrefix("blurhash ") { continue }` block is now functionally redundant with the generic imeta field collector below it. That is not a correctness problem, and keeping it can act as an explicit “recognized but unsupported placeholder” marker. Consider either deleting the special-case entirely or adding a one-line comment so a future reader does not treat it as mysterious dead code.

### NITPICK
none

### Summary
The fix is correct and minimal. Replacing the field-level `return nil` with `continue` prevents a standard optional NIP-92 `blurhash` field from discarding the whole attachment in the local media JSON fallback path. The added regression test exercises the intended path (`media: []` plus populated `mediaJson`) and would have failed before the change with zero parsed attachments.

Verification reviewed:
- Conflict-marker grep passed locally.
- `bash -n scripts/ci/macos-sanity-checks.sh` passed locally.
- GitHub Mac CI `PR Checks` and `Static Analysis` are green.
- Local Swift/Xcode tests were not runnable on this Linux worker (`xcodebuild`, `xcrun`, `swift`, and `swift-format` missing).

Sensitive paths touched: none.

agent-p1p:
Addressed adversarial review suggestion in 961c90a: added a clarifying comment for the explicitly ignored `blurhash` imeta placeholder.

Checks are green again:
- PR Checks
- Static Analysis
- CodeRabbit

PR #262, opened 2026-07-01
fix: guard unavailable notification targets

### LINKED ISSUE #256: Bug: tapping a notification whose account was removed (or otherwise unresolvable) selects the group under the *currently active* account with no ownership check
## Severity

**LOW** (bug)

## Summary

`handleNotificationResponse` only performs an account switch when `notificationAccount(from:)` returns a non-nil account that differs from the active one. If that lookup returns `nil` — e.g. the notification's account was **removed** between delivery and tap, or `accountRef`/`accountIdHex` no longer matches any entry in `accounts` — `switchedAccounts` stays `false` and the code unconditionally selects and loads the notification's `groupIdHex` against whatever account happens to be active, with no check that the group belongs to it.

## Evidence

`whitenoise-mac/Core/WorkspaceState+Notifications.swift:56-80`:

```swift
func handleNotificationResponse(_ userInfo: [String: String]) {
    guard let groupIdHex = userInfo["groupIdHex"] else { return }

    let switchedAccounts: Bool
    if let account = notificationAccount(from: userInfo), activeAccountId != account.id {
        prepareForActiveAccountSwitch(to: account, preservingMessageCacheFor: groupIdHex)
        switchedAccounts = true
    } else {
        switchedAccounts = false            // <-- also taken when notificationAccount == nil
    }

    selection = .chat(groupIdHex)           // <-- selects a foreign group under the active account
    isChatListVisible = true
    if !switchedAccounts {
        closeNewChatComposer()
        pruneMessageCache(keeping: groupIdHex)   // retains a cache key for a group this account can't access
    }
    NSApplication.shared.activate(ignoringOtherApps: true)
    beginTimelineInitialLoadIfNeeded(groupIdHex: groupIdHex)
    Task {
        await reloadChats()
        await loadMessages(groupIdHex: groupIdHex)
    }
}
```

`notificationAccount(from:)` (`:257-268`) returns `nil` when neither `accountIdHex` nor `accountRef` matches any current account.

## Failure scenario

1. A notification for account B is delivered and persists in macOS Notification Center.
2. The user removes account B.
3. The user later clicks the still-present B notification. `notificationAccount(from:)` returns `nil`, so `switchedAccounts == false`.
4. The code runs `selection = .chat(groupIdHex)` and `loadMessages(groupIdHex:)` against account A. That group does not belong to A, so the timeline load fails / shows an empty or foreign conversation, and `pruneMessageCache(keeping: groupIdHex)` retains a cache key for a group A cannot access.

## Fix

When `notificationAccount(from:)` returns `nil` (or the target account is signed-out/removed), do not blindly select `.chat(groupIdHex)`. Verify the group exists for the active account (`chatItem(accountId:chatId:)`) before selecting; otherwise surface a background status and bail.

## Related

Distinct from closed #99 (which covers the *successful* cross-account switch teardown) — here the account lookup fails entirely and there is no active-account ownership guard.

agent-p1p:
## Summary
- Guard notification taps whose account cannot be resolved or is signed out.
- Only select/load the notification chat without an account switch if the active account already owns that chat.
- Surface a background status and bail instead of selecting a foreign group.

## Test Plan
- `git diff --check`
- conflict marker sweep: `git grep -n -E '^(<<<<<<<|=======|>>>>>>>)' HEAD -- .` (no matches)
- duplicate Swift function-name heuristic over changed files (no matches)
- not run: `xcodebuild` / Swift tests unavailable on this Linux host

Closes #256

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/262">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Notifications.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Notifications.swift
@@ -56,11 +56,19 @@ extension WorkspaceState {
     func handleNotificationResponse(_ userInfo: [String: String]) {
         guard let groupIdHex = userInfo["groupIdHex"] else { return }
 
+        let account = notificationAccount(from: userInfo)
         let switchedAccounts: Bool
-        if let account = notificationAccount(from: userInfo), activeAccountId != account.id {
+        if let account, !account.signedOut, activeAccountId != account.id {
             prepareForActiveAccountSwitch(to: account, preservingMessageCacheFor: groupIdHex)
             switchedAccounts = true
         } else {
+            guard activeAccountContainsNotificationGroup(groupIdHex) else {
+                setBackgroundStatus(
+                    L10n.string("This notification is for an account or chat that is no longer available.")
+                )
+                NSApplication.shared.activate(ignoringOtherApps: true)
+                return
+            }
             switchedAccounts = false
         }
 
@@ -77,6 +85,11 @@ extension WorkspaceState {
             await reloadChats()
             await loadMessages(groupIdHex: groupIdHex)
         }
+    }
+
+    func activeAccountContainsNotificationGroup(_ groupIdHex: String) -> Bool {
+        guard let activeAccountId else { return false }
+        return chatItem(accountId: activeAccountId, chatId: groupIdHex) != nil
     }
 
     func startNotificationListener() {
--- a/whitenoise-mac/Core/WorkspaceState+Notifications.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Notifications.swift
@@ -61,14 +61,16 @@ extension WorkspaceState {
         if let account, !account.signedOut, activeAccountId != account.id {
             prepareForActiveAccountSwitch(to: account, preservingMessageCacheFor: groupIdHex)
             switchedAccounts = true
-        } else {
-            guard activeAccountContainsNotificationGroup(groupIdHex) else {
+        } else if account == nil || account?.signedOut == true {
+            guard activeAccountHasChat(groupIdHex: groupIdHex) else {
                 setBackgroundStatus(
                     L10n.string("This notification is for an account or chat that is no longer available.")
                 )
                 NSApplication.shared.activate(ignoringOtherApps: true)
                 return
             }
+            switchedAccounts = false
+        } else {
             switchedAccounts = false
         }
 
@@ -87,7 +89,7 @@ extension WorkspaceState {
         }
     }
 
-    func activeAccountContainsNotificationGroup(_ groupIdHex: String) -> Bool {
+    func activeAccountHasChat(groupIdHex: String) -> Bool {
         guard let activeAccountId else { return false }
         return chatItem(accountId: activeAccountId, chatId: groupIdHex) != nil
     }
--- a/whitenoise-mac/Localizable.xcstrings
+++ b/whitenoise-mac/Localizable.xcstrings
@@ -23903,6 +23903,65 @@
         }
       }
     },
+    "This notification is for an account or chat that is no longer available." : {
+      "extractionState" : "manual",
+      "localizations" : {
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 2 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Add localized error message for unavailable notifications](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/262/chapters/1) |
| 2 | [Guard notification taps for unavailable accounts or chats](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/262/chapters/2) |
| 3 | [Test notification response guarding and account switching](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/262/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/262">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2a473db272f9d9b1dfec5fb4bc1bef917f5fd526 on Jul 1, 2026 10:12pm UTC.</sub>

agent-p1p:
## Pip adversarial review

VERDICT: REQUEST_CHANGES

BLOCKING:
- Same-account notification taps for not-yet-synced groups are now rejected before the reload can discover them: `whitenoise-mac/Core/WorkspaceState+Notifications.swift:64-71`. The new `activeAccountContainsNotificationGroup(groupIdHex)` guard runs in the shared `else` path before `reloadChats()` is started. That `chatItem(...)` lookup only reflects the current `chatLookupByAccount` snapshot. The old flow selected first, then reloaded, which allowed legitimate active-account notifications for fresh group invites or newly-arrived chats to open even if the chat list snapshot had not caught up yet. This PR now treats those valid same-account taps as “account or chat no longer available” and returns without ever reloading. The ownership guard should only apply to the unresolved/signed-out target cases, or it should reload/verify before bailing.

SUGGESTION:
- Split the notification-response control flow by case instead of using one broad `else`: `WorkspaceState+Notifications.swift:61-72`. Today that branch covers `account == nil`, `account.signedOut`, and `account.id == activeAccountId`. Only the first two are the issue’s removed/signed-out target cases; the active-account happy path should keep the existing select-then-reload behavior.
- The new background-status bail duplicates `NSApplication.shared.activate(...)` with the success path (`WorkspaceState+Notifications.swift:69` and `:81`). Not blocking, but a small refactor could avoid the duplicate.

NITPICK:
- `activeAccountContainsNotificationGroup(_:)` is slightly awkward naming. `activeAccountHasChat(groupIdHex:)` would match the `chatItem(accountId:chatId:)` vocabulary more closely.
- Add a regression test for the blocking case: a notification that resolves to the active account, with the group absent from the current chat lookup before reload, should not be rejected as unavailable.
- The new localized string relies on `L10n.string` identity fallback. Functionally fine, but confirm the string catalog/localizations are updated if this app expects translated UI strings.

SUMMARY: The PR correctly blocks removed/unresolvable and signed-out account notifications from selecting a foreign group under the active account, and the added tests cover those negative paths. The guard is placed too broadly, though: it also gates ordinary active-account notifications using the stale pre-reload chat snapshot, which regresses valid notifications for chats that are not in `chatLookupByAccount` yet. No sensitive paths touched.

SENSITIVE_PATHS: none

agent-p1p:
Address-review update pushed in `2a473db`:

- Split `handleNotificationResponse` so active-account notifications no longer hit the stale pre-reload ownership guard.
- Kept the active-account path on the old select-then-reload behavior.
- Kept the unavailable-target guard only for unresolved/signed-out notification accounts.
- Added regression coverage for the active-account/stale-chat-lookup case.
- Added `Localizable.xcstrings` entries for the unavailable-notification status.

Verification:
- GitHub Mac CI: PR Checks ✅, Static Analysis ✅
- CodeRabbit status ✅
- Local Linux checks: `git diff --check`, committed conflict-marker sweep, `.xcstrings` JSON/locale validation, changed-Swift duplicate-symbol heuristic.

PR #273, opened 2026-07-01
fix: purge unreadable media cache metadata

### LINKED ISSUE #248: Privacy: per-account media-cache purge silently skips (does not delete) entries whose metadata cannot be decrypted/decoded
## Summary

`MessageMediaDiskCache.removeEntries(matchingAccountDigest:root:symmetricKey:)` — the per-account purge path — decides whether to delete each cached entry by decrypting and decoding its `metadata.bin` and comparing `metadata.accountDigest`. Any entry whose metadata cannot be authenticated/decoded hits the `else { continue }` and is **left on disk**, rather than being removed. This is a completeness gap in the same privacy-purge subsystem hardened by #230 (in-flight store re-creating wiped media), #236 (reject direct stores during wipe), and #177 (avatar cache not cleared).

## Evidence

`whitenoise-mac/Core/MessageMediaDiskCache.swift:550-563`:

```swift
for entryDirectory in entryDirectories {
    let cacheID = entryDirectory.lastPathComponent
    let metadataURL = entryDirectory.appendingPathComponent(metadataFileName)
    guard let metadataData = try? Data(contentsOf: metadataURL),
        let metadataPlaintext = try? open(
            metadataData,
            using: symmetricKey,
            authenticatedBy: metadataAAD(for: cacheID)
        ),
        let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext),
        metadata.accountDigest == accountDigest
    else { continue }                              // <-- undecryptable/undecodable entry is SKIPPED, not deleted
    try? FileManager.default.removeItem(at: entryDirectory)
}
```

A truncated/corrupted `metadata.bin`, a metadata file whose directory name no longer matches the AAD `cacheID` (so authentication fails), or a metadata blob that fails `JSONDecoder` all fall through the `guard` and survive the per-account purge.

## Impact

LOW. The orphaned payload remains AES-GCM-encrypted and unreadable without the symmetric key, so this is a cleanup-completeness gap rather than a plaintext-residue leak. The full-wipe path (`purgeAll`) still removes everything. Filing because this subsystem's purge semantics are explicitly load-bearing for the app's privacy guarantees (a chain of issues — #230/#236/#177/#174 — exists specifically to make "remove this account's data" complete), and a silently-retained ciphertext shard undermines the "this identity left no residue on disk" expectation.

## Suggested fix

During a per-account purge, treat an entry whose metadata cannot be authenticated/decoded as garbage and remove it (an undecryptable entry has no retained value), e.g. replace the `continue` with a removal in the un-decodable case:

```swift
guard let metadataData = try? Data(contentsOf: metadataURL),
    let metadataPlaintext = try? open(metadataData, using: symmetricKey, authenticatedBy: metadataAAD(for: cacheID)),
    let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext)
else {
    try? FileManager.default.removeItem(at: entryDirectory)   // unreadable shard -> drop it
    continue
}
if metadata.accountDigest == accountDigest {
    try? FileManager.default.removeItem(at: entryDirectory)
}
```

(If the intent is to be conservative and not touch shards that might belong to another account, at minimum log/track the skipped count rather than silently leaving them.)

## Severity

LOW — undecryptable/corrupt cache shards survive a per-account privacy purge; encrypted-only, so cleanup-completeness rather than a plaintext leak.

agent-p1p:
## Summary
- delete media-cache entries whose metadata cannot be decrypted/decoded during per-account purge
- keep the existing accountDigest check for readable metadata so other accounts' healthy cache entries survive
- add regression coverage for unreadable metadata alongside a preserved other-account entry

Closes #248

## Test plan
- `git diff --check`
- `xcodebuild -scheme whitenoise-mac -configuration Debug build-for-testing CODE_SIGNING_ALLOWED=NO` (fails locally: `xcodebuild` is not installed on this Linux host)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/273">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/MessageMediaDiskCache.swift
+++ b/whitenoise-mac/Core/MessageMediaDiskCache.swift
@@ -556,10 +556,14 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                         using: symmetricKey,
                         authenticatedBy: metadataAAD(for: cacheID)
                     ),
-                    let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext),
-                    metadata.accountDigest == accountDigest
-                else { continue }
-                try? FileManager.default.removeItem(at: entryDirectory)
+                    let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext)
+                else {
+                    try? FileManager.default.removeItem(at: entryDirectory)
+                    continue
+                }
+                if metadata.accountDigest == accountDigest {
+                    try? FileManager.default.removeItem(at: entryDirectory)
+                }
             }
         }
     }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -1163,6 +1163,61 @@ struct whitenoise_macTests {
         #expect(!fileManager.fileExists(atPath: entryDirectory.path))
     }
 
+    @Test func messageMediaDiskCacheAccountPurgeDeletesUnreadableEntries() async throws {
+        let fileManager = FileManager.default
+        let root = fileManager.temporaryDirectory
+            .appendingPathComponent("whitenoise-media-cache-tests-\(UUID().uuidString)", isDirectory: true)
+        defer { try? fileManager.removeItem(at: root) }
+
+        let cache = messageMediaDiskCache(root: root)
+        let corruptPlaintext = Data("account media with unreadable metadata".utf8)
+        let preservedPlaintext = Data("other account media".utf8)
+        let corruptKey = MessageMediaDiskCacheKey(
+            accountId: "account-a",
+            groupIdHex: "group-a",
+            reference: mediaDiskCacheReference(plaintext: corruptPlaintext, ciphertextByte: 0xaa)
+        )
+        let preservedKey = MessageMediaDiskCacheKey(
+            accountId: "account-b",
+            groupIdHex: "group-b",
+            reference: mediaDiskCacheReference(plaintext: preservedPlaintext, ciphertextByte: 0xbb)
+        )
+
+        await cache.store(
+            MessageMediaDownload(
+                data: corruptPlaintext,
+                fileName: "corrupt.jpg",
+                mediaType: "image/jpeg",
+                sizeBytes: UInt64(corruptPlaintext.count),
+                payloadId: "corrupt"
+            ),
+            for: corruptKey
+        )
+        await cache.store(
+            MessageMediaDownload(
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Purge unreadable media cache metadata](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/273/chapters/1) |
| 2 | [Add regression test for unreadable metadata](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/273/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/273">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit b3167e9444b2b0920e3c4cd40f65e07acb7c4587 on Jul 2, 2026 12:06am UTC.</sub>

agent-p1p (commented):
Pip adversarial review for #248.

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- Duplicated `removeItem` call: `whitenoise-mac/Core/MessageMediaDiskCache.swift:561,565` The removal is now present in both the unreadable-metadata branch and the readable matching-account branch. This is clear and correct; a single `shouldRemove` boolean could make the control flow slightly more compact, but this is style-only.
- Directory-level deletion assertion: `whitenoise-macTests/whitenoise_macTests.swift:1182` The test asserts the corrupt entry directory is gone, which also covers the payload inside it. Adequate granularity; no change required.
- Optional public-API assertion: `whitenoise-macTests/whitenoise_macTests.swift:1182` The test could also assert `await cache.cachedDownload(for: corruptKey) == nil`, mirroring the sibling corrupt-entry test. The direct filesystem assertion already verifies the regression, so this is non-blocking.

SUMMARY: The fix is tight and correctly scoped. It deletes entries whose metadata cannot be decrypted/decoded during per-account purge, preserves readable entries for other accounts via the existing `accountDigest` check, and adds meaningful regression coverage with a corrupted entry plus a preserved other-account entry. I do not see a cross-account data-loss regression: the cache uses one device-wide symmetric key, so unreadable metadata is corruption/AAD mismatch rather than another account's healthy entry, and `readDownload` already evicts such entries on read. GitHub PR Checks and Static Analysis are green; local `git diff --check` passed on this host.

VERDICT: APPROVE

SENSITIVE_PATHS: `whitenoise-mac/Core/MessageMediaDiskCache.swift` (privacy-critical per-account media-cache purge path)

PR #274, opened 2026-07-02
fix: reject boolean media JSON strings

### LINKED ISSUE #246: Hardening: MarmotMapping media-JSON string() coerces JSON booleans into attachment field values (inconsistent with unsignedInteger()'s boolean guard)
## Summary

`MessageMediaParser.string(_:keys:)` coerces **any** `NSNumber` to its `.stringValue`, including JSON booleans (`true`/`false` bridge to `NSNumber` backed by `CFBoolean`). So a peer-controlled `media_json` can satisfy required string fields — including the crypto reference fields — with boolean literals, which become `"1"`/`"0"`. The sibling helper `unsignedInteger(_:)` was explicitly hardened against exactly this (it rejects `CFBooleanGetTypeID`), so the two parsers now disagree on what a boolean means.

This is the same parser hardened by #179 (negative/fractional numbers in `source_epoch`) and #137 (camelCase `sourceEpoch`), but the boolean-coercion gap in `string()` is a distinct, unreported site.

## Evidence

`whitenoise-mac/Core/MarmotMapping.swift:644-656`:

```swift
private static func string(_ dictionary: [String: Any], keys: [String]) -> String? {
    for key in keys {
        guard let value = dictionary[key] else { continue }
        if let string = value as? String {
            if let value = string.trimmingCharacters(in: .whitespacesAndNewlines).nilIfBlank {
                return value
            }
        } else if let number = value as? NSNumber {
            return number.stringValue        // <-- JSON booleans (CFBoolean) take this branch -> "1"/"0"
        }
    }
    return nil
}
```

Contrast `unsignedInteger(_:)` at `MarmotMapping.swift:662-672`, which deliberately rejects booleans:

```swift
private static func unsignedInteger(_ value: Any?) -> UInt64? {
    if let number = value as? NSNumber {
        if CFGetTypeID(number) == CFBooleanGetTypeID() {
            return nil
        }
        ...
```

`string()` is used for the required crypto reference fields (`ciphertext_sha256`, `plaintext_sha256`, `nonce`, `file_name`, `media_type`, `version`) and for media-locator `kind`/`value`.

## Failure scenario

A peer sends a message whose `media_json` is:

```json
{ "ciphertext_sha256": true, "plaintext_sha256": true, "nonce": true,
  "file_name": true, "media_type": true, "version": true }
```

Every field is a JSON boolean, so each `string()` call returns `"1"`, fabricating a `MediaAttachmentReferenceFfi` entirely from boolean inputs rather than treating the fields as absent/malformed.

## Impact

LOW. The crypto core re-validates these references downstream, so the worst realistic outcome is a non-crashing malformed attachment, not a memory-safety or crypto break. Filing because it is a clear, easily-closed inconsistency with the established hardening pattern, and because parser robustness on peer-controlled `media_json` is load-bearing for the rest of the attachment pipeline.

## Suggested fix

Apply the same boolean guard `string()` already gets for free in `unsignedInteger()`, and arguably restrict numeric coercion to genuine integer/decimal kinds:

```swift
} else if let number = value as? NSNumber {
    if CFGetTypeID(number) == CFBooleanGetTypeID() { continue }
    return number.stringValue
}
```

## Severity

LOW — peer-controlled JSON booleans are silently coerced into attachment string fields, inconsistent with the boolean rejection already applied to `unsignedInteger()`.

agent-p1p:
## Summary
- Reject CFBoolean-backed NSNumber values in media JSON string fields instead of coercing them to "1"/"0".
- Add regression coverage that boolean required attachment fields do not fabricate a media reference.

## Test plan
- `git diff --check`
- Conflict-marker scan on changed files
- Not run: `xcodebuild`/Swift tests unavailable on this Linux host (`xcodebuild` and `swift` are not installed)

Closes #246

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/274">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of media attachment data so boolean values are no longer mistaken for valid text fields.
  * Parser now ignores boolean-only reference fields and correctly uses later string values when available.
  * Fixed mixed-value cases so malformed entries are skipped and only valid media locators are kept.

### DIFF
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -649,6 +649,11 @@ private nonisolated enum MessageMediaParser {
                     return value
                 }
             } else if let number = value as? NSNumber {
+                if CFGetTypeID(number) == CFBooleanGetTypeID() {
+                    // JSON booleans bridge as NSNumber. Skip them rather than returning
+                    // "1"/"0", and allow later alias keys to supply the real string.
+                    continue
+                }
                 return number.stringValue
             }
         }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -2141,6 +2141,105 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func booleanMediaJSONStringFieldsDoNotFabricateAttachmentReference() async throws {
+        // Regression for whitenoise-mac#246: JSON booleans bridge back from
+        // JSONSerialization as CFBoolean-backed NSNumber values. Required string fields
+        // must reject them instead of accepting NSNumber.stringValue ("1"/"0").
+        let page = TimelinePageFfi(
+            messages: [
+                timelineMessage(
+                    id: "boolean-media-json-fields",
+                    groupIdHex: "group",
+                    sender: "alice",
+                    plaintext: "",
+                    recordedAt: 1_700_000_000,
+                    mediaJson: mediaJSONString(fromJSONObject: [
+                        "ciphertext_sha256": true,
+                        "plaintext_sha256": true,
+                        "nonce": true,
+                        "file_name": true,
+                        "media_type": true,
+                        "version": true,
+                        "locators": [
+                            ["kind": "blossom", "value": "https://blob.example/bool"]
+                        ],
+                    ])
+                )
+            ],
+            hasMoreBefore: false,
+            hasMoreAfter: false
+        )
+
+        let messages = MessageItem.timeline(from: page, activeAccountIdHex: "self")
+        let message = try #require(messages.first)
+
+        #expect(message.mediaAttachments.isEmpty)
+    }
+
+    @MainActor
+    @Test func booleanMediaJSONStringAliasValuesFallThroughAndBadLocatorsAreDropped() async throws {
+        // `string(_:keys:)` searches aliases in order. A boolean at an earlier alias
+        // should be treated as malformed for that key, not as "1" and not as a reason
+        // to reject a later valid alias.
+        let reference = mediaAttachmentReference(mediaType: "image/png", fileName: "photo.png")
+        let page = TimelinePageFfi(
+            messages: [
+                timelineMessage(
+                    id: "boolean-media-json-aliases",
+                    groupIdHex: "group",
+                    sender: "alice",
+                    plaintext: "",
+                    recordedAt: 1_700_000_000,
+                    mediaJson: mediaJSONString(fromJSONObject: [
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Reject boolean values in media JSON strings](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/274/chapters/1) |
| 2 | [Add regression tests for boolean media fields](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/274/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/274">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 9af9a71b1b1c525f4df46424d746b45921fe8bb5 on Jul 2, 2026 12:33am UTC.</sub>

agent-p1p (commented):
Pip adversarial review for #246.

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- Partial alias/locator coverage in the new test: `whitenoise-macTests/whitenoise_macTests.swift:2088` The regression test exercises one alias per required field and does not put booleans in short aliases (`m`, `v`, `nonce_hex`), optional fields (`dim`, `thumbhash`), or locator `kind`/`value`. All route through the same guarded `string()` helper, so this is optional; a single locator-value boolean case would make the peer-controlled locator surface explicit.
- Latent behavior improvement worth a comment: `whitenoise-mac/Core/MarmotMapping.swift:652` The `continue` also fixes a pre-existing alias-priority quirk: a boolean under an earlier alias no longer short-circuits a valid string under a later alias. That is the right behavior, but a one-line comment mirroring `unsignedInteger`'s CFBoolean rationale would aid future readers.

SUMMARY: The fix is correct, minimal, and directly resolves issue #246. It reuses the `CFGetTypeID(number) == CFBooleanGetTypeID()` idiom already established in `unsignedInteger(_:)`, preserves legitimate numeric-to-string coercion, and rejects CFBoolean-backed JSON booleans before they can fabricate required attachment fields. The regression test is meaningful and would fail on the old implementation. Scope is tight: 3 production lines plus one focused test. Local `git diff --check` and conflict-marker scan passed; GitHub PR Checks, Static Analysis, and CodeRabbit are green. Swift/xcodebuild tests remain not runnable on this Linux host.

VERDICT: APPROVE

SENSITIVE_PATHS: `whitenoise-mac/Core/MarmotMapping.swift` (parser for peer-controlled `media_json` attachment crypto-reference fields)

agent-p1p:
@coderabbitai review

PR #303, opened 2026-07-02
fix: cache media reference resolution

### LINKED ISSUE #286: Performance: resolvedMediaReference runs a full listMedia(limit: nil) + linear scan per attachment on every cache miss (no memoization/coalescing)
## Severity
**MEDIUM** (performance)

## Summary
When an attachment's `reference.sourceEpoch == 0`, `resolvedMediaReference` fetches the entire media-record list for the group (`limit: nil`) and linearly scans it to match a SHA. This runs once per attachment per disk-cache-miss, with no caching of the record list and no coalescing across attachments of the same group.

## Details
`whitenoise-mac/Core/WorkspaceState+Media.swift:481-498`, called from `loadMediaAttachment` at line 184 (only on a disk-cache miss). `sourceEpoch == 0` is not rare — `MarmotMapping` defaults imeta-tag-sourced references to `sourceEpoch: 0` and JSON references to `... ?? 0`. So the first render of a media-heavy chat (V visible attachments, M group media records) issues V independent full-list FFI calls + V full decodes + V linear scans — O(V·M) — all firing concurrently on the `.concurrent` `ffiQueue`, and duplicating the same list fetch V times.

## Evidence
```swift
guard reference.sourceEpoch == 0 else { return reference }
let records = try await runOffMain {
    try client.listMedia(accountRef: accountRef, groupIdHex: groupIdHex, limit: nil)
}
return records.first { record in
    record.reference.plaintextSha256 == reference.plaintextSha256
        || record.reference.ciphertextSha256 == reference.ciphertextSha256
}?.reference ?? reference
```

## Suggested fix
Cache the resolved records per `(accountId, groupIdHex)` (invalidated on new media / account switch), or coalesce concurrent `resolvedMediaReference` lookups for the same group into a single in-flight `listMedia` (the pattern already used by `RemoteImageLoadRegistry`). At minimum build a `[sha: reference]` index once per group rather than re-scanning per attachment.

## Scope note
Found during a full-codebase review; not covered by the existing media/perf issues.

agent-p1p:
## Summary
- cache sourceEpoch=0 media-reference resolution per account/group as a SHA index
- coalesce concurrent full-list `listMedia(limit: nil)` lookups for the same group
- clear cached indexes on account/group teardown and after media sends

## Test Plan
- `git diff --check HEAD^ HEAD`
- `xcodebuild -scheme whitenoise-mac -configuration Debug build CODE_SIGNING_ALLOWED=NO` (not run: `xcodebuild` is not installed on this Linux worker)

Closes #286

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303">
  <picture>
    Open in Stage
  </picture>
</a>


ghost:

| | Title |
|---|-------|
| 1 | [Define media reference cache types and utilities](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303/chapters/1) |
| 2 | [Implement cached media reference resolution logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303/chapters/2) |
| 3 | [Invalidate media cache on account and group changes](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303/chapters/3) |
| 4 | [Clear media cache after message operations](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303/chapters/4) |
| 5 | [Test media reference caching and coalescing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/303">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 47f8d7485a7d5fa1710d214a3f3af6b46d752e18 on Jul 4, 2026 1:59pm UTC.</sub>

agent-p1p:
⚠️ Could not submit a formal Request Changes review from the bot account, so recording the blocking review as a PR comment.

Pip adversarial review for marmot-protocol/whitenoise-mac#286 / PR #303.

BLOCKING:
- Misses against the new per-group SHA index still trigger a fresh full `listMedia(limit: nil)` on every sequential retry: `whitenoise-mac/Core/WorkspaceState+Media.swift:514-525`, `whitenoise-mac/Core/WorkspaceState+Media.swift:528-562`. `resolvedMediaReference` only returns from `mediaReferenceIndexes` when the lookup is a hit; a no-match result falls through to `mediaReferenceIndex`, which only coalesces in-flight tasks and then rebuilds the same full index. For any `sourceEpoch == 0` attachment whose SHA is not yet/ever present in `listMedia`, the PR still does full-list FFI + decode + index rebuild on each disk-cache miss. That leaves a remaining form of the performance bug this issue is meant to remove. Treat an existing fetched index as authoritative for nil lookups until invalidated, or otherwise add an explicit negative/last-fetched marker so misses do not refetch the group list repeatedly.
- Safe miss caching needs explicit inbound-media invalidation: `whitenoise-mac/Core/WorkspaceState+Timeline.swift:227-308`, `whitenoise-mac/Core/WorkspaceState+Timeline.swift:432`. The PR invalidates the group cache after outbound media sends, but timeline projection upserts for newly arrived inbound media never clear the cached index. Current correctness depends on the miss path re-fetching, which conflicts with the first required fix. Add an invalidation hook when timeline upserts include media attachments (or another media-record change signal) and cover it with a regression test.

SUGGESTION:
- Strengthen the new coalescing test so it proves the resolved reference is actually used for download: `whitenoise-macTests/whitenoise_macTests.swift:2651-2767`, `whitenoise-macTests/whitenoise_macTests.swift:12010-12018`. The fake `downloadMedia` looks up downloads by `plaintextSha256`, and the epoch-0 timeline reference shares the same SHA as the full record reference, so the test passes even if `resolvedMediaReference` never swaps in the nonzero `sourceEpoch`. Capture the reference passed to `downloadMedia` and assert the resolved `sourceEpoch`.
- `clearMediaReferenceResolutionCache(forAccountId:)` bumps the global generation counter while deleting only that account's keys: `whitenoise-mac/Core/WorkspaceState+Media.swift:572-583`. If a non-active account with cached media is removed while the active account has an in-flight index build, that active-account result will be discarded and refetched later. A per-key/account generation or no global bump in the filtered clear would avoid cross-account cache churn.

NITPICK:
- Group leave and local delete paths do not use the new per-group invalidator (`WorkspaceState+Groups.swift`). Not a correctness problem, but it would keep the cache lifecycle symmetric with outbound send/account teardown.

SUMMARY: The coalescing shape is main-actor serialized and the hit path avoids the old per-attachment linear scan, but the miss path still repeats the full-list work and the natural fix for that exposes a stale-inbound-index hole. Requesting changes for those in-scope performance/correctness edges before merge.

SENSITIVE_PATHS: `whitenoise-mac/Core/WorkspaceState+Account.swift` touches account removal/sign-out/switch cleanup. No crypto/key material is changed directly, but merge gate should treat the account-lifecycle touch as sensitive.

agent-p1p:
Addressed the adversarial review findings in 4934bcdd8eec606c363ddaab4b4983ad9764d057:

- cached fetched media-reference indexes as authoritative for misses until invalidated, so sequential no-match lookups no longer re-run `listMedia(limit: nil)`;
- invalidated the per-group media-reference index on inbound media timeline upserts, outbound sends, group leave, and local group delete;
- removed the global-generation discard from unrelated account/group clears to avoid cross-account cache churn;
- strengthened the coalescing test to assert `downloadMedia` receives the resolved nonzero `sourceEpoch`, and added regressions for cached misses plus inbound invalidation.

CI is green on the latest head.

PR #341, opened 2026-07-04
Conform to the darkmatter→mdk rename and add marmot:// deep links

erskingardner:
## What changed

Adopts [marmot-protocol/mdk#725](https://github.com/marmot-protocol/mdk/pull/725) — the repo-wide retirement of the experimental "darkmatter" naming (clean break, no compat shims).

Closes #340

### Bindings bump & build tooling
- `Vendored/MarmotKit` rebuilt from **mdk `e712232`** (was darkmatter `745959e`) via `scripts/sync-bindings.sh`, now driven by `MDK_DIR` (default `~/code/mdk`). `MARMOT_VERSION` stamps `mdk-sha:`/`mdk-branch:`; the provenance constant is `MarmotKitVersion.mdkSHA`.
- Binding artifact names (`MarmotKit.xcframework`, package/target `MarmotKit`) are unchanged upstream by design — no project-file churn.
- The bump also picked up the new `selfMembership: SelfMembershipFfi` field on `AppGroupRecordFfi`/`ChatListRowFfi`; test fixtures construct it as `.member`. (Surfacing left/removed state in the UI is follow-up work.)
- AGENTS.md, README, and .gitignore now reference `marmot-protocol/mdk`.

### marmot:// deep links (full #340)
- `CFBundleURLTypes` registers the `marmot` scheme; `.onOpenURL` routes through `WorkspaceState.handleDeepLinkURL`, with a pending-reference queue for links that arrive before bootstrap reaches ready (cold start / signed out), flushed from `activateReadyState()`.
- New `MarmotProfileLink` helper is the single source of truth for the strict `marmot://profile/<npub|nprofile>` form. Inbound URLs are treated as untrusted (the scheme is not exclusive): only the profile form routes, everything else is dropped with a status hint.
- Kit-emitted `marmot://profile/...` message autolinks are now consumed in-app via `MarkdownLinkPolicy` (never handed to LaunchServices).
- The public-identity QR now encodes `marmot://profile/<npub>?from=qr` (visible text and Copy button keep the bare npub).

### Accepted clean breaks (per mdk#725)
- `darkmatter://` links and QR codes are no longer recognized anywhere.
- The kit's keychain service moved `com.marmot.darkmatter` → `com.marmot.whitenoise` with no FFI override or migration: **previously signed-in users must re-import their nsec** after this update.

### Telemetry build-config rename (app-local alignment)
[...]

Issue #311, opened 2026-07-03
Privacy/Bug: navigating to Settings (or opening new-chat) does not stop an in-progress voice recording — mic stays hot with no way to stop it


erskingardner:
## Severity
**MEDIUM** (privacy / bug)

## Summary
The voice recorder (`AVAudioRecorder`) and its metering task live on `WorkspaceState`, and the only stop/cancel UI (`VoiceRecordingComposerView`) is rendered inside the conversation composer, which is shown only when a chat is selected. `selectChat` and `prepareForActiveAccountSwitch` both call `cancelVoiceRecording()`, but `showSettings` and `showNewChat` do **not**. Navigating to Settings sets `selection = .settings(...)`, removing the conversation composer (and its Stop/Cancel buttons) from the hierarchy while the recorder keeps running — microphone live, plaintext audio still being written, no visible indicator and no control to stop it.

## Evidence
`whitenoise-mac/Core/WorkspaceState+Navigation.swift`:
```swift
func selectChat(_ chat: ChatItem) {
    cancelVoiceRecording()          // ✅ cancels
    ...
}

func showNewChat() {                // ❌ no cancelVoiceRecording()
    isNewChatComposerVisible = true
    lastError = nil
    resetNewChatComposer()
}

func showSettings(_ page: SettingsPage = .profile) {   // ❌ no cancelVoiceRecording()
    stopTimelineListener()
    clearEnteredLoginIdentity()
    selection = .settings(page)
    closeNewChatComposer()
    pruneMessageCache(keeping: nil)
}
```
`cancelVoiceRecording()` is referenced only in `selectChat` (`WorkspaceState+Navigation.swift:20`), `prepareForActiveAccountSwitch` (`WorkspaceState+Account.swift:82`), and the composer's Cancel button. The recording UI is gated on `workspace.isRecordingVoiceMessage` inside the conversation composer (`MessengerShellView.swift:463`), which is not present on the Settings/new-chat panes. There is no `onDisappear` cancellation.

## Failure scenario
[...]

PR #343, opened 2026-07-04
feat: surface self-membership state for left/removed groups

erskingardner:
## What

The mdk e712232 bindings (vendored in #341) expose a new `selfMembership: SelfMembershipFfi` (`.member` / `.left` / `.removed`) on both `ChatListRowFfi` and `AppGroupRecordFfi` — whether the local account is still a member of a group, and if not, whether it left voluntarily or was removed. The app previously ignored it. This maps the field through `Core/MarmotMapping.swift` into the view models and surfaces it everywhere the user can act on the chat:

- **Model**: new app-side `ChatSelfMembership` enum with presentation helpers (`sidebarBadgeLabel`, `endedDescription`, `endedSymbolName`); `ChatItem.selfMembership` / `isNoLongerMember` and `GroupDetailsSnapshot.selfMembership`. The field survives both chat-list merge paths the same way `pendingConfirmation` does (fresh-row wins on row updates, current wins during metadata enrichment).
- **Sidebar**: a "Left" / "Removed" capsule badge on chat rows, mirroring the existing `PendingInviteBadge`, with the full explanation as tooltip.
- **Conversation**: for non-member groups the composer (reply/media strips included) is replaced by an explanatory notice — "You left this group" / "You were removed from this group" — since the chat stays listed for reading history but the core rejects sends with `invalid_transition` (see wn-cli changelog). `canSend` and `sendDraft` gate on membership as defense in depth, and the header's "Set group image" button (a commit under the hood) is hidden.
- **Group details**: a "Membership" explanatory section plus a "Self membership" developer-diagnostics row.
- **Edge case** per the #311/#325 "mic must never stay hot without a stop button" rule: a membership flip that lands while recording a voice message in that chat tears the recorder down (`cancelVoiceRecordingIfSelectedMembershipEnded`, hooked into both chat-list update paths).

## Why

Upstream, mdk#724 clears the removed marker when branch selection supersedes a removal, and sends to self-evicted groups fail with `invalid_transition` — without this state the app offered a composer whose sends could only fail, with no explanation of why the group went quiet.

## Reviewer notes

- The fake runtime's `chatListRow(for:)` now derives `selfMembership` from the installed `AppGroupRecordFfi` instead of hardcoding `.member`, so tests can drive the full pipeline.
- Tests (317 total, all passing): FFI→`ChatItem` variant mapping, presentation labels, metadata-merge preservation, `GroupDetailsSnapshot` mapping, an end-to-end bootstrap test proving a `.removed` group stays listed but cannot send, and the recorder-teardown test.
- New user-facing strings are not yet in `Localizable.xcstrings`, matching existing practice for recent keys (e.g. "Group invite pending"); they fall back to English.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+ChatList.swift
+++ b/whitenoise-mac/Core/WorkspaceState+ChatList.swift
@@ -208,8 +208,18 @@ extension WorkspaceState {
         clearComposerDrafts(for: Array(removedChatIds), accountId: account.id)
         setChats(sortedChatItems(chatItems), forAccountId: account.id)
         dismissGroupImagePickerIfSelectedChatUnavailable()
+        cancelVoiceRecordingIfSelectedMembershipEnded()
         ensureSelectedMessageTimelineStore()
         startChatListEnrichment(rows: rows, account: account)
+    }
+
+    /// A membership flip to left/removed swaps the selected chat's composer (its recording
+    /// Stop/Cancel controls included) for the membership-ended notice, so any chat-list
+    /// update that can carry that flip must also tear down an in-progress recorder — the
+    /// microphone must never stay hot with no visible way to stop it (#311).
+    func cancelVoiceRecordingIfSelectedMembershipEnded() {
+        guard isRecordingVoiceMessage, selectedChat?.isNoLongerMember == true else { return }
+        cancelVoiceRecording()
     }
 
     func applyChatListSubscriptionUpdate(
@@ -275,6 +285,7 @@ extension WorkspaceState {
         let needsInitialMetadata = !shouldEnrich && readStateRowNeedsMetadataEnrichment(row, current: current)
 
         upsertChat(chat, forAccountId: account.id)
+        cancelVoiceRecordingIfSelectedMembershipEnded()
         ensureSelectedMessageTimelineStore()
         if shouldEnrich {
             startChatListEnrichment(rows: [row], account: account, replacingCurrent: false)
@@ -418,7 +429,8 @@ extension WorkspaceState {
                 unreadCount: current.unreadCount,
                 unreadMentionCount: current.unreadMentionCount,
                 isDirect: enrichedItem.isDirect,
-                pendingConfirmation: current.pendingConfirmation
+                pendingConfirmation: current.pendingConfirmation,
+                selfMembership: current.selfMembership
             )
             guard next != current else { continue }
             if incremental {
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -155,7 +155,8 @@ nonisolated struct ChatListOrdering {
             unreadCount: chat.unreadCount,
             unreadMentionCount: chat.unreadMentionCount,
             isDirect: current.isDirect,
-            pendingConfirmation: chat.pendingConfirmation
+            pendingConfirmation: chat.pendingConfirmation,
+            selfMembership: chat.selfMembership
         )
     }
 
@@ -1269,8 +1270,11 @@ final class WorkspaceState {
     }
 
     var canSend: Bool {
+        // The core rejects sends to a group the local account left or was removed
+        // from (`invalid_transition`), so an ended membership disables sending
+        // even though the chat stays selectable for reading history.
         client != nil
-            && selectedChat != nil
+            && selectedChat?.isNoLongerMember == false
             && (!draftText.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
                 || !pendingMediaAttachments.isEmpty)
             && !isSending
@@ -1357,6 +1361,7 @@ final class WorkspaceState {
             adminIds: details.group.admins,
             archived: details.group.archived,
             pendingConfirmation: details.group.pendingConfirmation,
+            selfMembership: ChatSelfMembership(details.group.selfMembership),
             members: members,
             isSelfAdmin: managementState.isSelfAdmin,
             isLastAdmin: managementState.isLastAdmin,
--- a/whitenoise-mac/Views/ComposerViews.swift
+++ b/whitenoise-mac/Views/ComposerViews.swift
@@ -502,7 +502,7 @@ struct MembershipEndedComposerNotice: View {
                 Text(membership.endedDescription ?? "")
                     .font(.callout.weight(.semibold))
 
-                Text("You can keep reading the history, but new messages can't be sent.")
+                Text(ChatSelfMembership.endedHistoryExplanation)
                     .font(.caption)
                     .foregroundStyle(.secondary)
             }
--- a/whitenoise-mac/Views/ComposerViews.swift
+++ b/whitenoise-mac/Views/ComposerViews.swift
@@ -486,6 +486,36 @@ struct ReplyComposerContextView: View {
     }
 }
 
+/// Replaces the composer for a group the local account left or was removed from:
+/// the transcript stays readable, but the core would reject any send
+/// (`invalid_transition`), so no input is offered.
+struct MembershipEndedComposerNotice: View {
+    let membership: ChatSelfMembership
+
+    var body: some View {
+        HStack(alignment: .firstTextBaseline, spacing: 10) {
+            Image(systemName: membership.endedSymbolName ?? "")
+                .font(.system(size: 13, weight: .semibold))
+                .foregroundStyle(.secondary)
+
+            VStack(alignment: .leading, spacing: 2) {
+                Text(membership.endedDescription ?? "")
+                    .font(.callout.weight(.semibold))
+
+                Text("You can keep reading the history, but new messages can't be sent.")
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
+            }
+
+            Spacer(minLength: 0)
+        }
+        .padding(.horizontal, 12)
+        .padding(.vertical, 8)
+        .glassCard()
+        .accessibilityIdentifier("composer.membershipEnded")
+    }
+}
+
 struct NewChatColumnView: View {
     @Environment(WorkspaceState.self) private var workspace
     @FocusState private var isSearchFocused: Bool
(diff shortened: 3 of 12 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define ChatSelfMembership and update models](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343/chapters/1) |
| 2 | [Map FFI membership state to app models](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343/chapters/2) |
| 3 | [Gate sending and recording on membership status](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343/chapters/3) |
| 4 | [Surface membership state in the UI](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343/chapters/4) |
| 5 | [Update tests and fake runtime](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/343">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit cc2c71349d84458afcefbf8c3ef088182b2d0c1c on Jul 4, 2026 1:41pm UTC.</sub>

whitenoise-mac/Views/MessengerShellView.swift
```diff
@@ -447,83 +447,13 @@ private struct ConversationView: View {
...
-                        durationSeconds: workspace.voiceRecordingDurationSeconds,
-                        onCancel: workspace.cancelVoiceRecording,
-                        onStop: {
-                            Task { await workspace.finishVoiceRecording() }
-                        }
-                    )
+                // The core rejects sends to a group the local account left or was
+                // removed from (`invalid_transition`), so the whole composer —
+                // reply/media drafts included — gives way to an explanatory notice.
+                if chat.isNoLongerMember {
```
erskingardner, line 453: 🔴 **Drag-and-drop / file import bypass the membership gate.**

When `chat.isNoLongerMember`, the composer is replaced by `MembershipEndedComposerNotice`, but `.fileImporter` and `.dropDestination` further down in this view (unchanged lines ~478–495) stay active and still call `workspace.addMediaAttachments(from:)`.

`addMediaAttachments` only checks `canBeginMediaAttachmentSelection()` (client + selected chat + slot count) — not membership — so a user can silently accumulate pending attachments they can never send and never see (the pending-media strip lives inside the hidden `composerControls`).

Consider gating the modifiers on `!chat.isNoLongerMember`, and/or adding `!selectedChat.isNoLongerMember` to `canBeginMediaAttachmentSelection()` / `startVoiceRecording()` for defense in depth.

whitenoise-mac/Views/GroupViews.swift
```diff
@@ -114,6 +114,25 @@ struct GroupDetailsSheet: View {
...
                     }
 
+                    if let endedDescription = snapshot.selfMembership.endedDescription {
+                        Section("Membership") {
+                            Label {
+                                VStack(alignment: .leading, spacing: 2) {
+                                    Text(endedDescription)
+                                        .font(.callout.weight(.semibold))
+
+                                    Text("You can keep reading the history, but new messages can't be sent.")
```
erskingardner: 🟡 **Group details still exposes mutating actions for non-members.**

The conversation header correctly hides "Set group image" when `chat.isNoLongerMember`, but the Profile section below (Save profile, "Search Web Image", disappearing-messages picker, invite section when `canInvite`, etc.) remains fully enabled.

If the core rejects these the same way it rejects sends (`invalid_transition`), users get opaque error toasts with no explanation — the same UX failure mode this PR addresses for the composer.

Consider disabling or hiding write affordances in group details when `snapshot.selfMembership != .member`, mirroring the composer/header treatment.

whitenoise-mac/Views/ComposerViews.swift
```diff
@@ -486,6 +486,36 @@ struct ReplyComposerContextView: View {
...
+        HStack(alignment: .firstTextBaseline, spacing: 10) {
+            Image(systemName: membership.endedSymbolName ?? "")
+                .font(.system(size: 13, weight: .semibold))
+                .foregroundStyle(.secondary)
+
+            VStack(alignment: .leading, spacing: 2) {
+                Text(membership.endedDescription ?? "")
+                    .font(.callout.weight(.semibold))
+
+                Text("You can keep reading the history, but new messages can't be sent.")
```
erskingardner: 🟡 **Hardcoded English subcopy (also duplicated in `GroupViews.swift:124`).**

The PR description notes new strings aren't in `Localizable.xcstrings` yet, but this sentence ("You can keep reading the history…") is user-facing in two places and doesn't go through `L10n.string` unlike the badge/notice titles. Easy to miss when localization is eventually swept.

whitenoise-mac/Views/SidebarViews.swift
```diff
@@ -350,6 +350,9 @@ struct ChatRowContent: View {
                     if chat.pendingConfirmation {
                         PendingInviteBadge()
                     }
+                    if chat.isNoLongerMember {
```
erskingardner, line 353: 🟢 **Both badges can render simultaneously.**

If the FFI ever delivers `pendingConfirmation == true` alongside a non-`.member` `selfMembership`, the row shows both "Invite" and "Left"/"Removed". Probably impossible in practice, but if it isn't an invariant upstream it could confuse — worth confirming with mdk or prioritizing one badge.

whitenoise-mac/Core/WorkspaceState+ChatList.swift
```diff
@@ -208,10 +208,20 @@ extension WorkspaceState {
         clearComposerDrafts(for: Array(removedChatIds), accountId: account.id)
         setChats(sortedChatItems(chatItems), forAccountId: account.id)
         dismissGroupImagePickerIfSelectedChatUnavailable()
+        cancelVoiceRecordingIfSelectedMembershipEnded()
```
erskingardner, line 211: 🟢 **Recorder teardown is tested on `applyChatRow` only.**

`cancelVoiceRecordingIfSelectedMembershipEnded()` is also wired into the bulk `applyChatRows` path here. The #311 regression test covers the single-row subscription path; a sibling test through `applyChatRows` would lock in the bulk snapshot/reconnect path too.

whitenoise-mac/Core/WorkspaceState.swift
```diff
@@ -1269,8 +1270,11 @@ final class WorkspaceState {
     }
 
     var canSend: Bool {
+        // The core rejects sends to a group the local account left or was removed
+        // from (`invalid_transition`), so an ended membership disables sending
+        // even though the chat stays selectable for reading history.
         client != nil
-            && selectedChat != nil
+            && selectedChat?.isNoLongerMember == false
```
erskingardner, line 1277: 🟢 **`canSend` nil-selection semantics changed shape (still correct).**

`selectedChat != nil` → `selectedChat?.isNoLongerMember == false` evaluates to `false` when `selectedChat` is nil, so behavior is preserved. Just flagging for readers — the optional chaining form is slightly less obvious than an explicit nil check combined with membership.

erskingardner (commented):
## Adversarial review summary

Solid core work: FFI mapping, merge semantics (fresh row on updates / preserved on enrichment), composer replacement, `canSend`/`sendDraft` gating, header image button, and the #311 recorder teardown are all coherent and well-tested.

I found a few gaps where the UI still exposes **send-adjacent** affordances after membership ends — the same class of problem this PR fixes for the composer. Details in inline comments.

**Severity key:** 🔴 should fix before merge · 🟡 worth fixing / follow-up · 🟢 nit

whitenoise-mac/Views/MessengerShellView.swift
erskingardner: Fixed in cc2c713: `canBeginMediaAttachmentSelection()` now requires `selectedChat?.isNoLongerMember == false` (covers file import, drops, and voice recording), the `.dropDestination` closure refuses drops and suppresses the drop-target highlight for ended chats, and the end-to-end test asserts `addMediaAttachments` leaves `pendingMediaAttachments` empty.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

whitenoise-mac/Views/GroupViews.swift
erskingardner: Fixed in cc2c713: the Profile section (fields, Save, Search Web Image) is `.disabled` for non-members, the retention picker gains the membership condition ("Delete expired now" stays enabled — local prune), and the Invite section is hidden; Archive/Remove-From-This-Device stay enabled as local-only actions, and Leave/Step-Down were already FFI-gated via `canLeave`/`isSelfAdmin`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

whitenoise-mac/Views/ComposerViews.swift
erskingardner: Fixed in cc2c713: the subcopy is now `ChatSelfMembership.endedHistoryExplanation` (through `L10n.string`), shared by the composer notice and the group-details section.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

whitenoise-mac/Views/SidebarViews.swift
erskingardner: Fixed in cc2c713: the ended-membership badge now supersedes the invite badge (`if isNoLongerMember ... else if pendingConfirmation`), so at most one renders even if the FFI ever delivers both flags.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

whitenoise-mac/Core/WorkspaceState+ChatList.swift
erskingardner: Fixed in cc2c713: added `endedMembershipBulkRowsUpdateStopsInProgressVoiceRecording` driving the bulk `applyChatRows` path, sibling to the single-row test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

whitenoise-mac/Core/WorkspaceState.swift
erskingardner: No change — intentional: the optional chain folds the nil check and the membership check into one condition (`nil == false` → send disabled, same as before); the comment above it documents why membership gates sending.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #357, opened 2026-07-04
fix: coalesce concurrent timeline loads

### LINKED ISSUE #332: Bug/perf: concurrent loadMessages(groupIdHex:) for the same chat can double-subscribe the timeline and orphan a live subscription handle
## Severity
**LOW** (performance / bug — redundant FFI + transient leaked subscription)

## Summary
`loadMessages(groupIdHex:)` is invoked from untracked `Task { ... }`s at multiple call sites and its early idempotency guard only short-circuits once the store is already `isLoaded`. Two `loadMessages(X)` for the same freshly-selected chat that start close together (before either finishes loading) both pass the not-yet-loaded guard, both issue `subscribeTimelineMessages` + `snapshot()` (two full FFI round-trips + decrypt), and the second `startTimelineListener` tears down the first's listener — dropping the first subscription handle with no `next()` loop draining it.

## Evidence
Untracked spawns:
- `whitenoise-mac/Core/WorkspaceState+Navigation.swift:36` — `Task { await loadMessages(groupIdHex: chat.id) }`
- `whitenoise-mac/Core/WorkspaceState+ChatList.swift:334` — `Task { await loadMessages(groupIdHex: nextChat.id) }`

Guard body (`WorkspaceState+Timeline.swift:19-83`):
```swift
if timelineTaskGroupId == groupIdHex, ensureMessageTimelineStore(for: groupIdHex).isLoaded {
    finishTimelineInitialLoad(groupIdHex: groupIdHex); return
}
```
The bail path at lines ~71-76 explicitly guards against an orphaned subscription, but the double-run path does not.

## Failure scenario
`selectChat(X)` runs while an earlier `removeChat`/`selectMostRecentChatIfNeeded`-spawned `loadMessages(X)` for the same chat is still in flight and neither has reached `isLoaded`. Both subscribe + snapshot; the second listener replaces the first; the first `TimelineSubscription` is dropped without being drained — redundant decryption/DB work plus a transient orphaned live handle.

## Suggested fix
Coalesce in-flight timeline loads per group (mirror the `reloadChatsTask`/`reloadChatsTaskAccountId` pattern) so a second same-group `loadMessages` awaits the existing load instead of re-subscribing.

## Scope note
Found during a full-codebase review. No filed issue addresses duplicate concurrent `loadMessages` subscription; the timeline issues on file concern pagination truncation/cancellation, not double-subscribe.

agent-p1p:
## Summary
- Coalesce overlapping `loadMessages(groupIdHex:)` calls for the same chat onto one in-flight timeline load.
- Move the actual subscription/snapshot pass into a single-owner task and cancel only superseded different-chat loads.
- Add a regression test that starts a duplicate same-chat load while the first subscription is suspended.

## Test Plan
- `git diff --check`
- `git grep -n '^<<<<<<<\\|^=======\\|^>>>>>>>' HEAD -- . ':!Vendored'` (no conflict markers)
- `xcodebuild -scheme whitenoise-mac -configuration Debug build-for-testing CODE_SIGNING_ALLOWED=NO` (not run: `xcodebuild` is unavailable on this Linux worker)

Fixes #332

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Account.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Account.swift
@@ -84,6 +84,7 @@ extension WorkspaceState {
     ) {
         leaveActiveConversation()
         stopTimelineListener()
+        cancelTimelineLoad()
         cancelChatListReload()
         stopChatListListener()
         clearEnteredLoginIdentity()
@@ -244,6 +245,7 @@ extension WorkspaceState {
         do {
             if wasActive {
                 stopTimelineListener()
+                cancelTimelineLoad()
                 cancelChatListReload()
                 stopChatListListener()
             }
@@ -307,6 +309,7 @@ extension WorkspaceState {
     /// drafts, settings snapshots). Shared by account removal and sign-out.
     func resetActiveAccountUIState() {
         stopTimelineListener()
+        cancelTimelineLoad()
         cancelChatListReload()
         stopChatListListener()
         cachedMessageChatIds.removeAll()
@@ -350,6 +353,7 @@ extension WorkspaceState {
         do {
             if wasActive {
                 stopTimelineListener()
+                cancelTimelineLoad()
                 cancelChatListReload()
                 stopChatListListener()
             }
--- a/whitenoise-mac/Core/WorkspaceState+Timeline.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Timeline.swift
@@ -17,6 +17,32 @@ import UserNotifications
 @MainActor
 extension WorkspaceState {
     func loadMessages(groupIdHex: String) async {
+        if timelineTaskGroupId == groupIdHex, ensureMessageTimelineStore(for: groupIdHex).isLoaded {
+            finishTimelineInitialLoad(groupIdHex: groupIdHex)
+            return
+        }
+
+        if let existing = timelineLoadTask, timelineLoadGroupId == groupIdHex {
+            await existing.value
+            return
+        }
+
+        timelineLoadTask?.cancel()
+        let task = Task<Void, Never> { [weak self] in
+            await self?.performTimelineLoad(groupIdHex: groupIdHex)
+        }
+        timelineLoadTask = task
+        timelineLoadGroupId = groupIdHex
+
+        await task.value
+
+        if timelineLoadTask == task {
+            timelineLoadTask = nil
+            timelineLoadGroupId = nil
+        }
+    }
+
+    func performTimelineLoad(groupIdHex: String) async {
         guard let client, let activeAccount else {
             finishTimelineInitialLoad(groupIdHex: groupIdHex)
             return
@@ -77,6 +103,8 @@ extension WorkspaceState {
             guard timelineTaskGroupId == groupIdHex else { return }
             activeTimelineSubscription = subscription
             activeTimelineGroupId = groupIdHex
+        } catch is CancellationError {
+            return
         } catch {
             lastError = error.localizedDescription
         }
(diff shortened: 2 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define state for coalescing timeline loads](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357/chapters/1) |
| 2 | [Implement concurrent timeline load coalescing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357/chapters/2) |
| 3 | [Cancel timeline loads on account changes](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357/chapters/3) |
| 4 | [Verify timeline coalescing with regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/357">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 028a08e4e2ccb162c9b01dfc2e06ad02a73889fa on Jul 4, 2026 8:18pm UTC.</sub>

agent-p1p:
Pip adversarial review for PR #357.

VERDICT: REQUEST_CHANGES

BLOCKING:
- Coalescing key omits account identity: `whitenoise-mac/Core/WorkspaceState+Timeline.swift:25`. The new in-flight join checks only `timelineLoadGroupId == groupIdHex`. The issue suggested mirroring the chat-list reload pattern, but that pattern keys by account id and generation (`WorkspaceState+ChatList.swift:31,41-42,99-105`). MLS group ids are shared across members, so two local accounts can both be in group X. If account A has `loadMessages(X)` suspended, then the user switches to account B before that load completes, `prepareForActiveAccountSwitch` changes `activeAccountId` but does not clear `timelineLoadTask`; B's `loadMessages(X)` will await A's stale task and return without ever subscribing/rendering B's timeline. Pre-PR behavior would at least start B's own subscription. Add account identity to the coalescing key, or cancel/clear timeline-load ownership on account teardown; preferably do both, matching `reloadChatsTaskAccountId` + generation.

SUGGESTION:
- Add a `cancelTimelineLoad()` teardown path. `prepareForActiveAccountSwitch`, `resetActiveAccountUIState`, sign-out/removal paths all call `cancelChatListReload()` / `stopTimelineListener()`, but `stopTimelineListener()` does not touch `timelineLoadTask` or `timelineLoadGroupId`. The new coalescing state should not outlive the account/selection lifecycle it belongs to.
- Add coverage for account switch while a timeline load is in flight. The new test proves same-account coalescing, but not the stale-account case above. A regression can suspend account A's `loadMessages(X)`, switch to account B that also has group X, then assert B opens its own timeline subscription and renders B's messages.
- The `isLoaded` short-circuit is now duplicated in `loadMessages` and `performTimelineLoad`. The defensive inner check is reasonable, but a short comment would make the intentional double-check clear.

NITPICK:
- The duplicate-load path relies on the owner task's `defer { finishTimelineInitialLoad(...) }` to clear the spinner. That is correct today; a comment would make the dependency explicit.

SUMMARY: The same-account double-subscribe fix is directionally right and the regression test is meaningful. The remaining gap is ownership: the in-flight task is keyed only by group id and is not canceled on account teardown, so a stale load from one account can satisfy a same-group request from another account. That is in scope for this fix.

SENSITIVE_PATHS: none.

agent-p1p:
Addressed in 028a08e:

- timeline load coalescing is now keyed by account + group with a generation guard
- account teardown paths cancel/clear in-flight timeline load ownership
- added `switchedAccountTimelineLoadDoesNotJoinStaleSameGroupLoad` for the stale same-group/account-switch case
- added comments for the intentional inner `isLoaded` guard and same-account spinner ownership

PR Checks and Static Analysis are green on the latest run. The earlier red was a swift-format lint failure on my first address-review commit; fixed before this green rerun.

Issue #334, opened 2026-07-04
Bug: MessageMediaDiskCache per-account purge deletes an entry on ANY metadata read/decrypt failure, causing cross-account collateral cache deletion


erskingardner:
## Severity
**LOW** (bug — incorrect eviction; self-heals via re-download)

## Summary
`removeEntries(matchingAccountDigest:)` walks **every** cache entry across all accounts (they share one symmetric key) and, in its failure branch, deletes the entry rather than skipping it. So a transient I/O error or a concurrent read on *another* account's `metadata.bin` during a single-account purge deletes that unrelated account's valid cache entry.

## Evidence
`whitenoise-mac/Core/MessageMediaDiskCache.swift:549-566`:
```swift
for entryDirectory in entryDirectories {
    let cacheID = entryDirectory.lastPathComponent
    let metadataURL = entryDirectory.appendingPathComponent(metadataFileName)
    guard let metadataData = try? Data(contentsOf: metadataURL),
        let metadataPlaintext = try? open(metadataData, using: symmetricKey,
                                          authenticatedBy: metadataAAD(for: cacheID)),
        let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext)
    else {
        try? FileManager.default.removeItem(at: entryDirectory)   // <-- deletes on ANY read failure
        continue
    }
    if metadata.accountDigest == accountDigest {
        try? FileManager.default.removeItem(at: entryDirectory)
    }
}
```
The `else` cannot distinguish "this entry belongs to the purged account" from "I momentarily failed to read an entry belonging to a *different* account."

## Failure scenario
Accounts A and B both have cached media. User signs out / purges account A. During A's purge, a read of one of B's `metadata.bin` files hits a transient I/O error (or races a concurrent open). The `else` branch deletes B's valid, unrelated entry. Impact is bounded (B re-downloads on next view), but it is cross-account collateral deletion triggered by activity unrelated to the purge target.

## Suggested fix
In a per-account purge, treat a bare read failure as "skip, not mine" — `continue` without deleting. Leave corrupt-entry reclamation to `purgeAll` or a dedicated GC pass that is explicitly account-agnostic.

## Scope note
Found during a full-codebase review. No filed issue covers `removeEntries(matchingAccountDigest:)` collateral deletion.

PR #360, opened 2026-07-04
fix: clean corrupt media cache entries after account purge

### LINKED ISSUE #356: Bug: Corrupt media cache entries can survive account purge indefinitely
## Severity
LOW (bounded disk-cache orphan; encrypted data, full wipe clears it)

## Summary
PR #354 correctly fixes #334 by skipping unreadable `metadata.bin` entries during `purgeAccount(_:)` instead of deleting them blindly. That avoids cross-account collateral deletion, but it also means an unreadable entry that actually belongs to the account being removed cannot be identified and will survive the per-account purge.

There does not appear to be a dedicated GC/size-budget eviction path for these unreadable entries. The existing corrupt-entry cleanup in `readDownload` only runs when the exact cache key is read again; after an account is removed, those keys normally will not be read. `purgeAll` / delete-all-local-data still clears everything.

## Evidence
- `whitenoise-mac/Core/MessageMediaDiskCache.swift:575-580` now skips unreadable metadata during per-account purge.
- `whitenoise-mac/Core/MessageMediaDiskCache.swift:456-458` deletes corrupt entries only on `cachedDownload` / read of that exact key.
- `WorkspaceState+Account.swift:252` calls `mediaDiskCache.purgeAccount(removedAccountId)` during account removal.

## Impact
A corrupt encrypted media cache entry for a removed account can remain on disk until a full local-data wipe. No plaintext is exposed, and the impact is bounded to disk usage / stale cache artifacts.

## Why this is separate from #334
Deleting unreadable entries inside a per-account purge caused the #334 cross-account data-loss bug because the purge could not prove the unreadable entry belonged to the target account. A safe fix for this follow-up likely needs an explicitly account-agnostic cleanup pass, a cache size/age eviction policy, or another design that does not reintroduce per-account collateral deletion.

## Suggested direction
Add a safe account-agnostic media-cache cleanup path for undecodable entries, or explicitly document the orphan behavior as an accepted limitation if no GC is desired.

agent-p1p:
## Summary
- run an account-agnostic corrupt metadata cleanup pass after per-account media cache purge
- keep the per-account purge guard: entries with metadata that cannot be read at all are skipped instead of treated as target-account matches
- update the media disk cache regression to prove readable entries for other accounts survive while corrupt metadata entries are reclaimed

## Test Plan
- `git diff --check`
- GitHub: PR Checks pass
- GitHub: Static Analysis pass
- GitHub: CodeRabbit status success, but the posted comment says review was rate-limited and no actionable findings were produced
- Local `xcodebuild -scheme whitenoise-mac -configuration Debug build-for-testing CODE_SIGNING_ALLOWED=NO` unavailable on this Linux host (`xcodebuild: command not found`)
- Local `swift --version` unavailable on this Linux host (`swift: command not found`)

Fixes #356

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved account-based cache cleanup to remove leftover corrupted media entries during purge.
  * Ensured valid media from other accounts remains available after clearing one account’s cache.
  * Better handling of unreadable cache data so it no longer lingers in storage.

### DIFF
--- a/whitenoise-mac/Core/MessageMediaDiskCache.swift
+++ b/whitenoise-mac/Core/MessageMediaDiskCache.swift
@@ -687,9 +687,13 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                 using: symmetricKey,
                 authenticatedBy: metadataAAD(for: cacheID)
             ),
-            let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext),
-            metadata.version == 1
+            let metadata = try? JSONDecoder().decode(Metadata.self, from: metadataPlaintext)
         else { return .corrupt }
+        guard metadata.version == 1 else {
+            // A decryptable, decodable record with a different version may belong to a
+            // newer cache format. Skip it instead of reclaiming it account-agnostically.
+            return .unavailable
+        }
         return .readable(metadata)
     }
 
--- a/whitenoise-mac/Core/MessageMediaDiskCache.swift
+++ b/whitenoise-mac/Core/MessageMediaDiskCache.swift
@@ -322,6 +322,10 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                 root: root,
                 symmetricKey: symmetricKey
             )
+            // Keep account matching separate from corrupt-entry reclamation. The first
+            // pass only removes entries it can attribute to the target account; this
+            // second pass is account-agnostic and deletes entries no account can read.
+            Self.removeUnreadableEntries(root: root, symmetricKey: symmetricKey)
         }
         await task.task.value
         finishPurge(task)
@@ -449,6 +453,12 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
     private struct CacheFootprint {
         let entryCount: Int
         let byteCount: UInt64
+    }
+
+    private enum MetadataStatus {
+        case readable(Metadata)
+        case corrupt
+        case unavailable
     }
 
     private static func readDownload(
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define metadata status types](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/360/chapters/1) |
| 2 | [Implement corrupt metadata detection and cleanup](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/360/chapters/2) |
| 3 | [Wire cleanup pass into account purge](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/360/chapters/3) |
| 4 | [Update and expand cache regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/360/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/360">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2de745f8110ead9ff7b65ffed6462c4787e44014 on Jul 4, 2026 10:32pm UTC.</sub>

agent-p1p:
## Pip adversarial review — REQUEST_CHANGES equivalent

I reviewed the live PR state, diff, and surrounding media-cache purge/eviction code. CI is currently green, and `git diff --check origin/master...HEAD` passes locally. This PR touches sensitive encrypted media-cache deletion logic, so the regression guard needs to be tight.

BLOCKING:
- `.unavailable` guard is untested: `whitenoise-mac/Core/MessageMediaDiskCache.swift:674-682`. The safety property that keeps this from reintroducing #334 depends on `metadataStatus` returning `.unavailable` when `Data(contentsOf:)` throws, so `removeUnreadableEntries` skips the entry instead of deleting another account's readable cache because of a transient metadata read failure. The new test only covers the deterministic decrypt/decode failure path by overwriting `metadata.bin` with plaintext. Add a regression test where metadata exists but `Data(contentsOf:)` throws (for example, replace `metadata.bin` with a directory) and assert the other-account entry survives `purgeAccount(_:)`.

SUGGESTION:
- Account-agnostic corrupt cleanup is currently coupled to account removal and full-scans the tree twice per purge: `whitenoise-mac/Core/MessageMediaDiskCache.swift:319-329` and `640-667`. This fixes the reported post-removal orphan path, but it is an awkward home for a general cache-GC operation. Consider making the corrupt sweep an explicit maintenance helper / shared scan path, or folding it into the existing eviction-maintenance flow if that can stay simple.
- Future metadata versions are now deleted account-agnostically: `whitenoise-mac/Core/MessageMediaDiskCache.swift:684-692`. `metadata.version != 1` is classified as `.corrupt`, so any future decryptable v2 cache entry would be removed on account purge by an older build. No live issue while only v1 exists, but this is a forward-compat sharp edge worth making intentional.

NITPICK:
- Corrupt-entry deletion is now repeated in three places (`readDownload`, `cacheEntries`, and `removeUnreadableEntries`). `MetadataStatus` centralizes classification nicely; a tiny helper for `removeItem` + `removeEmptyDirectory(shardDirectory)` would centralize the deletion invariant too.

SUMMARY: The core approach is directionally correct: readable entries for other accounts survive, while deterministically corrupt metadata can be reclaimed without attributing it to a specific account. I am requesting changes because the branch that prevents the #334 collateral-deletion regression is the one branch not covered by the test suite.

SENSITIVE_PATHS: `whitenoise-mac/Core/MessageMediaDiskCache.swift` — encrypted media disk cache, AES-GCM metadata/payload handling, account-scoped deletion/GC.

agent-p1p:
@coderabbitai review

agent-p1p:
Address-review update:

- `f6c8e89` adds the missing regression for `.unavailable`: replacing another account's `metadata.bin` with a directory makes `Data(contentsOf:)` fail, and `purgeAccount(_:)` preserves that entry while removing the target account's entry.
- `2de745f` addresses the CodeRabbit/version nit: decryptable+decodable metadata with an unsupported version is now treated as non-reclaimable (`.unavailable`) instead of account-agnostic `.corrupt`, with a v2 metadata regression test.
- I left the broader maintenance-scan/deletion-helper suggestions out of scope to keep this low-risk cache-purge fix focused.

Checks are green: PR Checks, Static Analysis, and CodeRabbit status. CodeRabbit's latest full re-review for `2de745f` was rate-limited; there are no inline comments, and the previous CodeRabbit nitpick is covered by the new commit.

PR #368, opened 2026-07-05
fix: avoid redundant media cache invalidation

### LINKED ISSUE #365: Performance: timeline media-reference resolution cache is wiped on every projection delta that touches a media message, including pure delivery-state ticks
## Severity
**LOW** (redundant media re-resolution during a media send burst)

## Summary
`applyTimelineProjection` clears the whole group's media-reference resolution cache whenever *any* upsert record in the delta "may contain media." Because a single media send emits a burst of projection deltas (the new row, then each delivery-state transition / relay echo / per-relay ack) and every one of those upserts re-carries the same message with its media fields intact, the predicate is true on every tick and the cache is invalidated N times — forcing every visible `sourceEpoch == 0` media tile to re-run full media resolution repeatedly instead of once.

## Evidence
- `whitenoise-mac/Core/WorkspaceState+Timeline.swift:262-264`:
  ```swift
  if upsertRecords.contains(where: { timelineRecordMayContainMediaAttachment($0) }) {
      clearMediaReferenceResolutionCache(forAccountId: account.id, groupIdHex: groupIdHex)
  }
  ```
- `timelineRecordMayContainMediaAttachment` (`:328-332`) returns true if `record.media` is non-empty, or `record.mediaJson` is non-empty, or a tag is `imeta` — all of which persist across delivery-state transitions of the same message.
- The doc comment at `:204-210` itself notes the delta burst: "the new row, then each delivery-state transition, the relay echo, per-relay acks."
- `clearMediaReferenceResolutionCache` bumps the generation and drops `mediaReferenceIndexes[key]` (`WorkspaceState+Media.swift:588-597`); the next render of each `sourceEpoch == 0` tile then misses and re-resolves, which per #286 is a `listMedia(limit: nil)` + linear scan per attachment.

## Failure scenario
Sending (or receiving) one media message triggers ~N delivery-state deltas; each invalidates the entire group's media index, so the visible media set is re-resolved ~N times during a single send even though the source-epoch mapping never changed across delivery transitions.

- #286 is about the *cost of one resolution* (unbounded `listMedia` per miss), not the invalidation trigger.
- #170 / #251 concern chat-list re-sort/enrichment on read-state deltas, not the timeline media-index cache.

## Suggested fix
Invalidate the media-reference cache only when a genuinely **new** media message id is introduced by the delta (or when a media record's reference-relevant fields actually change), rather than on every projection carrying an already-known media record.

*Found during a full-codebase review; verified against source.*

agent-p1p:
Fixes #365

## Summary
- Move timeline media-reference cache invalidation to after projection records are mapped to `MessageItem`s.
- Compare each mapped upsert's `mediaAttachments` against the pre-mutation timeline store entry, so known delivery-state/projection ticks that re-carry unchanged media fields do not clear the cache.
- Keep invalidation for new media message ids and media attachment changes, and add a regression test for the unchanged delivery-state tick path.

## Verification
- `git diff --check` ✅
- conflict-marker grep ✅
- `xcrun swift-format lint --configuration .swift-format --recursive --parallel --strict whitenoise-mac whitenoise-macTests whitenoise-macUITests` not run: `xcrun` is not installed on this Linux host.
- `scripts/ci/macos-sanity-checks.sh` not run: requires `xcodebuild`, not installed on this Linux host.
- `xcodebuild test -project whitenoise-mac.xcodeproj -scheme whitenoise-mac -testPlan PR -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` not run: `xcodebuild` is not installed on this Linux host.

## Sensitive paths
none

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/368">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Timeline.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Timeline.swift
@@ -259,10 +259,6 @@ extension WorkspaceState {
         }
         guard !upsertRecords.isEmpty || !removalIds.isEmpty else { return }
 
-        if upsertRecords.contains(where: { timelineRecordMayContainMediaAttachment($0) }) {
-            clearMediaReferenceResolutionCache(forAccountId: account.id, groupIdHex: groupIdHex)
-        }
-
         // Resolve senders for just the changed records (the common case is an all-cached
         // lookup) and map only those records — not the entire window.
         let senderProfiles = await TimelineSignpost.mapping.asyncInterval(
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Refine media cache invalidation logic in timeline](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/368/chapters/1) |
| 2 | [Verify cache persistence during delivery state updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/368/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/368">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit f5d65e8dd5471da85690e171101839f8bcae971e on Jul 5, 2026 3:45am UTC.</sub>

agent-p1p:
## Pip adversarial review

I reviewed the diff against #365, including the timeline projection path, `MessageTimelineStore` pre-mutation lookup behavior, media-reference cache invalidation, and the new regression coverage.

Findings:
- Blocking: 0
- Suggestions: 0
- Nitpicks: 0

Checks:
- Conflict-marker scan passed (`0` matches outside `Vendored/MarmotKit`).
- GitHub Actions are green for PR Checks and Static Analysis. PR Checks covered Swift format lint, macOS sanity checks, unit tests, and release build smoke.
- Local macOS checks could not be run on this Linux worker because `xcodebuild`/`xcrun` are unavailable; CI covered those paths.

Notes:
- Touched paths are client timeline/cache code and tests only; I did not see sensitive crypto/auth/MLS/key-handling paths in this diff.
- The implementation keeps the cache for unchanged media delivery-state ticks and still invalidates for new or changed mapped media attachments before the store mutation, which matches the issue's target behavior.

PR #370, opened 2026-07-05
fix: stop voice recording when removed chat reselects

### LINKED ISSUE #362: Bug/Privacy: removeChat auto-reselection does not stop an in-progress voice recording — mic stays hot and audio is misfiled to the newly-selected chat
## Severity
**MEDIUM** (hot microphone with no visible stop affordance + private-content misdirection)

## Summary
When the selected chat is removed out from under the user by a chat-list subscription delta, `removeChat` reselects a different conversation but never tears down a live voice recording. Unlike every other selection-changing path, it does not call `leaveActiveConversation()` / `cancelVoiceRecording()`. The microphone keeps recording, the recording controls now float over the *new* chat's composer, and — because `finishVoiceRecording` captures the draft key **after** the reselection — the audio recorded for chat A is filed under chat B.

## Evidence
- `removeChat(groupIdHex:account:)` sets `selection = nextChat.map { .chat($0.id) }` and spawns `loadMessages(nextChat.id)` — `whitenoise-mac/Core/WorkspaceState+ChatList.swift:341-346` — but never calls `leaveActiveConversation()`/`cancelVoiceRecording()`/`cancelGroupTranscriptExport()`.
- The centralized teardown's own contract says every selection-changing path *must* run it: `leaveActiveConversation()` doc comment at `whitenoise-mac/Core/WorkspaceState+Navigation.swift:19-25` ("Any navigation path that removes those from the hierarchy … must run this so the microphone can never stay hot"). `selectChat`/`showSettings`/`showNewChat` all call it (`Navigation.swift:32,43,55`); `removeChat` does not.
- Entry points into `removeChat`: an archived row (`applyChatRow` at `WorkspaceState+ChatList.swift:265-267`, which `return`s **before** the membership-ended cancel at `:288`), or a `removeRow` delta.
- `cancelVoiceRecordingIfSelectedMembershipEnded()` cannot save this: its guard is `isRecordingVoiceMessage, selectedChat?.isNoLongerMember == true` (`WorkspaceState+ChatList.swift:220-221`), but after reselection `selectedChat` is chat B (a normal member) or `nil`, so the guard is false — and it isn't called from `removeChat` anyway.
- `isRecordingVoiceMessage` is global (`WorkspaceState.swift:542`), so the live recording UI renders over chat B's composer.
- `finishVoiceRecording()` captures `let draftKey = selectedComposerDraftKey` at `WorkspaceState+Media.swift:337` — evaluated *after* the reselection, so it equals **chat B**. It then calls `appendPendingMediaAttachmentIfSelectionUnchanged(attachment, for: draftKey)` (`:360`), whose guard `selectedComposerDraftKey == draftKey` (`:395`) is `B == B` → **true** → the audio is appended to chat B. The #245 misdirection guard is defeated precisely because the selection changed before the draft key was captured.

## Failure scenario
1. User is recording a voice message in chat A.
2. Chat A is archived on another device (or otherwise removed via a chat-list delta) → `removeChat(A)` runs.
3. Selection flips to chat B; recording keeps going, controls now over B's composer. (If `nextChat == nil`, the mic simply stays hot indefinitely with no stop UI.)
4. User taps "finish" → chat A's recorded audio is prepared for send **to chat B**.

- #311 fixed the hot-mic teardown for the **Settings / new-chat** navigation entry points specifically; the subscription-driven `removeChat` auto-reselection path is not covered.
- #245 covers stale selection during the media-preparation *await*; here the selection is forced to change *before* the draft-key capture, so its guard doesn't engage.

## Suggested fix
Call `leaveActiveConversation()` (which cancels the recorder and the transcript export) inside `removeChat` before/at the point it mutates `selection`, so the removal path shares the same teardown invariant as the explicit navigation paths.

*Found during a full-codebase review; verified against source.*

agent-p1p:
## Summary
- cancel active conversation resources when `removeChat` removes the selected chat before auto-reselecting another chat
- add a regression test for an in-progress voice recording during removeChat auto-reselection

## Test plan
- `git diff --check`
- Not run: `xcodebuild` / Swift tests (xcodebuild, swift, and swiftc are unavailable on this Linux worker)

Fixes #362

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/370">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+ChatList.swift
+++ b/whitenoise-mac/Core/WorkspaceState+ChatList.swift
@@ -336,6 +336,7 @@ extension WorkspaceState {
             selectedGroupId == groupIdHex
         else { return }
 
+        leaveActiveConversation()
         closeGroupImagePicker()
         let nextChat = mostRecentChat(in: chats)
         selection = nextChat.map { .chat($0.id) }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -10020,6 +10020,25 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func removeSelectedChatStopsInProgressVoiceRecordingBeforeAutoReselection() throws {
+        // #362: subscription deltas can remove the selected chat and auto-select a different
+        // conversation. That implicit navigation must tear down the recorder before the composer
+        // belongs to the new chat, or finishing would file chat A's audio under chat B.
+        let state = WorkspaceState.preview()
+        let account = AccountItem.samples[0]
+        let removedChatId = try #require(state.selectedChat?.id)
+        let url = try armInProgressVoiceRecording(on: state)
+
+        state.removeChat(groupIdHex: removedChatId, account: account)
+
+        #expect(state.selection != .chat(removedChatId))
+        #expect(state.selectedChat != nil)
+        #expect(!state.isRecordingVoiceMessage)
+        #expect(state.voiceRecordingMeterTask == nil)
+        #expect(!FileManager.default.fileExists(atPath: url.path))
+    }
+
+    @MainActor
     @Test func resetToNewInstallStateStopsInProgressVoiceRecording() throws {
         // #311: the full local-data teardown must also release the recorder so a wipe cannot
         // leave the microphone active or plaintext audio writes running in the old state.
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -10020,14 +10020,22 @@ struct whitenoise_macTests {
     }
 
     @MainActor
-    @Test func removeSelectedChatStopsInProgressVoiceRecordingBeforeAutoReselection() throws {
+    @Test func removeSelectedChatStopsInProgressVoiceRecordingBeforeAutoReselection() async throws {
         // #362: subscription deltas can remove the selected chat and auto-select a different
-        // conversation. That implicit navigation must tear down the recorder before the composer
-        // belongs to the new chat, or finishing would file chat A's audio under chat B.
+        // conversation. That implicit navigation must tear down active conversation resources
+        // before the composer belongs to the new chat, or finishing would file chat A's audio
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Cancel active conversation resources on chat removal](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/370/chapters/1) |
| 2 | [Verify voice recording stops during chat reselection](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/370/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/370">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit dc68519ea17a6d728af86dec655b082ca36615cd on Jul 5, 2026 4:13am UTC.</sub>

agent-p1p:
Pip adversarial review for #370 / #362.

BLOCKING:
- none

SUGGESTION:
- Add an explicit transcript-export cancellation assertion on this path: `whitenoise-macTests/whitenoise_macTests.swift:9839` — the fix correctly routes through `leaveActiveConversation()`, which cancels both voice recording and group transcript export. The new regression test locks down the hot-mic/audio-misdirection side. A small companion assertion with an armed `groupTranscriptExportTask` would lock down the full centralized-teardown contract for `removeChat` specifically. Non-blocking.

NITPICK:
- none

SUMMARY: The fix is correct and minimal. `removeChat(groupIdHex:account:)` now calls `leaveActiveConversation()` after confirming the removed chat is the current selection and before mutating `selection`, so the recorder is cancelled before `finishVoiceRecording()` can capture the newly-selected chat's draft key. The regression test is meaningful and would have failed before the fix because the recording state, meter task, and plaintext temp file survived auto-reselection. Scope is clean: one production line plus a targeted test.

VERDICT: APPROVE
SENSITIVE_PATHS: none per repo merge-gate rubric; changed client navigation/recording state only.

agent-p1p:
Addressed the non-blocking transcript-export suggestion in dc68519: the #362 regression now arms groupTranscriptExportTask before removeChat() and asserts the shared leaveActiveConversation() teardown cancels it, alongside the voice-recorder checks.

PR #416, opened 2026-07-08
fix: prune media downloads with timeline window

### LINKED ISSUE #394: Performance/Privacy: decrypted attachment plaintext in mediaDownloads is never evicted within a conversation — a long media chat accumulates unbounded resident plaintext
## Severity
**MEDIUM** (memory growth + privacy). Every attachment ever downloaded in the *open* conversation stays in RAM as full decrypted plaintext for the whole time that chat is open — including messages that have since scrolled out of the timeline window. Hundreds of MB of decrypted image/video plaintext can accumulate, and decrypted bytes for no-longer-visible messages remain resident.

## Summary
The only eviction path for `mediaDownloads` (`pruneMediaDownloadCache`) drops entries for **other** groups; it never prunes entries for the currently-open group. As the timeline window is trimmed by pagination (messages leave the window per #11/#331), the corresponding `.loaded(MessageMediaDownload)` entries — each holding the full decrypted `payload.data` — are not released.

## Evidence
- `whitenoise-mac/Core/WorkspaceState+Media.swift:490-503` — prune only removes keys **not** prefixed by the active group:
  ```swift
  func pruneMediaDownloadCache(keeping groupIdHex: String?) {
      guard let activeAccountId, let groupIdHex else { resetMediaDownloadStateStores(); return }
      let prefix = [activeAccountId, groupIdHex, ""].joined(separator: "\u{1F}")
      let removedKeys = mediaDownloads.keys.filter { !$0.hasPrefix(prefix) }   // OTHER groups only
      for key in removedKeys { mediaDownloads[key]?.update(.idle); mediaDownloads[key] = nil }
  }
  ```
  Its only caller is `pruneMessageCache(keeping:)` (`WorkspaceState+Timeline.swift`), invoked on chat switch — so within a conversation nothing is pruned.
- Each `.loaded` store holds the decrypted payload (`whitenoise-mac/Core/WorkspaceState+Media.swift` store path, ~`:222-231`): `DownloadedMediaPayload(id:, data: plaintext)`.

## Failure scenario
A user scrolls a media-heavy conversation. Each visual tile auto-downloads on visibility and decrypts full plaintext into `mediaDownloads`. As they keep scrolling and pagination trims old messages from the window, the decrypted-plaintext stores for scrolled-past messages remain resident. Memory grows monotonically for the session-in-conversation, and decrypted plaintext for messages no longer visible stays in RAM until a chat switch or app quit.

- **#144** is `mediaDownloadState` over-reading the dict (re-render), not retention.
- **#328** is the *disk*-cache unbounded growth (the encrypted on-disk cache), not in-memory plaintext.
- **#132/#142** are full-res `NSImage` decode in body (display decode), separate from the retained source `Data`.
- **#141/#177/#288** clear on account removal / sign-out. None addresses the lack of *within-conversation* eviction as messages leave the timeline window.

## Suggested fix
Prune `mediaDownloads` entries whose message id is no longer in the active timeline window (hook the same window-trim path that drops messages), and/or bound the store with an LRU byte cap that evicts back to `.idle` — the encrypted disk cache already backs re-hydration, so eviction is cheap to recover from.

## Scope note

agent-p1p:
## Summary
- prune `mediaDownloads` to the attachment keys still present in the active timeline window
- run that pruning after full timeline window replacement and incremental projection mutations
- add regression coverage for stale plaintext eviction and update existing media download tests to seed their synthetic messages into the active window

## Test plan
- `git diff --check`
- conflict marker scan with `search_files` over Swift sources
- GitHub Mac CI: `Static Analysis` pass, `PR Checks` pass
- not run locally: `xcodebuild` / Swift tests (xcodebuild, swift, and swiftformat are not installed on this Linux host)

Fixes #394

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Media.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Media.swift
@@ -246,7 +246,7 @@ extension WorkspaceState {
 
     /// Lazily allocates per-attachment stores from SwiftUI body lookup without observing the
     /// backing dictionary; `mediaDownloads` is `@ObservationIgnored`, and pruning bounds it to
-    /// the active conversation.
+    /// attachments that still belong to the active timeline window.
     func mediaDownloadStateStore(forKey key: String) -> MediaDownloadStateStore {
         if let store = mediaDownloads[key] {
             return store
@@ -484,7 +484,19 @@ extension WorkspaceState {
     }
 
     func mediaDownloadKey(message: MessageItem, attachment: MessageMediaAttachment) -> String {
-        [activeAccountId ?? "", message.groupIdHex, attachment.id].joined(separator: "\u{1F}")
+        mediaDownloadKey(
+            accountId: activeAccountId ?? "",
+            groupIdHex: message.groupIdHex,
+            attachmentId: attachment.id
+        )
+    }
+
+    private func mediaDownloadKey(
+        accountId: String,
+        groupIdHex: String,
+        attachmentId: String
+    ) -> String {
+        [accountId, groupIdHex, attachmentId].joined(separator: "\u{1F}")
     }
 
     func pruneMediaDownloadCache(keeping groupIdHex: String?) {
@@ -494,12 +506,31 @@ extension WorkspaceState {
         }
 
         let prefix = [activeAccountId, groupIdHex, ""].joined(separator: "\u{1F}")
-        let removedKeys = mediaDownloads.keys.filter { !$0.hasPrefix(prefix) }
+        let retainedKeys = retainedMediaDownloadKeys(groupIdHex: groupIdHex, accountId: activeAccountId)
+        let removedKeys = mediaDownloads.keys.filter { key in
+            guard key.hasPrefix(prefix) else { return true }
+            return !retainedKeys.contains(key)
+        }
         for key in removedKeys {
             // Notify any lingering per-attachment observers before dropping the store.
             mediaDownloads[key]?.update(.idle)
             mediaDownloads[key] = nil
         }
+    }
+
+    private func retainedMediaDownloadKeys(groupIdHex: String, accountId: String) -> Set<String> {
+        guard let timelineStore = messageTimelineStores[groupIdHex] else { return [] }
+        return Set(
+            timelineStore.messages.flatMap { message in
+                message.mediaAttachments.map { attachment in
+                    mediaDownloadKey(
+                        accountId: accountId,
+                        groupIdHex: message.groupIdHex,
+                        attachmentId: attachment.id
+                    )
+                }
+            }
+        )
     }
 
     func resetMediaDownloadStateStores() {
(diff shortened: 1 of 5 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Track message removal in timeline projections](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416/chapters/1) |
| 2 | [Implement media download pruning logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416/chapters/2) |
| 3 | [Guard media publication against stale state](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416/chapters/3) |
| 4 | [Trigger pruning during timeline mutations](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416/chapters/4) |
| 5 | [Add regression tests for media pruning](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/416">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 1bd164c7469197daa7538d9f4aafc175a0d41dc1 on Jul 8, 2026 10:00am UTC.</sub>

agent-p1p:
APPROVE-equivalent review (GitHub does not allow agent-p1p to formally approve its own PR).

Pip adversarial review for #416 / issue #394.

BLOCKING:
- none

SUGGESTION:
- Prune only when a projection can actually evict media: `whitenoise-mac/Core/WorkspaceState+Timeline.swift:424` / `:759`. `finalizeTimelineStoreMutation` is reached for any `didChange == true` projection, including edits/reactions/per-relay delivery-state acks. The new unconditional prune then rebuilds retained media keys by scanning the full active window and joining strings for every attachment. The window is bounded, so this is not blocking, but eviction is only possible for removals or `result.didTrimOlderMessages`; passing that signal into `finalizeTimelineStoreMutation` would keep the hot delivery path cheaper.
- Re-check before publishing an async download into a pruned store: `whitenoise-mac/Core/WorkspaceState+Media.swift:213-231`. If a tile starts downloading and the message leaves the window before the await completes, pruning removes the store from `mediaDownloads` and sets it `.idle`, but the in-flight load still updates the captured store object to `.loaded` after decryption. It does not reinsert into `mediaDownloads`, so the persistent leak is fixed, but any view still holding that store can transiently retain/render plaintext for a message that has already left the active window. Consider checking `mediaDownloads[key] === stateStore` and/or membership in the retained window before the final `.loaded` update.

NITPICK:
- `retainedMediaDownloadKeys` uses `message.groupIdHex` while constructing keys (`whitenoise-mac/Core/WorkspaceState+Media.swift:528`) even though it already received `groupIdHex`. Equivalent with today's store invariant; using the parameter would make the invariant explicit.

SUMMARY: The fix covers the reported bug. Full window replacement and incremental projection mutations now prune `mediaDownloads` down to attachments still present in the active timeline window, while chat-switch/deselect/account reset eviction paths remain intact. Disk cache rehydration covers scroll-back, and the new regression test exercises real stale-plaintext eviction. The remaining points are bounded hot-path overhead and a transient in-flight-download edge case, not blockers.

VERDICT: APPROVE

SENSITIVE_PATHS: `whitenoise-mac/Core/WorkspaceState+Media.swift` and `whitenoise-mac/Core/WorkspaceState+Timeline.swift` touch privacy-sensitive decrypted attachment plaintext lifetime; no crypto/key/auth/storage format code touched.

PR #438, opened 2026-07-09
fix: invalidate selected chat on metadata deltas

### LINKED ISSUE #388: Bug: selectedChat resolves through @ObservationIgnored chatLookupByAccount, so the open conversation's header/composer never re-render on live chat-metadata deltas (rename, membership-ended, invite flip)
## Severity
**MEDIUM** (correctness / stale UI). A live group rename, avatar/enrichment update, invite-confirmation flip, or *membership-ended* flip mutates the selected chat but leaves the entire open-conversation UI showing a stale snapshot until the user navigates away and back. The sidebar row updates; the conversation pane does not. It does **not** allow an actual send to a group you were removed from — the send path re-reads fresh state (see "Bounded impact") — so this is UI staleness, not a data-integrity defect.

## Summary
`selectedChat` resolves the chat **only** through `chatLookupByAccount`, which is `@ObservationIgnored`. Under Swift Observation, a computed property's tracked dependency set is the observed stored properties it reads; `selectedChat` reads only `selection` and `activeAccountId` as observed inputs (the lookup dict is ignored). `ConversationView` receives the chat **by value** from `DetailPaneView`, whose only observed dependencies are `{phase, selection, activeAccountId}`. So a chat-metadata delta that mutates `chatsByAccount`/`chatLookupByAccount` — but not `selection`/`activeAccountId` — cannot invalidate the conversation pane.

## Evidence
- `whitenoise-mac/Core/WorkspaceState.swift:460-461` — `var chatsByAccount` is observed; `@ObservationIgnored var chatLookupByAccount: [String: [String: ChatItem]] = [:]` is **not**.
- `whitenoise-mac/Core/WorkspaceState.swift:1187-1190` — `selectedChat` resolves purely via the ignored index:
  ```swift
  var selectedChat: ChatItem? {
      guard case .chat(let chatId) = selection else { return nil }
      return activeAccountId.flatMap { chatLookupByAccount[$0]?[chatId] }
  }
  ```
  Observed dependency set = `{selection, activeAccountId}`; it never touches observed `chatsByAccount`.
- `whitenoise-mac/Views/MessengerShellView.swift:196-203` — `DetailPaneView.body` reads only `phase`, `selection`, `selectedChat`, then passes the value by copy: `if let chat = workspace.selectedChat { ConversationView(chat: chat) }`. `DetailPaneView` is a separate subtree from `ChatListDrawerView`, so it is not co-invalidated with the sidebar.
- `whitenoise-mac/Views/MessengerShellView.swift:295` — `ConversationView` stores the frozen value `let chat: ChatItem`. Its `body` (`:298-303`) reads `selectedMessages`/`selectedMessageIDs`/`selectedTimelinePaging`/`selectedTimelineIsLoadingInitialPage` (the per-chat timeline stores) — **none** of which change on a metadata-only delta — and renders `ConversationHeader(chat: chat)` (`:306`), the `chat.isNoLongerMember` composer notice (`:456`), and the file-drop gate (`:498, 502`) from the frozen value.
- `whitenoise-mac/Views/ComposerViews.swift:805, 812, 856` — `ConversationHeader` renders `chat.title`, `chat.avatarSeed`, and the `chat.pendingConfirmation`/group affordances from the passed value, not a fresh read.
- Mutation path that goes unobserved: `WorkspaceState+ChatList.swift` (chat-list subscription → `applyChatRow` → `upsertChat`) writes `chatsByAccount` **and** `chatLookupByAccount` and bumps `chatListGenerationByAccount` (`WorkspaceState.swift:1205-1222`), but changes neither `selection` nor `activeAccountId`. `ChatItem` is a value type (`Models/MessengerModels.swift:105`), so the copy in `ConversationView.chat` is frozen at the last `DetailPaneView` evaluation.

## Concrete failure scenario
User is viewing group G. From another device the group is renamed (or the local account is removed → `selfMembership != .member`). The chat-list listener applies the row via `upsertChat`. `ChatListDrawerView` re-renders and shows the new title / removed state, but `DetailPaneView.body` has no changed observed dependency, so `ConversationView` keeps the stale `chat`: the header still shows the old name, and after removal the composer never swaps to `MembershipEndedComposerNotice` (`MessengerShellView.swift:456`) — it keeps showing an active composer. It corrects only when the user selects another chat and returns (which re-runs `DetailPaneView`). A pure rename produces **no** timeline delta at all, so nothing coincidentally refreshes the pane.

## Bounded impact (why this is not a send-integrity bug)
The actual send action re-reads `selectedChat` fresh at call time and rejects removed-group sends: `WorkspaceState+Timeline.swift:445-448` (`guard let selectedChat, … !selectedChat.isNoLongerMember`), and `canSend` (`WorkspaceState.swift:1392-1397`) likewise. Because `chatLookupByAccount` holds the fresh value, these guards read the up-to-date `isNoLongerMember` even while the view is stale — so no message is actually delivered to a group the user left. The defect is that the UI misrepresents state (wrong title/avatar, a still-active composer + enabled file-drop after removal, stale accept-invite button).

- **#176** is the *opposite* problem — `ConversationView` *over-subscribing* to `messagesByChat` (too many re-renders), since resolved by the per-chat `messageTimelineStores`. This finding is *under-subscription* on the different `chatLookupByAccount` index (too few re-renders). Different property, opposite direction.
- **#144** is the same over-subscription pattern on `mediaDownloads`. No filed issue covers `selectedChat` reading ignored storage and thus failing to invalidate the conversation pane on metadata deltas.

## Suggested fix
Make `selectedChat` read observed storage so the pane invalidates on chat updates. Options:
1. Resolve from the observed `chatsByAccount[activeAccountId]` (keep `chatLookupByAccount` purely as the internal O(1) index for non-UI callers) — simplest, but reintroduces whole-dict observation cost on the selected-chat read.
2. Add a small observed `selectedChatRevision` bumped in `bumpChatListGeneration`, and have `selectedChat` (or `DetailPaneView`) read it, so a rename/membership delta invalidates the pane without whole-dict subscription.
3. Have `ConversationView` re-read the chat by id through an observed accessor each body pass rather than holding a by-value snapshot.

## Scope note
Found during a full-codebase review. The `@ObservationIgnored` index was presumably introduced for O(1) lookup / to avoid the #176-class over-render; the tradeoff is that the primary UI reader of it under-subscribes.

agent-p1p:
## Summary
- add an observed selected-chat revision token while keeping selectedChat on the ignored O(1) lookup index
- bump that token when the active account chat list changes so the open conversation re-reads fresh rename/avatar/invite/membership-ended metadata
- add a Swift Observation regression test for a selected-chat metadata delta

Fixes #388

## Test plan
- `git diff --check`
- `git grep -n '^<<<<<<<\\|^=======\\|^>>>>>>>' -- . ':!Vendored/MarmotKit/MarmotKit.xcframework'`
- Not run on this Linux host: `xcrun swift-format ...` (`xcrun` missing), `xcodebuild test ...` (`xcodebuild` missing)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/438">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -468,7 +468,9 @@ final class WorkspaceState {
     /// Observed invalidation token for `selectedChat`, whose actual lookup stays in the
     /// ignored O(1) indexes below. Live chat-list deltas mutate those indexes, so the selected
     /// conversation needs this tracked scalar to re-read fresh metadata without subscribing to
-    /// the whole chat dictionary.
+    /// the whole chat dictionary. It is intentionally coarse: any active-account chat-list
+    /// mutation invalidates the selected conversation chrome, while transcript reads remain
+    /// scoped through per-chat timeline stores.
     var selectedChatRevision = 0
     @ObservationIgnored var chatLookupByAccount: [String: [String: ChatItem]] = [:]
     @ObservationIgnored var chatIndexByAccount: [String: [String: Int]] = [:]
@@ -1283,6 +1285,9 @@ final class WorkspaceState {
     }
 
     func resetChats() {
+        // Currently used only by `resetToNewInstallState`, which clears `activeAccountId` and
+        // `selection` in the same synchronous reset path. A future caller that keeps a selected
+        // chat alive while clearing these ignored indexes must also bump `selectedChatRevision`.
         chatsByAccount = [:]
         chatLookupByAccount = [:]
         chatIndexByAccount = [:]
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -465,6 +465,11 @@ final class WorkspaceState {
     var phase: Phase = .bootstrapping
     var accounts: [AccountItem]
     var chatsByAccount: [String: [ChatItem]]
+    /// Observed invalidation token for `selectedChat`, whose actual lookup stays in the
+    /// ignored O(1) indexes below. Live chat-list deltas mutate those indexes, so the selected
+    /// conversation needs this tracked scalar to re-read fresh metadata without subscribing to
+    /// the whole chat dictionary.
+    var selectedChatRevision = 0
     @ObservationIgnored var chatLookupByAccount: [String: [String: ChatItem]] = [:]
     @ObservationIgnored var chatIndexByAccount: [String: [String: Int]] = [:]
     @ObservationIgnored var chatListGenerationByAccount: [String: Int] = [:]
@@ -1208,6 +1213,7 @@ final class WorkspaceState {
 
     var selectedChat: ChatItem? {
         guard case .chat(let chatId) = selection else { return nil }
+        _ = selectedChatRevision
         return activeAccountId.flatMap { chatLookupByAccount[$0]?[chatId] }
     }
 
@@ -1333,6 +1339,9 @@ final class WorkspaceState {
 
     private func bumpChatListGeneration(forAccountId accountId: String) {
         chatListGenerationByAccount[accountId, default: 0] += 1
+        if activeAccountId == accountId {
+            selectedChatRevision += 1
+        }
         if filteredChatsCache?.accountId == accountId {
             filteredChatsCache = nil
         }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -9086,6 +9086,75 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func selectedChatObservationInvalidatesOnSelectedChatMetadataDelta() async throws {
+        // Regression for #388: `selectedChat` resolves through ignored O(1) indexes, so it must
+        // still read an observed token that live chat-list deltas bump. Otherwise the open
+        // conversation keeps rendering a frozen ChatItem after renames/membership-ended flips.
+        let account = AccountItem.samples[0]
+        let selected = ChatItem(
+            id: "selected-chat",
+            title: "Original group",
+            subtitle: "Group message",
+            preview: "before",
+            updatedAt: Date(timeIntervalSince1970: 100),
+            avatarSeed: "selected-chat",
+            pictureURL: nil,
+            unreadCount: 0
+        )
+        let background = ChatItem(
+            id: "background-chat",
+            title: "Background group",
+            subtitle: "Group message",
+            preview: "background",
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce invalidation token for selected chat](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/438/chapters/1) |
| 2 | [Trigger invalidation on chat list updates](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/438/chapters/2) |
| 3 | [Verify selected chat invalidation with regression test](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/438/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/438">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 7964f307b30e5754d1329084eccefbf1b5ee9f82 on Jul 9, 2026 10:26am UTC.</sub>

agent-p1p (commented):
## Pip adversarial review (cursor-reviewer)

BLOCKING:
- none

SUGGESTION:
- Coarse selected-chat invalidation granularity: `whitenoise-mac/Core/WorkspaceState.swift:1340-1344`. `bumpChatListGeneration` now increments `selectedChatRevision` for any active-account chat-list mutation, not only mutations to the currently selected chat. That fixes #388, but background-chat preview/unread/read-state deltas can also invalidate the open conversation pane's header/composer/file-drop chrome. The timeline remains scoped to per-chat stores, so this is not a blocker. Consider either documenting the coarse granularity in the new token comment or tightening a follow-up to bump only when the selected chat row changed.

NITPICK:
- `resetChats()` is the only `chatLookupByAccount` mutator that does not bump `selectedChatRevision` (`WorkspaceState.swift:1285-1291`). Current callers clear `selection`/`activeAccountId`, so this is harmless. A one-line invariant comment would make that dependency explicit for future callers.

SUMMARY: The fix is correct for #388. `selectedChat` now reads an observed scalar while keeping the ignored O(1) lookup, and the live chat-list mutation paths that can change selected-chat metadata route through `bumpChatListGeneration`. The regression test is meaningful: without the new observed read, `withObservationTracking` would only depend on `selection`/`activeAccountId` and would not fire on the `upsertChat` metadata delta.

VERDICT: APPROVE

SENSITIVE_PATHS: none

Local checks run:
- `git diff --check master...HEAD` passed
- conflict-marker scan passed
- `xcrun` / `xcodebuild` unavailable on this Linux host, so Swift formatting and Xcode tests were not run locally

PR #439, opened 2026-07-09
fix: stop recording on last account sign-out

### LINKED ISSUE #386: Bug/Privacy: signing out the last active account leaves the microphone hot and plaintext voice recording on disk (bypasses leaveActiveConversation)
## Severity
**HIGH** (privacy: the microphone keeps capturing with **no** UI affordance to stop it after the app has dropped to onboarding, and the decrypted/plaintext `.m4a` recording persists on disk. Strictly worse than the already-filed MEDIUM mic-hot issues because the composer that owns the Stop control is gone.)

## Summary
`leaveActiveConversation()` is the centralized teardown every path that removes the conversation composer from the hierarchy must run so "the microphone can never stay hot with no visible way to stop it (#311)" and an in-flight transcript export cannot keep paginating (#316). `signOutAccount(_:)` — reachable directly from the always-visible account-rail context menu, **not** just Settings — never calls it, and neither does the shared `resetActiveAccountUIState()`.

When the signed-out account is the **last** signed-in account, sign-out takes the `else` (onboarding) branch, which performs no voice-recording teardown at all. The mic stays live, `AVAudioRecorder` keeps writing plaintext audio, and the recording file is never deleted.

## Evidence
- `whitenoise-mac/Core/WorkspaceState+Account.swift:343-382` — `signOutAccount(_:)`. Active-account teardown is only `stopTimelineListener()/cancelChatListReload()/stopChatListListener()` (`:351-355`) then `resetActiveAccountUIState()` (`:365`). **No** `leaveActiveConversation()` / `cancelVoiceRecording()`.
- `whitenoise-mac/Core/WorkspaceState+Account.swift:306-338` — `resetActiveAccountUIState()` (the shared removal/sign-out teardown) clears timelines, caches, drafts, avatars, read markers — but never touches the voice recorder.
- Multi-account path is safe by accident: `:366-367` `switchActiveAccount(nextActive, …)` → `prepareForActiveAccountSwitch` (`WorkspaceState+Account.swift:85`) → `leaveActiveConversation()`. The **`else` branch `:370-378`** ("no next account → `phase = .onboarding`") has no such call, so the leak is **guaranteed** when the last account is signed out.
- `whitenoise-mac/Views/SidebarViews.swift:108-124` — "Sign Out" lives on `AccountRailAvatar`'s `contextMenu`, i.e. the persistent account rail shown *alongside* an open conversation, so it is reachable without the `showSettings` → `leaveActiveConversation()` teardown (`WorkspaceState+Navigation.swift:54-61`).
- `leaveActiveConversation()` (`WorkspaceState+Navigation.swift:26-29`) → `cancelVoiceRecording()` → `resetVoiceRecording(deleteFile: true)` (`WorkspaceState+Media.swift`) is what stops `AVAudioRecorder` and deletes the plaintext scratch file. `signOutAccount` keeps `voiceRecorder` running and leaves `voiceRecordingURL` on disk (sign-out never purges `MediaPlaybackTempStore`).

## Failure scenario
1. User has a **single** signed-in account, opens a chat, taps record — mic active, plaintext audio streaming to a protected `.m4a` scratch file.
2. Without stopping, they right-click their avatar in the account rail and choose **Sign Out**.
3. `signOutAccount` runs; no other signed-in account exists → onboarding branch (`:370-378`).
4. The app shows onboarding while `AVAudioRecorder` **keeps capturing the microphone** (`isRecordingVoiceMessage`/`voiceRecorder` stay set) and the plaintext recording remains in `MediaPlaybackTempStore` — with no composer and therefore no Stop button anywhere in the UI.

- **#311** covers only navigating to Settings / opening new-chat.
- **#374** covers the notification-tap same-account path.
- **#329** is the *stale `wasActive` snapshot* race in the same function — orthogonal to the missing mic teardown.
None of them touch `signOutAccount` / `resetActiveAccountUIState` failing to run `leaveActiveConversation()`, nor the last-account→onboarding path where the leak is guaranteed and no Stop affordance survives.

## Suggested fix
Call `leaveActiveConversation()` at the top of `resetActiveAccountUIState()` (covers both `signOutAccount` and `removeAccount`), mirroring `resetToNewInstallState()` and `deleteAllData()`, which both call it first. That closes the last-account onboarding path and is idempotent on the multi-account path where `prepareForActiveAccountSwitch` already runs it.

## Scope note
Found during a full-codebase review. `removeAccount` shares the same `resetActiveAccountUIState()` omission, but its only callers are inside Settings (`SettingsViews.swift:294,763`), and reaching Settings runs `showSettings` → `leaveActiveConversation()` first, so the mic is already down on that path. Fixing `resetActiveAccountUIState()` covers both regardless.

agent-p1p:
## Summary
- Route active-account UI reset through `leaveActiveConversation()` so sign-out/removal teardown cancels recorder-owned conversation resources before the composer disappears.
- Add regression coverage for signing out the only active account while a voice recording is in progress; onboarding now clears recorder state and deletes the plaintext `.m4a` scratch file.

## Test Plan
- `git diff --check`
- `git grep -nE '^<<<<<<<|^=======|^>>>>>>>' -- .` (no conflict markers)
- `xcodebuild -scheme whitenoise-mac -configuration Debug build-for-testing CODE_SIGNING_ALLOWED=NO` not run: `xcodebuild` is unavailable on this Linux worker
- `swift-format` / `swift` not run: unavailable on this Linux worker

Fixes #386

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/439">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Account.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Account.swift
@@ -310,9 +310,10 @@ extension WorkspaceState {
         }
     }
 
-    /// Clears all active-account-scoped in-memory UI state (timelines, caches,
-    /// drafts, settings snapshots). Shared by account removal and sign-out.
+    /// Clears all active-account-scoped in-memory UI state (conversation resources,
+    /// timelines, caches, drafts, settings snapshots). Shared by account removal and sign-out.
     func resetActiveAccountUIState() {
+        leaveActiveConversation()
         stopTimelineListener()
         cancelTimelineLoad()
         cancelChatListReload()
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -776,6 +776,38 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func signOutLastActiveAccountStopsInProgressVoiceRecording() async throws {
+        // #386: the account-rail sign-out path can send the last signed-in account
+        // straight to onboarding, removing the composer that owns the Stop control.
+        // It must still run active-conversation teardown so the mic is not left hot
+        // and the plaintext scratch recording does not survive sign-out.
+        let previousActiveAccount = UserDefaults.standard.object(forKey: WorkspaceState.activeAccountKey)
+        defer { restoreDefault(previousActiveAccount, forKey: WorkspaceState.activeAccountKey) }
+
+        let primary = desktopAccount()
+        let runtime = FakeMarmotRuntime(accounts: [primary])
+        UserDefaults.standard.set(primary.label, forKey: WorkspaceState.activeAccountKey)
+        let state = WorkspaceState(clientFactory: { runtime })
+
+        await state.bootstrap()
+        let url = try armInProgressVoiceRecording(on: state)
+        let desktopAccount = try #require(state.accounts.first { $0.id == primary.label })
+
+        await state.signOutAccount(desktopAccount)
+
+        #expect(runtime.signedOutAccountRefs == [primary.label])
+        #expect(state.phase == .onboarding)
+        #expect(state.activeAccountId == nil)
+        #expect(!state.isRecordingVoiceMessage)
+        #expect(state.voiceRecorder == nil)
+        #expect(state.voiceRecordingURL == nil)
+        #expect(state.voiceRecordingMeterTask == nil)
+        #expect(state.voiceRecordingSamples.isEmpty)
+        #expect(state.voiceRecordingDurationSeconds == 0)
+        #expect(!FileManager.default.fileExists(atPath: url.path))
+    }
+
+    @MainActor
     @Test func removingBackgroundAccountSelectedMidFlightRecoversActiveAccount() async throws {
         // Regression for the account-switch/remove race: if the user selects the
         // background account that is currently being removed while removal is in flight,


ghost:

| | Title |
|---|-------|
| 1 | [Teardown conversation resources on account sign-out](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/439/chapters/1) |
| 2 | [Verify voice recording cleanup during sign-out](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/439/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/439">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit f00f00bb7ed020a7e9e68f6782338f088d06a055 on Jul 9, 2026 10:36am UTC.</sub>

agent-p1p (commented):
Pip adversarial review for #386 / PR #439.

Verdict: APPROVE-equivalent comment. No blocking findings.

BLOCKING:
- none

SUGGESTION:
- Test does not prove a live `AVAudioRecorder` was stopped: `whitenoise-macTests/whitenoise_macTests.swift:793,802` — `armInProgressVoiceRecording` never assigns a real non-nil `state.voiceRecorder`, so the `voiceRecorder == nil` assertion is true before and after sign-out. The test still meaningfully proves the important state reset, meter-task cancellation, onboarding transition, and plaintext `.m4a` deletion, and this matches the existing sibling teardown tests. If true recorder-stop coverage becomes important, it likely needs a lightweight recorder seam or an explicit comment acknowledging the hardware-permission limitation.
- Redundant listener teardown remains in `signOutAccount`: `whitenoise-mac/Core/WorkspaceState+Account.swift:377-382` — the early `if wasActive` block stops/cancels the same listeners that `resetActiveAccountUIState()` now stops unconditionally at `:401`. Non-blocking because it predates this PR and is harmless, but deleting the duplicate block later would reduce the two-lists-that-must-stay-in-sync hazard.

NITPICK:
- none

SUMMARY:
The fix is correct and minimal. `resetActiveAccountUIState()` now calls `leaveActiveConversation()` before the sign-out branch split, so the last-account -> onboarding path runs the same teardown that stops voice recording and deletes the in-flight plaintext scratch file. The multi-account path may call `leaveActiveConversation()` again via `prepareForActiveAccountSwitch`, but the teardown is idempotent. No scope creep or privacy regression found.

Validation:
- `git diff --check origin/master...HEAD` passed locally.
- GitHub checks are green: PR Checks, Static Analysis, CodeRabbit.

SENSITIVE_PATHS:
- `whitenoise-mac/Core/WorkspaceState+Account.swift` — account sign-out/removal teardown; privacy-sensitive microphone/plaintext-audio lifecycle.

PR #442, opened 2026-07-09
fix: keep filtered chats observed on cache hits

### LINKED ISSUE #390: Bug: filteredChats memoization cache short-circuits before reading observed chatsByAccount, so the sidebar stops updating on incoming messages after any cache-hit render
## Severity
**MEDIUM** (correctness / stale UI). After the single most common interaction — clicking a chat — the always-visible chat list stops reflecting new messages in *other* chats: no reorder-to-top, no preview update, no unread badge, until an unrelated observed change forces a re-render. Self-heals on the next chat selection or search keystroke, and OS-level notifications still fire, so the user is not fully blind — hence MEDIUM, not HIGH.

## Summary
`filteredChats` is memoized in an `@ObservationIgnored` cache. On a cache **hit** it returns `cache.result` *without ever reading the observed `chatsByAccount`*. Under Swift Observation, a view's tracked dependency set is exactly the observed stored properties it reads during a body pass. So a cache-hit render of the sidebar drops `chatsByAccount` from its dependency set, and the next `chatsByAccount` mutation (a new message via `upsertChat`) does not invalidate the sidebar.

## Evidence
- `whitenoise-mac/Core/WorkspaceState.swift:460` — `var chatsByAccount` is observed.
- `whitenoise-mac/Core/WorkspaceState.swift:463-464` — the cache key inputs are **not** observed:
  ```swift
  @ObservationIgnored var chatListGenerationByAccount: [String: Int] = [:]
  @ObservationIgnored var filteredChatsCache: FilteredChatsCache?
  ```
- `whitenoise-mac/Core/WorkspaceState.swift:1175-1195` — the getter reads only `activeAccountId` + `searchText` as observed inputs on the hit path; `chatsByAccount` is read **only in the miss branch**:
  ```swift
  var filteredChats: [ChatItem] {
      guard let activeAccountId else { return [] }
      let generation = chatListGenerationByAccount[activeAccountId] ?? 0   // ObservationIgnored
      let query = searchText
      if let cache = filteredChatsCache, cache.accountId == activeAccountId,
          cache.generation == generation, cache.query == query {
          return cache.result                                              // returns WITHOUT reading chatsByAccount
      }
      let result = ChatFilter.filtered(chatsByAccount[activeAccountId] ?? [], query: query)  // only on miss
      ...
  }
  ```
- `whitenoise-mac/Views/SidebarViews.swift:163` — the sidebar body reads `let filteredChats = workspace.filteredChats` and (`:193-204`) iterates it; the only other observed reads in that subtree are `selection` and `searchText`. Nothing else in the drawer body reads `chatsByAccount`.
- `whitenoise-mac/Core/WorkspaceState.swift:1215-1232` — `upsertChat` mutates observed `chatsByAccount` (`:1222`) and bumps the **ignored** `chatListGenerationByAccount` via `bumpChatListGeneration` (`:1231`). The generation bump is what will later invalidate the cache, but it emits no observation change.

## Failure scenario
1. Sidebar first renders → `filteredChats` misses → reads observed `chatsByAccount` → view is subscribed to it; cache is populated.
2. User clicks a chat. `selection` changes (observed) → sidebar body re-evaluates. This time `filteredChats` **hits** (same account/generation/query) and returns the cached array **without touching `chatsByAccount`** → SwiftUI drops `chatsByAccount` from the view's dependency set.
3. A message arrives in a *background* chat → `applyChatRow` → `upsertChat` mutates `chatsByAccount` and bumps the ignored generation.
4. The sidebar is no longer subscribed to `chatsByAccount`, so it is **not** invalidated: the background chat does not move to the top, its preview/unread badge do not update. It stays stale until any observed change the view *is* still subscribed to (another selection, a search keystroke) forces a re-render — which then misses (generation bumped) and refreshes.

Because "read one chat while messages land in others" is a core messenger scenario and step 2 is exactly what produces the cache-hit render, this is a routine, not exotic, staleness.

- **#388** is the analogous under-subscription but for a *different* property (`selectedChat` via `chatLookupByAccount`) affecting the *conversation pane* header/composer. This finding is a *different* property (`filteredChats`), a *different* backing store (`filteredChatsCache`/`chatListGenerationByAccount`), a *different* view (the sidebar chat **list**), and a *different* mechanism (a memoization cache that returns *before* reading the observed store, rather than a direct ignored-lookup).
- **#176** is the opposite direction (ConversationView *over*-subscribing to `messagesByChat`). **#129** is the per-keystroke `ChatFilter.filtered` allocation. **#170/#251** are enrichment cost. None address the chat-list under-subscription.

## Suggested fix
Force the observed dependency to register on every read, independent of cache outcome. Simplest: read `chatsByAccount[activeAccountId]` (or a cheap `_ = chatsByAccount`) *before* the cache short-circuit, and only use it in the miss branch. Alternatively, make `chatListGenerationByAccount` an observed property so the generation read itself establishes the dependency (then the cache-hit path is still invalidated on every mutation).

## Scope note
Found during a full-codebase review. The `@ObservationIgnored` cache was presumably introduced to avoid re-running `ChatFilter.filtered` per body pass (cf. #129); the tradeoff is that the memoization defeats Observation's read-tracking on the hit path.

agent-p1p:
## Summary
- keep `filteredChats` subscribed to the observed active-account chat list before cache-hit returns
- add a Swift Observation regression test for cache-hit filtered chat reads followed by background chat updates

## Test Plan
- `git diff --check`
- Not run: `xcodebuild` / Swift tests (Linux worker has no `xcodebuild` or `swift` toolchain)

Fixes #390

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/442">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState.swift
+++ b/whitenoise-mac/Core/WorkspaceState.swift
@@ -1204,6 +1204,9 @@ final class WorkspaceState {
 
     var filteredChats: [ChatItem] {
         guard let activeAccountId else { return [] }
+        // Read the observed list before a cache-hit return so SwiftUI keeps invalidating the
+        // sidebar on incoming chat updates even when filtering work is memoized.
+        let chats = chatsByAccount[activeAccountId] ?? []
         let generation = chatListGenerationByAccount[activeAccountId] ?? 0
         let query = searchText
         if let cache = filteredChatsCache,
@@ -1214,7 +1217,7 @@ final class WorkspaceState {
             return cache.result
         }
 
-        let result = ChatFilter.filtered(chatsByAccount[activeAccountId] ?? [], query: query)
+        let result = ChatFilter.filtered(chats, query: query)
         filteredChatsCache = FilteredChatsCache(
             accountId: activeAccountId,
             generation: generation,
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -9384,6 +9384,69 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func filteredChatsObservationInvalidatesOnChatListDeltaAfterCacheHit() async throws {
+        // Regression for #390: `filteredChats` is memoized through ignored cache state, but each
+        // body pass still has to read the observed chat list. Otherwise a cache-hit pass drops the
+        // sidebar's `chatsByAccount` subscription and background-chat updates do not repaint it.
+        let account = AccountItem.samples[0]
+        let selected = ChatItem(
+            id: "selected-chat",
+            title: "Selected group",
+            subtitle: "Group message",
+            preview: "selected preview",
+            updatedAt: Date(timeIntervalSince1970: 100),
+            avatarSeed: "selected-chat",
+            pictureURL: nil,
+            unreadCount: 0
+        )
+        let background = ChatItem(
+            id: "background-chat",
+            title: "Background group",
+            subtitle: "Group message",
+            preview: "old background",
+            updatedAt: Date(timeIntervalSince1970: 90),
+            avatarSeed: "background-chat",
+            pictureURL: nil,
+            unreadCount: 0
+        )
+        let state = WorkspaceState(
+            accounts: [account],
+            chatsByAccount: [account.id: [selected, background]],
+            clientFactory: { FakeMarmotRuntime(accounts: []) }
+        )
+        state.activeAccountId = account.id
+        state.selection = .chat(selected.id)
+
+        #expect(state.filteredChats.map(\.id) == [selected.id, background.id])
+
+        let invalidated = ObservationInvalidationFlag()
+        withObservationTracking {
+            _ = state.filteredChats.map(\.id)
+        } onChange: {
+            invalidated.markInvalidated()
+        }
+
+        state.upsertChat(
+            ChatItem(
+                id: background.id,
+                title: background.title,
+                subtitle: background.subtitle,
+                preview: "new background",
+                updatedAt: Date(timeIntervalSince1970: 200),
+                avatarSeed: background.avatarSeed,
+                pictureURL: background.pictureURL,
+                unreadCount: 3
+            ),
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Ensure filtered chats remain observed on cache hits](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/442/chapters/1) |
| 2 | [Add regression test for chat list observation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/442/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/442">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit b2843bf34450cf6b132fb451b03c9173843e7d97 on Jul 9, 2026 2:26pm UTC.</sub>

agent-p1p:
@coderabbitai review

agent-p1p (commented):
## Pip adversarial review

BLOCKING:
- none

SUGGESTION:
- none

NITPICK:
- `whitenoise-mac/Core/WorkspaceState.swift:1364-1366` already clears `filteredChatsCache` for the changed account when bumping the generation, so the active-account cache hit only survives within an unchanged generation. Pre-existing and not worth changing in this PR.
- `whitenoise-mac/Core/WorkspaceState.swift:1207-1209` could mention `#390` inline for future archeologists. The new regression test already cites it, so this is optional.

SUMMARY: The fix is minimal and targets the actual Observation invariant: `filteredChats` now reads observed `chatsByAccount[activeAccountId]` before any cache-hit return, keeping SwiftUI subscribed to chat-list mutations while preserving the memoized filter result. The regression test is meaningful: it primes the cache, performs a cache-hit tracked read, mutates a background chat through `upsertChat`, then verifies invalidation plus updated ordering/preview/unread state. No security/privacy surface and no unrelated scope creep.

VERDICT: APPROVE (comment-only; `agent-p1p` cannot formally approve its own PR)

SENSITIVE_PATHS: none

PR #459, opened 2026-07-10
fix: preserve same-second projection order

### LINKED ISSUE #451: Bug: MessageTimelineStore projection insert uses a (timelineAt, id) binary-search key, but the stored window preserves the runtime's own within-second tie-break — same-second live upserts are mis-ordered (and can be dropped in a detached window)
## Severity
**MEDIUM** (correctness / ordering). The projection insert path orders same-second messages by a key (`id`) that the stored window is **not** sorted by, so a newly-arrived same-second message can be placed above an earlier-received same-second message (a visible, repeatable reorder), and in a detached (scrolled-back) window a same-second upsert can be silently dropped. This is the projection/insertion-path counterpart of the load-path fix for the closed #7.

## Summary
`MessageTimelineStore` deliberately **never re-sorts** the window: it trusts the runtime's authoritative order and only dedups. `MarmotMapping.timeline` documents why — the runtime/database already tie-broke same-second records, and re-sorting in the client would reshuffle them. But `applyProjection` inserts live upserts using a binary search keyed on `TimelineSortKey = (timelineAt, id)`, and uses `messages.last` as the greatest key for the detached-window head test. Since `timelineAt` is second-granular, within any one second the stored block is ordered by the runtime tie-break — **not** by `id`. A binary search keyed on `id` over a block that is not `id`-sorted yields an arbitrary in-block index.

## Evidence
- `whitenoise-mac/Core/MarmotMapping.swift:206-208` — the store preserves runtime order and must not re-sort:
  ```swift
  // MarmotKit returns an authoritative timeline window. Keep that order
  // intact: `timelineAt` is second-granular, and re-sorting in the client
  // can reshuffle records that the runtime/database already tie-broke.
  ```
- `whitenoise-mac/Core/WorkspaceState.swift:285-289` / `:365-373` — `replace(with:)` and `rebuildIndexes()` only dedup (`uniquedLastWins`/`deduplicatedMessages`, order-preserving); there is no sort anywhere.
- `whitenoise-mac/Core/WorkspaceState.swift:409-422` — insertion uses an `id`-tie-broken binary search:
  ```swift
  private func insertionIndex(for item: MessageItem) -> Int {
      let key = TimelineSortKey(item)          // (timelineAt, id)
      var lowerBound = messages.startIndex
      var upperBound = messages.endIndex
      while lowerBound < upperBound {
          let middle = lowerBound + (upperBound - lowerBound) / 2
          if TimelineSortKey(messages[middle]) < key { lowerBound = middle + 1 }
          else { upperBound = middle }
      }
      return lowerBound
  }
  ```
- `whitenoise-mac/Core/WorkspaceState.swift:424-438` — the key's tie-break is `id` string comparison:
  ```swift
  static func < (lhs: TimelineSortKey, rhs: TimelineSortKey) -> Bool {
      if lhs.timelineAt != rhs.timelineAt { return lhs.timelineAt < rhs.timelineAt }
      return lhs.id < rhs.id
  }
  ```
- `whitenoise-mac/Core/WorkspaceState.swift:325, 342-344` — the detached-window head test treats `messages.last` as the max key and drops upserts beyond it:
  ```swift
  let newestKey = messages.last.map(TimelineSortKey.init)
  ...
  let isInsideDetachedWindow = newestKey.map { TimelineSortKey(item) <= $0 } ?? false
  guard anchoredToNewest || isInsideDetachedWindow else { continue }
  ```

## Failure scenarios
Same-second `timelineAt` collisions are common in active groups (the runtime timestamp is second-granular; #7 established this).

- **Anchored window — visible reorder.** Existing same-second block (runtime order A, then B) whose ids are, say, `idA < idB` but true receive order is A→B. A new same-second message C arrives via `.projection`. `insertionIndex` binary-searches C's `id` against `[A, B]`. Depending on where `idC` falls lexicographically, C is inserted *between* A and B (or before A) rather than after B, even though C was received last. Result: a repeatable, wrong on-screen ordering of same-second bubbles that does not reflect receive order.
- **Detached window — dropped upsert.** In a scrolled-back window (`anchoredToNewest == false`), `newestKey = messages.last` is the last element by *runtime* order, not the `id`-max of the head second. A same-second upsert whose `id` sorts **after** `messages.last.id` fails `TimelineSortKey(item) <= newestKey` → `guard ... else { continue }` → the message is not inserted at all until a full re-window.

The whole class is latent only because the store never re-sorts to reconcile the two orderings; the load path already chose runtime order (the #7 resolution), while the projection path silently reintroduces the `id` tie-break #7 warns against.

- **#7** (closed) fixed the divergent-sort-key reorder on the **load/merge** path by standardizing on the runtime order; the current `insertionIndex`/`messages.last` machinery is a *different* code path that reintroduces the `(timelineAt, id)` key, so this is the incomplete-fix counterpart, not the same site.
- **#331** is specifically about capturing `newestKey` *before* removals (already fixed by moving it after); this bug is independent of removals — it is the ordering key itself not matching the stored order.
- **#422** (rebuildIndexes cost), **#423** (reaction re-sort), **#419** (edit folding), **#330** (`timelineAt == 0`) are unrelated.

## Confidence boundary
The visible-reorder and detached-drop effects hold whenever the runtime's within-second tie-break differs from ascending `id` order — which both `MarmotMapping.swift:206-208` and #7 assert is the case ("the id tie-break … has no relationship to actual send order"). If the runtime happened to emit same-second records in ascending-`id` order, the mismatch would be moot; the code's own comments indicate it does not.

## Suggested fix
Make the insert path use the **same** ordering the store preserves. Options: (a) carry a stable per-record runtime sequence index from the FFI window and use it as the `TimelineSortKey` tie-break (instead of `id`), so binary search and the detached-head test match the stored order; or (b) resolve the insert position by scanning the actual same-second block for its runtime-order tail rather than binary-searching an `id` key over a non-`id`-sorted region.

## Scope note
Found during a full-codebase review; verified against the Swift store model.

agent-p1p:
Fixes #451

## Summary
- remove the lexical message-id tie-break from projection insertion and detached-window gating
- append new equal-timestamp projection rows after the authoritative runtime-ordered block
- add regressions for anchored same-second ordering and detached same-second acceptance while retaining strictly-newer suppression

## Verification
- added-line static/security scan: passed
- git diff --check: passed
- exhaustive timestamp upper-bound model: passed
- #451 anchored fixture model: fails under the old lexical key and passes under the timestamp-only upper bound
- independent pre-commit review: passed with no blocking findings
- macOS Swift format, Xcode unit tests, release build, and static analysis: not runnable on this Linux host; delegated to PR CI

## Sensitive paths
None. Client timeline-store ordering and unit tests only.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/459">
  <picture>
    Open in Stage
  </picture>
</a>


agent-p1p:
## Adversarial review

**Verdict: changes requested — 2 blocking findings.**

### Blocking: the timestamp-only key contradicts the vendored runtime ordering contract

This PR assumes MarmotKit's same-second order is not ascending message ID. The exact bundle in this PR says otherwise: `Vendored/MarmotKit/MARMOT_VERSION` is built from MDK `f9ef165a50410ca1f5b892316d71cb4f16dc4903`, where:

- [storage defines `(timeline_at, message_id_hex)` as the cross-client canonical order](https://github.com/marmot-protocol/mdk/blob/f9ef165a50410ca1f5b892316d71cb4f16dc4903/crates/storage-sqlite/src/timeline.rs#L20-L31);
- [the runtime sorts every materialized window by that same pair](https://github.com/marmot-protocol/mdk/blob/f9ef165a50410ca1f5b892316d71cb4f16dc4903/crates/marmot-app/src/runtime/subscriptions.rs#L346-L353);
- [detached-window projection gating compares the same pair](https://github.com/marmot-protocol/mdk/blob/f9ef165a50410ca1f5b892316d71cb4f16dc4903/crates/marmot-app/src/runtime/subscriptions.rs#L399-L476), with runtime regressions that specifically require a higher-ID same-second head message to be suppressed.

The change at `WorkspaceState.swift:322-345,410-421` therefore makes the client diverge from both the runtime's already-mutated subscription window and the database cursor order:

- Anchored, existing `a,b @ 100` plus new `0 @ 100`: runtime/current client produce `0,a,b`; this PR produces `a,b,0`. The new anchored test starts from `runtime-b,runtime-a`, which is not a window the vendored runtime can return.
- Detached, existing `a,b @ 100` plus new `c @ 100`: runtime suppresses `c`; this PR admits it and grows a head the runtime did not grow.

That guarantees a reshuffle on the next authoritative `.page` and breaks the documented client/runtime lockstep while detached. Keep `TimelineSortKey = (timelineAt, id)` unless the MDK/storage ordering contract is changed first and new bindings are vendored. With the current bundle, the issue premise is invalid; the smallest correct change is to close/revert this PR rather than add a second ordering contract.

### Blocking: Mac CI is red

[`PR Checks` failed in the unit-test step](https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29106321368/job/86407586157) and the release-build smoke step was skipped. The reported failures are in `messageMediaDiskCacheReadDeletionMaintainsTrackedFootprint` (3 expectations) and `workspaceCoalescesAndCachesSourceEpochZeroMediaReferenceResolution` (1 expectation); `Static Analysis` passed. Regardless of whether those failures are unrelated flakes, this PR currently has red CI and is not mergeable under Pip's gate.

No out-of-scope follow-up issue filed.

agent-p1p:
Addressed the adversarial review in 365ea2f by reverting the proposed change.

The vendored MDK runtime (`f9ef165a`) defines `(timeline_at, message_id_hex)` as the canonical storage order, re-sorts materialized windows by that pair, and uses the same pair for detached projection gating. The timestamp-only implementation and its fixtures therefore contradicted the actual runtime contract. The branch now has zero net diff from `master`; this PR is closed rather than merged.

The original SHA also had red Mac CI. No merge was attempted.

Issue #457, opened 2026-07-10
CI: restore the macOS PR test plan after media-cache changes


agent-p1p:
## Summary

The `Mac CI / PR Checks` unit-test step is red on `master` and on unrelated PRs after the recent media-cache changes. This blocks reliable PR validation.

## Evidence

- Master run at base SHA `0454dbf0dff6320802313037ef34bf44352706d7`: https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29100021269
  - `messageMediaDiskCacheReadDeletionMaintainsTrackedFootprint()` fails at `whitenoise_macTests.swift:2224` because it still expects `staleKey.cacheID == invalidatingKey.cacheID`.
  - PR #444 intentionally changed `MessageMediaDiskCacheKey.cacheID` to include `plaintextSha256`, so those keys now differ. The later assertions in that test consequently exercise a path that no longer exists.
- Unrelated PR #454 run: https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29100571786
  - Reproduces the stale cache-key test failure.
  - Also intermittently fails `workspaceCoalescesAndCachesSourceEpochZeroMediaReferenceResolution()` and `mediaAttachmentDownloadTimesOutAndReleasesLimiterSlot()`; neither failure touches #454's one-line SwiftUI diff.

## Expected behavior

- Update the tracked-footprint regression to exercise a valid read-path deletion under the new plaintext-hash cache identity.
- Stabilize or isolate the two media async tests if they reproduce independently.
- Verify the full 418-test PR plan is green on `master` before using it as a merge gate.

## Scope

Found while reviewing #454 / issue #452. This is out of scope for that transcript hit-testing fix; #454 changes only `whitenoise-mac/Views/MessengerShellView.swift`.

agent-p1p:
Additional reproduction from #461: https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29106886686

- The deterministic `messageMediaDiskCacheReadDeletionMaintainsTrackedFootprint()` failure exactly matches the current `master` run (three assertions).
- `workspaceCoalescesAndCachesSourceEpochZeroMediaReferenceResolution()` also intermittently lost the first downloaded reference (`expected sourceEpoch 11`, dictionary lookup was `nil`) even though `downloadMediaCallCount == 3` and all three attachment states loaded.
[...]

PR #461, opened 2026-07-10
fix: reconcile cache footprint after failed purge

### LINKED ISSUE #453: Bug: MessageMediaDiskCache.purgeAll zeroes trackedFootprint even when the on-disk removal fails, permanently under-counting surviving files and under-enforcing the disk cap
## Severity
**LOW** (eviction under-enforcement; does NOT self-correct). Unlike over-count drift, an under-count from this path never triggers `enforceEvictionPolicy`, so it is never reconciled — the encrypted media cache can grow past its 512 MB / 2048-entry bound indefinitely. Trigger (a filesystem error removing the app's own cache subtree while leaving files behind) is uncommon, hence LOW.

## Summary
`purgeAll` removes the cache root with `try?` (error swallowed) and then **unconditionally** sets `trackedFootprint` to zero. If `removeItem(at: root)` fails and the encrypted entries survive on disk, the accountant now claims an empty cache while real bytes remain. Because the incremental accountant only re-seeds from an authoritative scan when `trackedFootprint == nil`, a non-nil zero is never reconciled: subsequent stores increment from zero, so the cap is under-enforced by roughly the orphaned amount.

## Evidence
- `whitenoise-mac/Core/MessageMediaDiskCache.swift:319-327` — removal failure swallowed, footprint zeroed regardless:
  ```swift
  let task = beginPurge(scope: .all) { [self] in
      if let root {
          try? FileManager.default.removeItem(at: root)   // failure swallowed
      }
      if removeEncryptionKey { deleteKey() }
      trackedFootprint = CacheFootprint(entryCount: 0, byteCount: 0)  // unconditional
  }
  ```
- `whitenoise-mac/Core/MessageMediaDiskCache.swift:872-881` — the incremental path only re-seeds from an authoritative scan when the footprint is `nil`; a non-nil zero takes the incremental branch forever:
  ```swift
  private func updateTrackedFootprintAfterCommit(...) {
      if trackedFootprint == nil {
          trackedFootprint = Self.cacheFootprint(root: root)   // authoritative re-seed
          return
      }
      let tracked = trackedFootprint!
      ... // incremental add; surviving orphaned files never re-counted
  }
  ```

## Failure scenario
1. `purgeAll` runs; `removeItem(at: root)` throws (permissions, a file held open, an I/O error) — swallowed by `try?`. The encrypted cache files remain on disk (say ~500 MB).
2. `trackedFootprint` is set to `{entryCount: 0, byteCount: 0}` anyway.
3. Every later `store` uses `updateTrackedFootprintAfterCommit`'s incremental branch (footprint is non-nil), so it counts only newly-committed bytes and never re-counts the ~500 MB of survivors.
4. `enforceEvictionPolicy` is entered only when `trackedFootprint` exceeds the cap; starting from a false zero, that now requires ~512 MB of *additional* stores before any eviction — so on-disk usage reaches ~1 GB (survivors + new) before the cap engages. The drift never self-corrects because it is in the under-count direction.

- **#426** is `readDownload`'s on-disk deletions bypassing the accountant — *over*-count drift from reads, which self-corrects on the next enforce scan.
- **#377** is redundant full-tree walks on every store; **#328** is the original no-eviction-bound (fixed); **#356 / #334 / #376** concern corrupt/staging/collateral deletion.
- None address `purgeAll` optimistically zeroing the footprint when its own filesystem removal fails. This is the one drift path in the *under*-count (never-self-correcting) direction.

## Suggested fix
Only set `trackedFootprint = CacheFootprint(entryCount: 0, byteCount: 0)` when the removal actually succeeded (or the root did not exist). On a removal error, set `trackedFootprint = nil` so the next store re-seeds from an authoritative `cacheFootprint(root:)` scan, matching how `purgeAccount` forces a re-seed.

## Scope note
Found during a full-codebase review; verified against source.

agent-p1p:
Fixes #453

## Summary
- invalidate `trackedFootprint` when cache-root resolution or removal fails
- preserve an empty footprint only after confirmed root removal
- inject deterministic root-removal failure in tests and verify the next store re-scans survivors before enforcing the entry cap
- preserve encryption-key deletion even when filesystem removal fails

## Verification
- `git diff --check` (pass)
- independent pre-commit review after one fix cycle (pass; no blocking findings)
- macOS Swift format, build, analyze, and unit tests unavailable on this Linux host; GitHub Actions will run the repository's macOS CI

## Sensitive paths
None.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/461">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/MessageMediaDiskCache.swift
+++ b/whitenoise-mac/Core/MessageMediaDiskCache.swift
@@ -147,6 +147,7 @@ nonisolated enum MessageMediaDiskCacheKeychain {
 
 nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
     typealias DirectoryResolver = @Sendable () throws -> URL
+    typealias RootRemover = @Sendable (URL) throws -> Void
     typealias KeyProvider = @Sendable () throws -> SymmetricKey
     typealias KeyDeleter = @Sendable () -> Void
     typealias TimestampProvider = @Sendable () -> TimeInterval
@@ -178,6 +179,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
     private static let payloadFileName = "payload.bin"
 
     private let directoryResolver: DirectoryResolver
+    private let rootRemover: RootRemover
     private let keyProvider: KeyProvider
     private let keyDeleter: KeyDeleter
     private let timestampProvider: TimestampProvider
@@ -211,6 +213,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
 
     init(
         directoryResolver: @escaping DirectoryResolver = MessageMediaDiskCache.defaultDirectoryURL,
+        rootRemover: @escaping RootRemover = { try FileManager.default.removeItem(at: $0) },
         keyProvider: @escaping KeyProvider = MessageMediaDiskCacheKeychain.symmetricKey,
         keyDeleter: @escaping KeyDeleter = MessageMediaDiskCacheKeychain.deleteKey,
         timestampProvider: @escaping TimestampProvider = { Date().timeIntervalSince1970 },
@@ -218,6 +221,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
         sessionStartedAtUnixSeconds: TimeInterval = Date().timeIntervalSince1970
     ) {
         self.directoryResolver = directoryResolver
+        self.rootRemover = rootRemover
         self.keyProvider = keyProvider
         self.keyDeleter = keyDeleter
         self.timestampProvider = timestampProvider
@@ -315,16 +319,28 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
     }
 
     func purgeAll(removeEncryptionKey: Bool = false) async {
-        let root = try? directoryResolver()
+        let root: URL?
+        do {
+            root = try directoryResolver()
+        } catch {
+            root = nil
+        }
+        let removeRoot = rootRemover
         let deleteKey = keyDeleter
         let task = beginPurge(scope: .all) { [self] in
             if let root {
-                try? FileManager.default.removeItem(at: root)
+                do {
+                    try removeRoot(root)
+                    trackedFootprint = CacheFootprint(entryCount: 0, byteCount: 0)
+                } catch {
+                    trackedFootprint = nil
+                }
+            } else {
+                trackedFootprint = nil
             }
             if removeEncryptionKey {
                 deleteKey()
             }
-            trackedFootprint = CacheFootprint(entryCount: 0, byteCount: 0)
         }
         await task.task.value
         finishPurge(task)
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce RootRemover abstraction for cache management](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/461/chapters/1) |
| 2 | [Reconcile cache footprint after failed purge](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/461/chapters/2) |
| 3 | [Verify footprint re-scanning after failed removal](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/461/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/461">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 204ad0b2bc6d955a41e33e9f26357be71c34a6e9 on Jul 10, 2026 4:18pm UTC.</sub>

agent-p1p:
## Pip adversarial review

**Verdict: blocking** — 1 blocking finding, 0 suggestions, 0 nitpicks.

### Blocking
- **CI is red on head `204ad0b`.** The [PR Checks job](https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29106886686/job/86409473177) failed in Unit tests with four issues across `messageMediaDiskCacheReadDeletionMaintainsTrackedFootprint` and `workspaceCoalescesAndCachesSourceEpochZeroMediaReferenceResolution`. The new `messageMediaDiskCachePurgeAllReSeedsFootprintWhenRootRemovalFails` regression test itself passed, and neither failing test was changed by this PR, but Pip policy forbids merging any PR that has had red CI. This PR is therefore not mergeable; address the failing suite and replace it with a clean-CI PR rather than relying on a rerun.

### Code review
- `purgeAll` now records an empty footprint only after confirmed root removal.
- Resolver/removal failures invalidate `trackedFootprint`, so the next committed store performs the authoritative scan before eviction.
- Encryption-key deletion remains independent of filesystem success.
- The injected remover is narrow and the regression test exercises survivor recounting plus cap enforcement without broadening production behavior.

### Verification
- `git diff --check origin/master...HEAD`: pass
- Swift format lint and project sanity checks: pass
- Static Analysis: pass
- New failed-removal regression: pass in CI
- Full PR Checks: **fail** (blocking, details above)

No sensitive paths touched.

agent-p1p:
Closing this draft because its first CI run inherited the current `master` failure, and Pip never merges an ever-red PR.

Evidence:
- PR #461: the new `messageMediaDiskCachePurgeAllReSeedsFootprintWhenRootRemovalFails()` regression passed; Static Analysis passed.
- `PR Checks` failed in the unchanged `messageMediaDiskCacheReadDeletionMaintainsTrackedFootprint()` test with the same three assertions as current `master` at base SHA `0454dbf`: https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29100021269
- This run also hit one intermittent `workspaceCoalescesAndCachesSourceEpochZeroMediaReferenceResolution()` assertion; upstream tracking issue #457 already covers both PR-plan defects.

After #457 lands, this branch should be rebased onto fixed `master` and published as a fresh PR so its first CI run can be green.

PR #537, opened 2026-07-17
fix: format collection counts as 64-bit values

### LINKED ISSUE #533: Bug: `%d` format specifier used with 64-bit Int counts ("%d members" / "%d attachments") — specifier/width mismatch
## Severity
**LOW** (latent correctness/consistency; no realistic failure at real-world counts)

## Summary
`String(format: "%d", …)` is used with 64-bit `Int` count values. `%d` consumes a 32-bit `CInt`, so it reads only the low 32 bits of the 64-bit argument. Values above 2³¹ would misprint. The codebase is inconsistent here: the retention-label code at `MarmotMapping.swift:711-719` correctly uses `%llu` + `CUnsignedLongLong`.

## Root cause
`whitenoise-mac/Models/MessengerModels.swift`:
- `:237` — `String(format: L10n.string("%d members"), members.count)` (`members.count` is `Int`)
- `:337` — `String(format: L10n.string("%d attachments"), attachments.count)` (`attachments.count` is `Int`)

## Failure scenario
None realistic — member/attachment counts never approach 2³¹. Flagged for correctness and consistency with the `%llu` pattern used elsewhere.

## Suggested fix
Use `%ld` with the `Int` value, or cast the argument to `CInt`, matching the explicit-width pattern already used at `MarmotMapping.swift:711-719`.

agent-p1p:
Fixes #533

## Summary
- format member and attachment collection counts through fixed semantic APIs using `%lld` and an explicit `CLongLong` argument
- preserve the existing translated `%lld members` catalog key instead of introducing an untranslated key
- cover values above `Int32.max` through the production formatter APIs and assert the selected Spanish localization path

## Verification
- `git diff --check`
- targeted source/catalog assertions confirm the unsafe `%d`/`%llu` call sites are gone, both production paths use the semantic 64-bit formatters, and the existing nine member-count translations remain intact
- `Localizable.xcstrings` parses as valid JSON
- Xcode/Swift checks not run locally: this worker is on Linux without `xcodebuild`, `swift`, or `swift-format`; Mac CI runs the repository gates

## Sensitive paths
None.

### DIFF
--- a/whitenoise-mac/Models/MessengerModels.swift
+++ b/whitenoise-mac/Models/MessengerModels.swift
@@ -6,8 +6,16 @@ import MarmotKit
 import UniformTypeIdentifiers
 
 nonisolated enum CollectionCountFormat {
-    static func localizedLabel(template: String, count: Int) -> String {
-        String(format: L10n.string(template), CUnsignedLongLong(UInt64(count)))
+    static func memberLabel(count: Int) -> String {
+        localizedLabel(template: "%lld members", count: count)
+    }
+
+    static func attachmentLabel(count: Int) -> String {
+        localizedLabel(template: "%lld attachments", count: count)
+    }
+
+    private static func localizedLabel(template: String, count: Int) -> String {
+        String(format: L10n.string(template), CLongLong(count))
     }
 }
 
@@ -240,7 +248,7 @@ struct GroupDetailsSnapshot: Hashable {
     let disappearingMessageSecs: UInt64
 
     var memberCountLabel: String {
-        CollectionCountFormat.localizedLabel(template: "%llu members", count: members.count)
+        CollectionCountFormat.memberLabel(count: members.count)
     }
 
     var disappearingMessagesEnabled: Bool { disappearingMessageSecs > 0 }
@@ -340,7 +348,7 @@ nonisolated struct MessageMediaAttachment: Identifiable, Hashable {
         if attachments.count == 1 {
             return first.previewLabel
         }
-        return CollectionCountFormat.localizedLabel(template: "%llu attachments", count: attachments.count)
+        return CollectionCountFormat.attachmentLabel(count: attachments.count)
     }
 }
 
--- a/whitenoise-mac/Models/MessengerModels.swift
+++ b/whitenoise-mac/Models/MessengerModels.swift
@@ -5,6 +5,12 @@ import ImageIO
 import MarmotKit
 import UniformTypeIdentifiers
 
+nonisolated enum CollectionCountFormat {
+    static func localizedLabel(template: String, count: Int) -> String {
+        String(format: L10n.string(template), CUnsignedLongLong(UInt64(count)))
+    }
+}
+
 struct AccountItem: Identifiable, Hashable {
     let id: String
     let accountRef: String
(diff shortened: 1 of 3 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Introduce 64-bit collection count formatter](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/537/chapters/1) |
| 2 | [Apply 64-bit formatting to member and attachment labels](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/537/chapters/2) |
| 3 | [Verify 64-bit formatting and localization](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/537/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/537">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 7d9f8d6c27f85de64b15b0ddbab173594d784629 on Jul 17, 2026 7:44pm UTC.</sub>

agent-p1p:
## Pip adversarial review — changes requested

Reviewed head `87c99ca3ca89bc3ab2bbd052c5d693ed96e1f487` against #533.

### Blocking (1)

- **`whitenoise-mac/Models/MessengerModels.swift:243` — the new member-count key bypasses nine existing translations.** `Localizable.xcstrings` already has a manually maintained `%lld members` entry translated for `de`, `es`, `fr`, `it`, `pt`, `ru`, `tr`, `zh-Hans`, and `zh-Hant`. This PR instead looks up `%llu members`, which has no catalog entry, so `L10n.string` falls through to the English key in every selected non-English bundle. The old `%d` call also missed the catalog entry; replacing it with a third spelling leaves that defect in place and contradicts the PR verification note that there are no affected pre-existing keys. Reuse `%lld members` with an explicit signed 64-bit argument such as `CLongLong(count)`, or migrate the catalog entry and all translations to `%llu`. Add a selected-language assertion so the exact lookup key is covered, not only the C-vararg width.

### Suggestion (1)

- **`whitenoise-macTests/PureValueTests.swift:36-50` — the regression test supplies the same free-form template as production.** It proves `%llu` formats a large value, but it cannot catch the production lookup-key mistake above because the test repeats that mistake. Prefer fixed semantic formatter APIs for members/attachments, or add a production-path/localized-key assertion while retaining the above-`Int32.max` width check.

### Checks

- `git diff --check origin/master...HEAD` — pass
- Independent catalog/source assertion — pass: `%lld members` has 9 locale entries; both proposed `%llu` keys have 0
- Mac CI unit-test step — pass
- Mac CI static analysis — pass
- Mac CI release smoke build was still running when this review was posted

No sensitive paths touched. No out-of-scope follow-up issue filed.

agent-p1p:
Addressed the blocking localization finding in `7d9f8d6`:

- member counts now reuse the existing `%lld members` catalog key with `CLongLong(count)`
- production exposes fixed member/attachment formatter APIs rather than accepting a free-form template
- the large-count regression calls those production APIs
- a serialized selected-language test asserts the Spanish result (`Miembros de 3`)

Mac CI was retriggered by the push.

mubarakcoded:
@agent-p1p superceeded

PR #573, opened 2026-07-18
perf: avoid decrypting media metadata during eviction

### LINKED ISSUE #508: Performance: over-cap eviction re-decrypts the ENTIRE media cache on every store (metadataStatus AES.GCM.open + JSONDecode per entry, under fileMutationLock)
## Severity
**MEDIUM** (performance)

## Summary
Once the media cache is at/near its cap — the normal steady state for an active user — every successful `store` calls `enforceEvictionPolicy`, which (because it needs each entry's `cachedAtUnixSeconds` for LRU ordering) runs `cacheScan`, and `cacheScan` calls `metadataStatus` on **every** entry: `Data(contentsOf:)` + `AES.GCM.open` + `JSONDecode` for up to ~`maxEntryCount` records, all while holding `fileMutationLock`.

## Location / Evidence
`whitenoise-mac/Core/MessageMediaDiskCache.swift`
- `enforceEvictionPolicy(root:symmetricKey:)` — `:924-966`; called from the commit path at `:515`, inside the `fileMutationLock` region (lock at `:487`).
- `cacheScan(root:symmetricKey:)` — `:996-1053`; for every entry it calls `metadataStatus(...)` (`:1025`).
- `metadataStatus(in:symmetricKey:)` decrypts the entry metadata (`AES.GCM.open` at `:1115`) and JSON-decodes it, purely to read `cachedAtUnixSeconds`.

The early-return guard at `:936` only avoids the *eviction loop* when the scan happens to come back under cap; the full decrypt+decode **scan itself already ran** by then.

## Failure scenario
Cache holds several hundred entries near `maxTotalBytes`. User scrolls a media-heavy chat; each auto-downloaded attachment stores, pushing over the cap, so **each** store performs a full N-entry decrypt+decode sweep under `fileMutationLock`, evicts one, and the next store repeats. This is a repeated O(cache) *cryptographic* scan on the hot path, blocking concurrent commits and the read-path deletion for the duration.

#377 removed the *under-cap* stat-only walk by introducing `trackedFootprint`. That optimization does not touch the over-cap path, which is far more expensive: it decrypts and decodes every entry (not a `stat`) because LRU ordering requires `cachedAtUnixSeconds` from the encrypted metadata. #426 (closed) is about read-path deletions drifting the accountant — also distinct.

## Suggested fix
Persist `cachedAtUnixSeconds` (and the byte count) outside the encrypted metadata so eviction ordering doesn't require decryption — e.g. encode it into the entry directory name, use the entry file's mtime, or keep an in-memory `[cacheID: (cachedAt, byteCount)]` index and only rescan on drift. Then the over-cap path becomes a cheap stat/sort instead of a full-cache decrypt.

## Scope note
Found during an automated full-codebase review; verified against the current source.

agent-p1p:
## Summary
- persist each cache entry timestamp in its directory modification date
- order pressure eviction with filesystem metadata instead of decrypting and decoding every metadata record
- retain structural-unavailable priority and add regression coverage for encrypted-metadata independence

## Test plan
- [x] git diff --check
- [x] independent staged-diff review
- [x] macOS unit tests and release build smoke (GitHub Actions; exact-head Mac CI passed)

Fixes #508

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/MessageMediaDiskCache.swift
+++ b/whitenoise-mac/Core/MessageMediaDiskCache.swift
@@ -325,7 +325,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                 Self.discardPreparedEntry(prepared)
                 return
             }
-            commitPreparedEntry(prepared, start: start, symmetricKey: symmetricKey)
+            commitPreparedEntry(prepared, start: start)
         }
     }
 
@@ -490,8 +490,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
 
     private func commitPreparedEntry(
         _ prepared: PreparedEntry,
-        start: AccessHandle,
-        symmetricKey: SymmetricKey
+        start: AccessHandle
     ) {
         // Hold only the narrow filesystem-mutation lock across the slow create/remove/move
         // so generation reads and purge bookkeeping (on `lock`) never block on this commit.
@@ -527,7 +526,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                 replacedByteCount: replacedByteCount,
                 root: prepared.root
             )
-            enforceEvictionPolicy(root: prepared.root, symmetricKey: symmetricKey)
+            enforceEvictionPolicy(root: prepared.root)
         } catch {
             try? FileManager.default.removeItem(at: prepared.stagingDirectory)
         }
@@ -748,6 +747,12 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
                 to: stagingDirectory.appendingPathComponent(payloadFileName),
                 options: [.atomic, .completeFileProtection]
             )
+            // Persist eviction ordering outside the sealed metadata. Reads do not change a
+            // directory's modification date, and moving the staged directory preserves it.
+            try FileManager.default.setAttributes(
+                [.modificationDate: Date(timeIntervalSince1970: cachedAtUnixSeconds)],
+                ofItemAtPath: stagingDirectory.path
+            )
             return PreparedEntry(root: root, stagingDirectory: stagingDirectory, finalDirectory: finalDirectory)
         } catch {
             discardStagingDirectory(stagingDirectory, root: root)
@@ -937,7 +942,7 @@ nonisolated final class MessageMediaDiskCache: @unchecked Sendable {
         )
     }
 
-    private func enforceEvictionPolicy(root: URL, symmetricKey: SymmetricKey) {
+    private func enforceEvictionPolicy(root: URL) {
         guard let tracked = trackedFootprint else { return }
         let trackedTotalBytes = Self.addingWithSaturation(
             tracked.byteCount,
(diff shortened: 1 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Persist cache entry timestamps in directory metadata](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573/chapters/1) |
| 2 | [Implement filesystem-based eviction timestamp retrieval](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573/chapters/2) |
| 3 | [Decouple cache eviction from metadata decryption](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573/chapters/3) |
| 4 | [Update and add regression tests for eviction](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/573">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 6de999e5aab86faae5082180239444144bdb1de8 on Jul 18, 2026 1:24am UTC.</sub>

agent-p1p (commented):
Adversarial review at exact head `aa6670112e384e21dca5dafbef73d1425c701ddb`.

BLOCKING:
- none

SUGGESTION:
- `whitenoise-macTests/whitenoise_macTests.swift:2865` — Strengthen `messageMediaDiskCacheEvictionReclaimsUnavailableEntriesBeforeReadableEntries` so the structurally unavailable entry is newer than a readable survivor when pressure is applied. The current unavailable entry is stored first, so ordinary mtime ordering would evict it even if `evictionTimestamp` stopped assigning `-Double.infinity`; the test therefore does not independently protect the structural-unavailable priority this PR retains. A cap of two with readable → unavailable → pressure-trigger entry would distinguish the behaviors.

NITPICK:
- none

SUMMARY: The fix resolves #508. `cacheScan` no longer reads, AES-GCM-opens, or JSON-decodes every metadata record under `fileMutationLock`; over-cap enforcement now uses file-size/resource-value stats and deterministic mtime ordering. The timestamp is stamped after both atomic writes and the same-root staging move preserves the entry directory metadata. Existing entries retain a usable natural directory mtime, every scanned entry remains counted and removable, and tracked committed/staging accounting is unchanged. The new regression proves eviction ordering is independent of encrypted metadata. Exact-head Mac CI passed format, project sanity, unit tests, release build smoke, and static analysis; `git diff --check` also passes. No correctness, security, or unnecessary-complexity blocker found.

VERDICT: APPROVE

SENSITIVE_PATHS: `whitenoise-mac/Core/MessageMediaDiskCache.swift` — encrypted media-cache eviction path changed from metadata decryption to filesystem timestamps; AES-GCM sealing/opening, AAD, key material, and Keychain lifecycle are unchanged.

agent-p1p:
Addressed the adversarial review suggestion in `6de999e`: the test now stores a readable entry first, a structurally unavailable newer entry second, then a third entry to trigger pressure. Ordinary timestamp ordering would evict the readable survivor, so the test now independently protects unavailable-entry priority.

agent-p1p:
@coderabbitai review

PR #649, opened 2026-07-20
fix: accept hinted edit target tags

### LINKED ISSUE #648: Robustness: editTargetMessageId silently downgrades a valid edit to a retract when the target "e" tag carries a NIP-10 relay hint/marker or when a second "e" tag is present
## Severity
**LOW** (interop robustness — contingent on the wire format of edit events from other Marmot-compatible clients; see caveat)

## Summary
`MessageEditOverlay.editTargetMessageId(in:)` accepts a target only from an `"e"` tag of **exactly** two elements and only when there is **exactly one** `"e"` tag. Any other shape returns `nil`, and the caller maps a `nil` target to `.retract(...)`, which *removes* the edit overlay. So an edit event whose target tag is the standard NIP-10 form `["e", <id>, <relay-url>]` / `["e", <id>, <relay>, <marker>]`, or an edit that also carries a second `"e"` tag (e.g. reply-and-edit), is treated as a **retraction** — a genuine edit silently fails to apply and the original text keeps showing.

## Location / Evidence
`whitenoise-mac/Core/MarmotMapping.swift:186-198`:
```swift
fileprivate static func editTargetMessageId(in tags: [MessageTagFfi]) -> String? {
    var target: String?
    for tag in tags where tag.values.first == "e" {
        guard tag.values.count == 2,                         // rejects ["e", id, relay, ...]
            let candidate = tag.values.dropFirst().first?.nilIfBlank,
            target == nil                                    // rejects a 2nd "e" tag
        else {
            return nil                                       // -> caller turns nil into .retract
        }
        target = candidate
    }
    return target
}
```
Caller: `mutations(from:)` at `:163-164` maps `nil` → `.retract(editMessageIdHex:)`.

## Failure scenario
A peer (or another Marmot-compatible client) sends a message-edit event whose target `"e"` tag includes a relay hint (`["e", <targetId>, wss://relay.example]`) — the common NIP-10 form. `editTargetMessageId` returns `nil`, `mutations` emits `.retract`, and the edit overlay is dropped: the recipient never sees the edited text and keeps rendering the pre-edit body.

## Confidence caveat
The in-repo tests only exercise the exact 2-element `["e", "target"]` form, which suggests the local core may normalize edit tags to that shape before they cross the FFI. If cross-client edit events never carry relay hints or extra `"e"` tags, this is by-contract-correct and not a live bug. It is worth flagging because the app parses raw peer `record.tags` here itself (unlike replies, which arrive pre-parsed as `record.replyToMessageIdHex`), so the tolerance should be confirmed against the actual wire format.

No filed issue covers edit-tag parsing tolerance in `MarmotMapping`.

## Suggested fix
Match the target as *the first* `"e"` tag with `values.count >= 2` and a non-blank second element, rather than requiring exactly one `"e"` tag of exactly length 2.

agent-p1p:
## Summary
- accept the first valid `e` edit-target tag even when it carries a NIP-10 relay hint or marker
- preserve the first valid edit target when additional `e` tags are present
- cover relay-hinted and multi-`e` edit records with regression tests

## Test plan
- [x] `git diff --check`
- [x] static parser regression check
- [ ] GitHub Mac CI (Xcode is unavailable on this Linux worker)

Fixes #648

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/649">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/MarmotMapping.swift
+++ b/whitenoise-mac/Core/MarmotMapping.swift
@@ -184,17 +184,10 @@ nonisolated struct MessageEditOverlay: Equatable, Sendable {
     }
 
     fileprivate static func editTargetMessageId(in tags: [MessageTagFfi]) -> String? {
-        var target: String?
-        for tag in tags where tag.values.first == "e" {
-            guard tag.values.count == 2,
-                let candidate = tag.values.dropFirst().first?.nilIfBlank,
-                target == nil
-            else {
-                return nil
-            }
-            target = candidate
-        }
-        return target
+        tags.lazy
+            .filter { $0.values.first == "e" }
+            .compactMap { $0.values.dropFirst().first?.nilIfBlank }
+            .first
     }
 }
 
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -4493,6 +4493,59 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func messageEditAcceptsNip10RelayHintedTargetTag() async throws {
+        let editRecord = timelineMessage(
+            id: "edit",
+            groupIdHex: "group",
+            sender: "alice",
+            plaintext: "Edited",
+            kind: 1_009,
+            tags: [MessageTagFfi(values: ["e", "target", "wss://relay.example", "reply"])],
+            recordedAt: 1_800_000_030
+        )
+
+        #expect(
+            MessageEditOverlay.mutations(from: [editRecord]) == [
+                .upsert(
+                    MessageEditOverlay(
+                        targetMessageIdHex: "target",
+                        editMessageIdHex: "edit",
+                        sender: "alice",
+                        plaintext: "Edited",
+                        timelineAt: 1_800_000_030
+                    ))
+            ])
+    }
+
+    @MainActor
+    @Test func messageEditUsesFirstValidTargetWhenAdditionalETagsArePresent() async throws {
+        let editRecord = timelineMessage(
+            id: "edit",
+            groupIdHex: "group",
+            sender: "alice",
+            plaintext: "Edited",
+            kind: 1_009,
+            tags: [
+                MessageTagFfi(values: ["e", "target"]),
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Relax edit target tag validation](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/649/chapters/1) |
| 2 | [Verify edit target tag parsing logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/649/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/649">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 0b16dd6bda1c6e9d68f0d81d13a425b7cee92fa0 on Jul 20, 2026 9:16am UTC.</sub>

agent-p1p (commented):
## Pip adversarial review — exact head `69263dd`

BLOCKING:
- none

SUGGESTION:
- **Lock in target-less edit handling:** `whitenoise-macTests/whitenoise_macTests.swift:4495` — The parser rewrite correctly still maps an edit with no valid `e` target (no `e` tag or a blank target id) to `.retract`, but the new tests only cover successful hinted/multi-tag parsing. A small negative regression assertion would make that retained contract explicit. Non-blocking.

NITPICK:
- none

SUMMARY: The implementation is minimal and solves #648 as described: it selects the first non-blank target from `e` tags, tolerates relay hints/markers and extra `e` tags, and preserves the existing no-valid-target → retract behavior. The focused tests exercise both newly accepted shapes. Exact-head Mac CI passed format lint, project sanity checks, unit tests, arm64 release build, and static analysis. No security/privacy or scope-creep concerns.

VERDICT: APPROVE

SENSITIVE_PATHS: none

*Approval-equivalent COMMENTED review; GitHub does not allow `agent-p1p` to formally approve its own PR.*

agent-p1p:
Addressed the non-blocking negative-path suggestion in 0b16dd6: missing-target and blank-target edit records now explicitly assert the retained retract behavior.

PR #650, opened 2026-07-20
test: preserve notification replay dedup across account switches

### LINKED ISSUE #646: Bug: deliveredNotificationKeys dedup set is cleared on sign-out / removal / Delete-All but NOT on active-account switch — retained keys from the previous identity can suppress a genuine notification for the new one (mirror of #628)
## Severity
**LOW** (in-memory-only; fails toward *under*-notifying, not a content leak; requires a `notificationKey` collision across identities within the 256-entry window)

## Summary
The notification-dedup set `deliveredNotificationKeys` (a single process-global `Set<String>`, LRU-bounded to 256 via `deliveredNotificationKeyOrder`) is treated as account-scoped residue and wiped on sign-out, account removal, and "Delete All Local Data" — but a plain **active-account switch** does not clear it. After switching A→B, account A's remembered keys keep occupying the shared 256-entry dedup window under B.

## Location / Evidence
- Declared process-global: `whitenoise-mac/Core/WorkspaceState.swift:1060-1061`.
- Consulted / populated: `whitenoise-mac/Core/WorkspaceState+Notifications.swift:21` (early-return dedup), `:157-162` (insert + LRU trim).
- **Cleared** on sign-out / removal (`resetActiveAccountUIState`): `whitenoise-mac/Core/WorkspaceState+Account.swift:357-358`; and on Delete-All: `:682-683`.
- **Not cleared** in `prepareForActiveAccountSwitch` (`whitenoise-mac/Core/WorkspaceState+Account.swift:84-117`), which clears read markers, peer caches, media caches, `conversationMetadataByChat`, group-member cache, etc., but leaves `deliveredNotificationKeys` / `deliveredNotificationKeyOrder` intact.

## Failure scenario
User switches from account A to account B. A genuine incoming notification for B whose `notificationKey` collides with a key still retained from A's session is silently suppressed by the early-return at `WorkspaceState+Notifications.swift:21` — B never sees a banner it should have. (In-memory only; no cross-account content is exposed.)

#628 is the exact mirror image — `conversationMetadataByChat` cleared on switch but not on sign-out. #241/#429 cover read markers. No filed issue covers `deliveredNotificationKeys` on the account-switch path.

## Suggested fix
Add `deliveredNotificationKeys.removeAll(); deliveredNotificationKeyOrder.removeAll()` to `prepareForActiveAccountSwitch`, or (better) make the dedup set account-scoped (keyed by account id) so it is structurally partitioned rather than clear-on-transition.

agent-p1p:
## Summary
- preserve the client-wide delivered-notification replay history across active-account switches
- verify the vendored MDK account-namespaced key contract with a behavior regression
- prove an account-A replay stays suppressed after A→B while account B receives its distinct key for the same message ID

## Review resolution
The original clear-on-switch change was removed. Vendored MDK `f3ece0e3` emits `message:<account_id_hex>:<message_id_hex>` and `invite:<account_id_hex>:<invite_ref>`, so the reported cross-account collision cannot occur. Clearing the shared history instead reopened a duplicate-alert window because the notification listener is client-wide.

## Test plan
- [x] Source invariant probe: original head RED, revised branch GREEN
- [x] `git diff --check origin/master...HEAD`
- [x] committed-tree conflict-marker and duplicate-test-name sweeps
- [x] Mac CI PR Checks and Static Analysis

Fixes #646

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/650">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Account.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Account.swift
@@ -111,13 +111,9 @@ extension WorkspaceState {
         clearGroupMemberCache()
         // Read markers are keyed by groupIdHex only; stale entries from the
         // previous account suppress the first legitimate advance for a shared
-        // group id under the new identity. Notification keys are likewise
-        // account-scoped and can suppress a genuine delivery after a collision.
-        // See #429/#646.
+        // group id under the new identity. See #429.
         lastMarkedReadMarkers.removeAll()
         lastConfirmedReadMarkers.removeAll()
-        deliveredNotificationKeys.removeAll()
-        deliveredNotificationKeyOrder.removeAll()
         refreshObservabilityRuntime()
     }
 
--- a/whitenoise-mac/Core/WorkspaceState+Account.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Account.swift
@@ -111,9 +111,13 @@ extension WorkspaceState {
         clearGroupMemberCache()
         // Read markers are keyed by groupIdHex only; stale entries from the
         // previous account suppress the first legitimate advance for a shared
-        // group id under the new identity. See #429.
+        // group id under the new identity. Notification keys are likewise
+        // account-scoped and can suppress a genuine delivery after a collision.
+        // See #429/#646.
         lastMarkedReadMarkers.removeAll()
         lastConfirmedReadMarkers.removeAll()
+        deliveredNotificationKeys.removeAll()
+        deliveredNotificationKeyOrder.removeAll()
         refreshObservabilityRuntime()
     }
 
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -1542,7 +1542,7 @@ struct whitenoise_macTests {
     }
 
     @MainActor
-    @Test func prepareForActiveAccountSwitchClearsAccountScopedResidue() async throws {
+    @Test func prepareForActiveAccountSwitchClearsReadMarkers() async throws {
         let primary = AccountSummaryFfi(
             label: "Desktop Account",
             accountIdHex: "abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890",
@@ -1574,8 +1574,6 @@ struct whitenoise_macTests {
         )
         state.lastMarkedReadMarkers["shared-group"] = marker
         state.lastConfirmedReadMarkers["shared-group"] = marker
-        state.deliveredNotificationKeys.insert("notif-from-primary")
-        state.deliveredNotificationKeyOrder.append("notif-from-primary")
         state.composeContacts = [
             ComposeContact(
                 accountIdHex: String(repeating: "2", count: 64),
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Expose message ID in notification test helper](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/650/chapters/1) |
| 2 | [Verify notification deduplication during account switches](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/650/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/650">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2ba9a42eb6cb8f92a38ffea63d5211f2322aaafe on Jul 20, 2026 9:08am UTC.</sub>

agent-p1p:
BLOCKING:
- The alleged cross-identity key collision is unreachable with the vendored core, while this change weakens real replay suppression: `whitenoise-mac/Core/WorkspaceState+Account.swift:119-120`. `Vendored/MarmotKit/MARMOT_VERSION` pins MDK `f3ece0e3`, whose notification builder already namespaces every key by the receiving account ID: messages use `message:<account_id_hex>:<message_id_hex>` and invites use `invite:<account_id_hex>:<invite_ref>` ([MDK source at the exact vendored commit, message keys](https://github.com/marmot-protocol/mdk/blob/f3ece0e31a568bdaab239e5808cf740b46b658b6/crates/marmot-app/src/notifications.rs#L1305-L1308), [invite keys](https://github.com/marmot-protocol/mdk/blob/f3ece0e31a568bdaab239e5808cf740b46b658b6/crates/marmot-app/src/notifications.rs#L1351-L1353)). Different identities therefore cannot produce the same shipped `notificationKey` in the scenario from #646. Conversely, `subscribeNotifications()` is client-wide and remains live during `prepareForActiveAccountSwitch`; clearing the process-wide 256-key history means a previously delivered update replayed after a switch is no longer suppressed and can alert again. Remove the two clears (and the regression test built around an impossible unscoped fixture key), or demonstrate a real collision source not produced by the pinned MDK before changing dedup semantics.
- The test proves the implementation detail rather than the reported behavior: `whitenoise-macTests/whitenoise_macTests.swift:1577-1598`. It injects the artificial key `notif-from-primary`, which does not follow the account-namespaced format emitted by the runtime, then only asserts that internal containers were emptied. It cannot establish that a genuine B notification was previously suppressible. Any replacement regression must drive `handleNotificationUpdate` with runtime-valid keys and verify posted requests; with the current MDK format, A and B naturally have distinct keys.

SUGGESTION:
- none

NITPICK:
- none

SUMMARY: The patch is small and internally consistent, but issue #646's collision premise conflicts with the exact MDK version shipped by this app. The change creates a real duplicate/re-alert window by discarding client-wide replay history on a UI account switch, and its state-only test masks the unreachable premise with an invalid fixture key.

VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: none

agent-p1p:
Addressed the blocking findings at head `2ba9a42`:

- removed the account-switch dedup clears and restored the production file exactly to `master`
- removed the state-only, unscoped fixture assertion
- added a behavior regression using vendored-MDK-valid `message:<account_id_hex>:<message_id_hex>` keys
- verified an account-A replay after A→B remains suppressed while B receives its distinct key for the same message ID

Mac CI is running on the revised head.

agent-p1p:
Revised head `2ba9a42` is green: Swift format, project sanity, unit tests, arm64 release build, and static analysis all passed. No production diff remains; the PR now adds the runtime-valid behavior regression only.

PR #654, opened 2026-07-20
fix: cache media after in-memory store pruning

### LINKED ISSUE #503: Performance: a completed media download is discarded (and re-fetched) when the in-memory store was pruned+recreated mid-fetch by a chat switch
## Severity
**LOW** (performance / redundant network + re-decrypt; no data corruption)

## Summary
On the media-download success path, both the in-memory publish **and** the disk-cache store are gated behind the same `canPublishMediaDownloadState(...)` predicate. That predicate is `isMediaDisplayAllowed(...) && mediaDownloads[key] === stateStore`. When it fails **only** because the store object was replaced by a `pruneMediaDownloadCache` cycle (same account still active, display still allowed), the fully-fetched, already-decrypted payload is dropped on the floor and is never written to the disk cache — even though the disk store has its own independent generation guard (`isMediaDiskStoreAllowed`) that would correctly admit it. The next time the tile appears, its recreated `.idle` store triggers a full network re-download of bytes we already had.

## Location / Evidence
`whitenoise-mac/Core/WorkspaceState+Media.swift:272-298`

```swift
let download = try await withMediaAttachmentDownloadTimeout { ... }
guard
    canPublishMediaDownloadState(forKey: key, stateStore: stateStore, accountId: accountId, groupIdHex: groupIdHex)
else {
    stateStore.update(.idle)
    return                      // <-- also skips scheduleMediaDiskCacheStore below
}
let mediaDownload = MessageMediaDownload(...)
stateStore.update(.loaded(mediaDownload))
scheduleMediaDiskCacheStore(mediaDownload, for: cacheKey, accountId: accountId, storeGuard: storeGuard)
```

`canPublishMediaDownloadState` (line 335) requires store identity:
```swift
isMediaDisplayAllowed(forAccountId: accountId, groupIdHex: groupIdHex)
    && mediaDownloads[key] === stateStore
```

But `scheduleMediaDiskCacheStore` is *already* independently guarded by a generation check that does **not** depend on store identity (line 126):
```swift
guard isMediaDiskStoreAllowed(forAccountId: accountId, storeGuard: storeGuard) else { return }
```

And `pruneMediaDownloadCache` drops+recreates the store on chat switch (line 656-660):
```swift
for key in removedKeys {
    mediaDownloads[key]?.update(.idle)
    mediaDownloads[key] = nil        // <-- old stateStore is no longer === mediaDownloads[key]
}
```

## Failure scenario
1. User opens chat A; a large attachment begins downloading (task captures `stateStore = mediaDownloads[key]`).
2. User switches to chat B and back to A **before** the fetch completes. On the switch, `pruneMediaDownloadCache(keeping:)` removes chat A's keys, so `mediaDownloads[key]` is now `nil`/a new object.
3. The in-flight download completes. `mediaDownloads[key] === stateStore` is now **false**, so the `guard` returns early — the decrypted payload is neither published nor **written to disk**, despite `isMediaDiskStoreAllowed` still being true.
4. Back on chat A, the recreated `.idle` store re-fires `loadMediaAttachment`, causing a **second full network download + decrypt** of the identical bytes instead of a disk-cache hit.

## Suggested fix
Persist the payload independently of the publish decision, since `scheduleMediaDiskCacheStore` self-guards on the disk generation:

```swift
let mediaDownload = MessageMediaDownload(...)
// Always persist to the disk cache (self-guarded by isMediaDiskStoreAllowed) so a store-identity
// miss from an interleaved prune doesn't force a re-download.
scheduleMediaDiskCacheStore(mediaDownload, for: cacheKey, accountId: accountId, storeGuard: storeGuard)
if canPublishMediaDownloadState(forKey: key, stateStore: stateStore, accountId: accountId, groupIdHex: groupIdHex) {
    stateStore.update(.loaded(mediaDownload))
} else {
    stateStore.update(.idle)
}
```

## Scope note
Found during a full-codebase review. The tracked media-cache issues (#499/#453/#426/#377 …) are all about *footprint/accounting drift of entries that DO get written*; this is the inverse — an eligible entry that is **never written at all** because the cache-store decision is coupled to the in-memory publish predicate. Distinct root cause and failure mode.

agent-p1p:
Fixes #503

## Summary
- persist a completed media download through the independently generation-guarded encrypted disk cache before checking whether its captured in-memory state store is still publishable
- preserve stale-store behavior: an in-flight task does not publish plaintext into either the pruned store or its replacement
- extend the gated regression test to recreate the store mid-fetch and prove the next load hits disk without a second runtime download

## Verification
- `git diff --check`
- portable source assertions verify disk scheduling precedes the stale-store publish guard and the regression covers cache reuse
- `scripts/ci/macos-sanity-checks.sh` attempted, but this Linux host has no `xcodebuild`; macOS build/unit/static-analysis coverage is delegated to PR CI

## Sensitive paths
None.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/654">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Media.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Media.swift
@@ -284,6 +284,21 @@ extension WorkspaceState {
                     reference: reference
                 )
             }
+            let mediaDownload = MessageMediaDownload(
+                payload: DownloadedMediaPayload(
+                    id: "\(key)|\(UUID().uuidString)",
+                    data: download.plaintext
+                ),
+                fileName: download.fileName,
+                mediaType: download.mediaType,
+                sizeBytes: download.sizeBytes
+            )
+            scheduleMediaDiskCacheStore(
+                mediaDownload,
+                for: cacheKey,
+                accountId: accountId,
+                storeGuard: storeGuard
+            )
             guard
                 canPublishMediaDownloadState(
                     forKey: key,
@@ -295,22 +310,7 @@ extension WorkspaceState {
                 stateStore.update(.idle)
                 return
             }
-            let mediaDownload = MessageMediaDownload(
-                payload: DownloadedMediaPayload(
-                    id: "\(key)|\(UUID().uuidString)",
-                    data: download.plaintext
-                ),
-                fileName: download.fileName,
-                mediaType: download.mediaType,
-                sizeBytes: download.sizeBytes
-            )
             stateStore.update(.loaded(mediaDownload))
-            scheduleMediaDiskCacheStore(
-                mediaDownload,
-                for: cacheKey,
-                accountId: accountId,
-                storeGuard: storeGuard
-            )
         } catch is CancellationError {
             guard isMediaDisplayAllowed(forAccountId: accountId, groupIdHex: groupIdHex) else {
                 stateStore.update(.idle)
--- a/whitenoise-mac/Core/WorkspaceState+Media.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Media.swift
@@ -666,15 +666,22 @@ extension WorkspaceState {
             return
         }
 
-        let prefix = [activeAccountId, groupIdHex, ""].joined(separator: "\u{1F}")
+        let accountPrefix = [activeAccountId, ""].joined(separator: "\u{1F}")
+        let groupPrefix = [activeAccountId, groupIdHex, ""].joined(separator: "\u{1F}")
         let retainedKeys = retainedMediaDownloadKeys(groupIdHex: groupIdHex, accountId: activeAccountId)
         let removedKeys = mediaDownloads.keys.filter { key in
-            guard key.hasPrefix(prefix) else { return true }
+            guard key.hasPrefix(groupPrefix) else { return true }
             return !retainedKeys.contains(key)
         }
         for key in removedKeys {
+            guard let store = mediaDownloads[key] else { continue }
+            if case .loading = store.state, key.hasPrefix(accountPrefix) {
+                // Keep this account's in-flight downloads across chat pruning so a reappearing
+                // tile reuses the same store instead of starting a second runtime call.
+                continue
+            }
             // Notify any lingering per-attachment observers before dropping the store.
-            mediaDownloads[key]?.update(.idle)
+            store.update(.idle)
             mediaDownloads[key] = nil
         }
     }
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -6790,7 +6790,7 @@ struct whitenoise_macTests {
     }
 
     @MainActor
-    @Test func mediaDownloadFinishingAfterTimelinePruneDoesNotPublishPlaintext() async throws {
+    @Test func mediaDownloadFinishingAfterTimelinePruneCachesWithoutPublishingPlaintext() async throws {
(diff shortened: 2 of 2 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Preserve in-flight media stores during pruning](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/654/chapters/1) |
| 2 | [Persist media to disk before pruning check](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/654/chapters/2) |
| 3 | [Verify cache persistence and store coalescing](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/654/chapters/3) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/654">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 073e046209f1838c7734d04965e7d88aca915a07 on Jul 20, 2026 9:54am UTC.</sub>

agent-p1p:
## Pip adversarial review — changes requested

Reviewed exact head `77945ec64fe20e5abb7756985e2e57e8c6c64ee0`.

### Blocking

1. **`whitenoise-macTests/whitenoise_macTests.swift:6882-6909` — the regression does not exercise the issue's A→B→A-before-completion path, and the implementation still permits the redundant fetch.**

   The test recreates `recreatedStateStore` while the first download is gated, but it does not start `loadMediaAttachment` for that replacement until *after* the first download has completed and its disk-store task has run. Production does not wait: once the recreated store is visible and `.idle`, `AutomaticMediaDownloadModifier` calls `loadMediaAttachment` immediately (`whitenoise-mac/Views/MessageMediaViews.swift:525-544`). If that happens before the original download completes, the replacement load misses disk and calls `downloadMedia` a second time. Moving `scheduleMediaDiskCacheStore` before the stale-store publish guard cannot retroactively prevent that already-started fetch.

   Please make the regression start the replacement load before releasing the first gated download and prove `downloadMediaCallCount` remains 1. The implementation needs to preserve/coalesce the in-flight work by media/cache key (or otherwise prevent the recreated store from starting a second runtime download), rather than only making a later post-completion load hit disk.

2. **`whitenoise-macTests/whitenoise_macTests.swift:6889-6896` — the new regression is red in required CI.**

   Mac CI run [29730361048](https://github.com/marmot-protocol/whitenoise-mac/actions/runs/29730361048) failed the new test at line 6895 because `mediaDiskStoreTasks` still contained the store task after the fixed 20-yield loop. The task must be awaited with a deterministic completion condition/helper rather than assuming 20 scheduler yields are sufficient. This failure also skipped the release build smoke step.

### Verification

- `git diff --check origin/master...HEAD`: pass
- portable source/order assertions: pass
- Static Analysis: pass
- Swift format + project sanity checks: pass in Mac CI
- Unit tests: **fail** (new regression above)
- Local Xcode execution unavailable on this Linux worker

No sensitive paths touched. No out-of-scope follow-up issues filed.

agent-p1p:
Addressed both blocking findings in signed head 073e046 and moved the final change to clean replacement PR #660 because #654 recorded red CI during the TDD RED step. The implementation now preserves same-account `.loading` stores across an actual A→B→A chat prune, so the replacement load sees the same in-flight store and `downloadMediaCallCount` remains 1. The regression starts that load before gate release and uses bounded `waitFor` completion; Mac CI run 29732640333 passed PR Checks and Static Analysis before the clean rebase.

agent-p1p:
Superseded by #660. Closing rather than merging an ever-red PR.

PR #660, opened 2026-07-20
fix: coalesce media downloads across chat pruning

### LINKED ISSUE #503: Performance: a completed media download is discarded (and re-fetched) when the in-memory store was pruned+recreated mid-fetch by a chat switch

agent-p1p:
Fixes #503

Supersedes #654. That PR intentionally recorded red Mac CI while proving the strengthened regression, so this clean replacement is required by Pip policy.

## Summary
- coalesce each underlying media runtime fetch by encrypted disk-cache key, independently of the transient SwiftUI waiter/state store
- let canceled or timed-out UI waiters release their limiter slots while a replacement tile can join the same cancellation-unaware UniFFI fetch
- persist a successful shared result exactly once through the existing generation-guarded encrypted disk cache, then token-clean the tracked task
- cancel/invalidate shared tasks on account/global media reset so stale completions cannot cache or publish
- add a deterministic A→B→A regression that cancels the first waiter, recreates the store, starts the replacement load before gate release, and proves one runtime download

## Verification
- RED: Mac run 29733609788 compiled the cancellation regression and failed with `downloadMediaCallCount == 2` at all three assertions
- GREEN: Mac workflow run 29734849804 passed Swift format, project sanity checks, all 541 unit tests, release arm64 build, and Static Analysis
- FINAL CLEAN PR: Mac run 29735336026 passed PR Checks and Static Analysis at signed head `a0fa3cf`
- rebased onto current `origin/master` (`a238a57`) with post-rebase marker, diff, source-order, task-lifecycle, and privacy assertions passing
- `git diff --check`

## Sensitive paths
None.

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Media.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Media.swift
@@ -18,6 +18,10 @@ private nonisolated struct MediaReferenceIndexTaskFailure: Error {
     let underlying: Error
 }
 
+private nonisolated struct MediaAttachmentDownloadTaskFailure: Error {
+    let underlying: Error
+}
+
 @MainActor
 extension WorkspaceState {
     func clearAllComposerDrafts() {
@@ -277,13 +281,17 @@ extension WorkspaceState {
                 return
             }
 
-            let download = try await withMediaAttachmentDownloadTimeout {
-                try await client.downloadMedia(
-                    accountRef: accountRef,
-                    groupIdHex: groupIdHex,
-                    reference: reference
-                )
-            }
+            let download = try await sharedMessageMediaDownload(
+                cacheID: cacheKey.cacheID,
+                cacheKey: cacheKey,
+                mediaDownloadKey: key,
+                accountId: accountId,
+                accountRef: accountRef,
+                groupIdHex: groupIdHex,
+                reference: reference,
+                client: client,
+                storeGuard: storeGuard
+            )
             guard
                 canPublishMediaDownloadState(
                     forKey: key,
@@ -295,22 +303,7 @@ extension WorkspaceState {
                 stateStore.update(.idle)
                 return
             }
-            let mediaDownload = MessageMediaDownload(
-                payload: DownloadedMediaPayload(
-                    id: "\(key)|\(UUID().uuidString)",
-                    data: download.plaintext
-                ),
-                fileName: download.fileName,
-                mediaType: download.mediaType,
-                sizeBytes: download.sizeBytes
-            )
-            stateStore.update(.loaded(mediaDownload))
-            scheduleMediaDiskCacheStore(
-                mediaDownload,
-                for: cacheKey,
-                accountId: accountId,
-                storeGuard: storeGuard
-            )
+            stateStore.update(.loaded(download))
         } catch is CancellationError {
             guard isMediaDisplayAllowed(forAccountId: accountId, groupIdHex: groupIdHex) else {
                 stateStore.update(.idle)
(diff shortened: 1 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Define media download task types and models](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660/chapters/1) |
| 2 | [Implement shared media download task logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660/chapters/2) |
| 3 | [Wire shared downloads into media loading flow](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660/chapters/3) |
| 4 | [Refine chat pruning and cleanup behavior](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660/chapters/4) |
| 5 | [Verify download coalescing with regression tests](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/660">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit a0fa3cf9b1e02adb464826ecc4d34cf4eb8e36ce on Jul 20, 2026 10:31am UTC.</sub>

agent-p1p:
Addressed both adversarial blockers from #654. The regression now models the real SwiftUI cancellation path: the first waiter is canceled on A→B, its store is pruned, and the recreated A tile joins a tokenized shared runtime task before the gate is released. It also waits deterministically for both shared-download and disk-store cleanup. Exact signed head `a0fa3cf` is green in clean PR run 29735336026 (PR Checks + Static Analysis); no actionable review comments remain.

PR #662, opened 2026-07-20
fix: normalize pasted marmot profile links

### LINKED ISSUE #632: Bug: pasting the app's own marmot://profile/<npub> share/QR link into New Chat or the group-invite field fails to resolve — the raw URL (scheme + ?from=qr) is handed to normalizeMemberRef unstripped
## Severity
**MEDIUM** (the app rejects its own advertised share-link format in the core "add contact / invite" flow)

## Summary
`looksLikeMemberRef` accepts a pasted `marmot://profile/<npub>` string, but `memberRefCandidate` returns that entire URL verbatim to the FFI `normalizeMemberRef`, which expects a bare reference / `nostr:<ref>` / `npub` / NIP-05 / hex — not a `marmot://` URL with a `?from=qr` query. The FFI parse throws and the New Chat (and group-invite) flow surfaces "Enter a valid NIP-05, npub, profile link, or hex public key." Every *trusted* in-app path strips the wrapper first; the pasted-input path does not.

## Location / Evidence
- `whitenoise-mac/Core/WorkspaceState+NewChat.swift:342-344` — `looksLikeMemberRef` returns `true` for the pasted link via `MarmotProfileLink.hasProfileLinkPrefix`.
- `WorkspaceState+NewChat.swift:348-358` — `memberRefCandidate(for:)`:
  ```swift
  if MarkdownLinkPolicy.containsNostrReferenceMarker(query) { return query }  // TRUE — URL contains "npub1"
  if NIP05Identifier(query) != nil { … }
  return query
  ```
  `containsNostrReferenceMarker` substring-matches `npub1`/`nprofile1`, so the whole `marmot://profile/npub1…?from=qr` string is returned unchanged and passed to `client.normalizeMemberRef(memberRef:)` at `WorkspaceState+NewChat.swift:69` and `WorkspaceState+Groups.swift:214`.
- Contrast the trusted path `openProfileReference` (`WorkspaceState+Links.swift:46-64`), which builds `"nostr:\(reference)"` from the **bare** npub extracted by `MarmotProfileLink.profileReference(from:)` before resolving — confirming the FFI contract is `nostr:` / bare forms, not the `marmot://` wrapper.
- The QR/share payload is exactly `marmot://profile/<npub>?from=qr` (`MarmotProfileLink.qrPayload`, `MarmotProfileLink.swift:19`; emitted from `SettingsViews.swift`).

## Failure scenario
User A shares their identity via the app's QR/share sheet (`marmot://profile/<npub>?from=qr`). User B copies that link and pastes it into New Chat (or the group "invite member" field). `looksLikeMemberRef` accepts it and auto-resolve fires, but `normalizeMemberRef` receives the raw `marmot://…?from=qr` string and throws → `lastError` is set and no recipient is added. The app rejects its own share-link format when pasted, even though clicking the same link (the OS deep-link path) works.

All existing `marmot://` issues concern the OS deep-link *entry* path — #402 (focus-steal on activate), #421/#510 (deep link mutating New Chat), #558 (deep link cancelling export), #493 (`@domain` NIP-05 misclassification). None touch `memberRefCandidate` / the pasted-into-New-Chat-or-invite resolution failing to normalize the `marmot://profile/` wrapper. #493 added the `containsNostrReferenceMarker` guard (to *prevent* NIP-05 resolution of an embedded `@domain`) but does not strip the scheme/query, so the raw URL still reaches the FFI.

## Suggested fix
In `memberRefCandidate`, before the nostr-marker check, detect `MarmotProfileLink.hasProfileLinkPrefix(query)` and, if so, extract the bare reference via `MarmotProfileLink.profileReference(from: URL(string: query))` and return `"nostr:\(reference)"` (symmetric with `openProfileReference`), rejecting the input if extraction fails.

agent-p1p:
## Summary
- unwrap strict marmot://profile links before FFI member-reference normalization
- pass the extracted profile reference in the same nostr: form used by deep links
- cover the app-generated QR payload with a regression test

## Verification
- source contract check: pass
- git diff --check: pass
- xcodebuild unavailable on this Linux worker; macOS CI must compile and run the Swift test

Fixes #632

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/662">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+NewChat.swift
+++ b/whitenoise-mac/Core/WorkspaceState+NewChat.swift
@@ -354,6 +354,11 @@ extension WorkspaceState {
     }
 
     func memberRefCandidate(for query: String) async throws -> String {
+        if let url = URL(string: query),
+            let reference = MarmotProfileLink.profileReference(from: url)
+        {
+            return "nostr:\(reference)"
+        }
         // A query carrying a nostr marker must never resolve as NIP-05 — an embedded `@domain`
         // would fire a request at that host before the authoritative FFI parse runs.
         if MarkdownLinkPolicy.containsNostrReferenceMarker(query) {
--- a/whitenoise-macTests/whitenoise_macTests.swift
+++ b/whitenoise-macTests/whitenoise_macTests.swift
@@ -16968,6 +16968,16 @@ struct whitenoise_macTests {
     }
 
     @MainActor
+    @Test func pastedMarmotProfileLinkBecomesNostrMemberRefCandidate() async throws {
+        let state = WorkspaceState(clientFactory: { FakeMarmotRuntime(accounts: []) })
+        let payload = MarmotProfileLink.qrPayload(npub: "npub1alyce")
+
+        #expect(state.looksLikeMemberRef(payload))
+        let candidate = try await state.memberRefCandidate(for: payload)
+        #expect(candidate == "nostr:npub1alyce")
+    }
+
+    @MainActor
     @Test func resolvingNewChatRecipientUsesNIP05() async throws {
         let account = desktopAccount()
         let aliceId = "alice1234567890alice1234567890alice1234567890alice1234567890"


ghost:

| | Title |
|---|-------|
| 1 | [Normalize pasted marmot profile links to nostr references](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/662/chapters/1) |
| 2 | [Verify marmot link normalization with regression test](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/662/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/662">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 6eb2e7cd42e57fd6f4ae443f7bc872f2e9ff726e on Jul 20, 2026 9:59am UTC.</sub>

agent-p1p:
BLOCKING:
- The PR encodes a false premise and adds a redundant production branch: `whitenoise-mac/Core/WorkspaceState+NewChat.swift:357-361` — At the pinned MDK commit (`Vendored/MarmotKit/MARMOT_VERSION` → `f3ece0e3`), `normalize_member_ref` already accepts `marmot://profile/{npub}?from=qr`. This is documented in the vendored Swift binding (`Vendored/MarmotKit/Sources/MarmotKit/MarmotKit.swift:1990-1993`) and explicitly exercised by MDK's `crates/marmot-uniffi/tests/smoke.rs::normalize_member_ref_accepts_profile_and_nostr_forms`. Before this PR, the pasted URL reached FFI verbatim through `containsNostrReferenceMarker` and was valid. Converting it to another already-valid form (`nostr:<ref>`) therefore fixes no observable defect and adds an unnecessary special case. Please drop the production change unless there is a concrete failing input against the pinned MDK that contradicts its contract and smoke test.

SUGGESTION:
- The new test only locks in an internal string rewrite, not the reported behavior: `whitenoise-macTests/whitenoise_macTests.swift:16971-16978` — It proves `memberRefCandidate` returns `nostr:npub1alyce`; it does not prove the pre-PR raw URL failed FFI normalization. A behavioral regression test would need to demonstrate the reported failure against the real/pinned FFI contract, which currently says and tests the opposite.
- The pre-existing end-to-end test is now misleading and passes accidentally through the fake's permissive fallback: `whitenoise-macTests/whitenoise_macTests.swift:17394-17413` — Its comment says the raw URL reaches FFI verbatim, and it installs a result under that raw key. After this PR FFI receives `nostr:npub1alyce`, so the installed fixture is unused; `FakeMarmotRuntime.normalizeMemberRef` falls back to a canned Alice result at `whitenoise-macTests/whitenoise_macTests.swift:20640-20649`. If the rewrite is retained, update the comment and exact installed key so the test asserts the actual call path instead of succeeding by fallback.

NITPICK:
none

SUMMARY: Exact-head CI is green (543 tests, release smoke, and static analysis), and the rewrite does not appear to introduce a security regression. The underlying issue is nevertheless contradicted by the app's vendored FFI contract, MDK's exact-commit smoke test, and the pre-existing app test: raw app-generated Marmot profile URLs are already accepted. The PR adds complexity for a non-existent current bug and leaves contradictory coverage behind.

VERDICT: REQUEST_CHANGES
SENSITIVE_PATHS: none

agent-p1p:
Reviewer finding confirmed. The pinned MDK revision `f3ece0e3` strips the `marmot://profile/` wrapper and query/fragment in `canonical_member_ref_input`. I also ran the exact pinned smoke test:

`cargo test -p marmot-uniffi --test smoke normalize_member_ref_accepts_profile_and_nostr_forms --locked -- --exact`

Result: 1 passed, 0 failed. The production rewrite is redundant, and the new candidate-only test does not reproduce a current failure. Closing this PR rather than preserving code for an invalid premise.

PR #669, opened 2026-07-20
fix: preserve media cache for suppressed projections

### LINKED ISSUE #631: Performance: media-reference resolution cache is cleared for a new media message even when the detached (scrolled-back) window suppresses its insert
## Severity
**LOW** (avoidable main-thread media re-resolution while reading history in an active group)

## Summary
In the timeline projection path, the per-group media-reference resolution cache is invalidated based on whether the mapped upserts introduce a media change, computed and applied *before* `applyProjection` and independently of whether the window actually accepts the row. When the window is detached (`anchored == false`, i.e. the user scrolled back), a brand-new head media message is suppressed by `applyProjection` and never inserted — yet the group-wide cache clear (and the `pruneMediaDownloads` it also gates) has already run.

## Location / Evidence
`whitenoise-mac/Core/WorkspaceState+Timeline.swift:510-555`:
```swift
let anchored = !paging.hasMoreAfter                              // :510
let introducesMediaChange = mappedUpserts.contains { upsert in
    let currentAttachments = timelineStore.lookup[upsert.id]?.mediaAttachments ?? []
    guard !currentAttachments.isEmpty || !upsert.mediaAttachments.isEmpty else { return false }
    return currentAttachments != upsert.mediaAttachments          // new msg: nil→[] vs [media] ⇒ true
}
if introducesMediaChange {
    clearMediaReferenceResolutionCache(forAccountId: account.id, groupIdHex: groupIdHex)  // :528 — runs before applyProjection
}
let result = … timelineStore.applyProjection(…, anchoredToNewest: anchored, …)            // :534
guard result.didChange else { return }                           // :542 — too late; clear already ran
…
pruneMediaDownloads: … || introducesMediaChange                  // :554
```
For a brand-new message `timelineStore.lookup[upsert.id]` is `nil` → `currentAttachments == []`; a media message makes the closure return `true`. When `anchored == false`, `applyProjection` suppresses that head row entirely, so `result.didChange` is `false` and the function returns at :542 — but `clearMediaReferenceResolutionCache` already fired at :528.

## Failure scenario
The user scrolls back to read history in a group with active media traffic. Each incoming photo arrives as a projection delta, is correctly suppressed from the detached window, yet still clears the whole group's media-reference resolution cache, forcing every currently-visible historical media tile to re-run `resolvedMediaReference` (full `listMedia` FFI + linear scan per attachment) on its next query. Steady media traffic keeps re-clearing while the user reads older media.

#365 fixed the *pure delivery-state tick* case (same media re-carried ⇒ `introducesMediaChange == false`). This is the distinct residual it did not address: a genuinely new/changed media upsert whose insert the `anchoredToNewest` suppression drops still invalidates the cache, because the clear is evaluated before `applyProjection` and never consults `anchored` / `result.didChange`. #526 is the resolution cost this amplifies, not the trigger.

## Suggested fix
Gate the cache clear (and the `pruneMediaDownloads` contribution) on the projection *result* — only when `result.didChange` and the change actually touched a media row retained in the window — or short-circuit `introducesMediaChange` when `!anchored` and the upsert is a new head row the window will suppress.

agent-p1p:
Fixes #631

## Summary
- report media-attachment changes from the final retained timeline projection, not proposed upserts
- keep the per-group media-reference cache warm when a detached window suppresses new media
- gate media-download pruning on the same retained-row result
- add store-level and end-to-end regression coverage

## Verification
- `git diff --check`
- signed commit and committed diff inspected
- macOS Xcode/Swift checks unavailable on this Linux worker; PR CI will run the exact project gates

## Sensitive paths
None.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669">
  <picture>
    Open in Stage
  </picture>
</a>

### DIFF
--- a/whitenoise-mac/Core/WorkspaceState+Timeline.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Timeline.swift
@@ -364,7 +364,9 @@ extension WorkspaceState {
         let currentPaging = timelinePagingByChat[groupIdHex]
         // Keep "Delete for me" hides out of every full window replace, not just the incremental path.
         let visibleMessages = filterHiddenMessages(mappedMessages, groupIdHex: groupIdHex)
-        TimelineSignpost.store.interval("replaceMessages", count: visibleMessages.count) {
+        let didChangeMediaAttachments = TimelineSignpost.store.interval(
+            "replaceMessages", count: visibleMessages.count
+        ) {
             replaceMessages(
                 visibleMessages,
                 groupIdHex: groupIdHex,
@@ -376,6 +378,9 @@ extension WorkspaceState {
                 ),
                 editMutations: editMutations
             )
+        }
+        if didChangeMediaAttachments {
+            clearMediaReferenceResolutionCache(forAccountId: account.id, groupIdHex: groupIdHex)
         }
         await markLatestVisibleMessageRead(groupIdHex: groupIdHex, account: account, client: client)
     }
@@ -1190,12 +1195,13 @@ extension WorkspaceState {
         }
     }
 
+    @discardableResult
     func replaceMessages(
         _ messages: [MessageItem],
         groupIdHex: String,
         paging: TimelinePagingState? = nil,
         editMutations: [MessageEditMutation] = []
-    ) {
+    ) -> Bool {
         // The window is already ordered, deduped, and capped by the runtime subscription,
         // so render it as-is. The per-chat id/lookup caches live on the store
         // (`MessageTimelineStore.replace` rebuilds them); we only mark this chat as the one
@@ -1209,7 +1215,7 @@ extension WorkspaceState {
 
         cachedMessageChatIds = [groupIdHex]
         messageTimelineStores = [groupIdHex: timelineStore]
-        timelineStore.replace(
+        let didChangeMediaAttachments = timelineStore.replace(
             with: messages,
             editMutations: editMutations,
             windowLimit: Self.timelineWindowLimit
@@ -1221,6 +1227,7 @@ extension WorkspaceState {
         }
         finishTimelineInitialLoad(groupIdHex: groupIdHex)
         pruneMediaDownloadCache(keeping: groupIdHex)
+        return didChangeMediaAttachments
     }
 
     func finalizeTimelineStoreMutation(
--- a/whitenoise-mac/Core/WorkspaceState+Timeline.swift
+++ b/whitenoise-mac/Core/WorkspaceState+Timeline.swift
@@ -548,21 +548,6 @@ extension WorkspaceState {
         let anchored = !paging.hasMoreAfter
         let timelineStore = ensureMessageTimelineStore(for: groupIdHex)
 
-        // Only invalidate the per-group media-reference resolution cache when a projection
-        // actually introduces new or changed media attachments. A single media send emits a
-        // burst of projection deltas (new row, delivery-state transitions, relay echo,
-        // per-relay acks) that re-carry the same media fields; clearing on every one of those
-        // re-runs full media resolution for visible `sourceEpoch == 0` tiles. Comparing the
-        // mapped `MessageItem.mediaAttachments` against the current store entry (still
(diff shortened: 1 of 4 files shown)


ghost:

| | Title |
|---|-------|
| 1 | [Track media changes in timeline store results](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669/chapters/1) |
| 2 | [Implement deferred media cache invalidation logic](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669/chapters/2) |
| 3 | [Wire media invalidation through WorkspaceState](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669/chapters/3) |
| 4 | [Add regression tests for suppressed media](https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise-mac/pull/669">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 8595fc8afffa622dc1c2be9b46ded19f38dbb0db on Jul 20, 2026 1:04pm UTC.</sub>

agent-p1p:
## Pip adversarial review

**Verdict: changes requested** — 1 blocking finding.

### Blocking

- **`whitenoise-mac/Core/WorkspaceState.swift:406-409` / `whitenoise-mac/Core/WorkspaceState+Timeline.swift:247-280,294-380,1193-1224` — a suppressed media change is forgotten, so the cache can remain stale when that message later enters the window.**

  `didChangeMediaAttachments` is intentionally false when the upsert is absent from the final `lookup`; that correctly avoids clearing the cache while the window is detached. However, no state records that suppressed media. Re-anchoring runs through `loadNewerMessages` → `applyTimelineWindow` → `replaceMessages`, and that full-window replacement never clears the media-reference resolution cache. A media row that arrives with a fallback `sourceEpoch == 0` reference can therefore become retained later while `resolvedMediaReference` still consults the per-group index cached before the media record existed. The index returns the old miss, `listMedia` is not called again, and the newly visible attachment remains unresolved.

  The added end-to-end test stops immediately after suppression, so it proves only that the cache stays warm now. Extend it through the deferred-acceptance path: prime a miss, install the matching full media record, suppress a `sourceEpoch == 0` media projection with `hasMoreAfter == true`, then apply/paginate a full window containing that message and resolve its reference. Expected: the cache invalidates at acceptance, `listMediaCallCount` becomes 2, and the full source epoch is returned. Current path keeps the count at 1 and returns epoch 0.

  Please defer rather than discard the invalidation. For example, retain suppressed media-change IDs (or equivalent pending state) and consume the invalidation when a later projection/window replacement first retains one. Avoid clearing on every full-window replacement; that would move the original performance problem into pagination.

### Checks

- Exact reviewed head: `83b8e5a28e2c5804d3229a114627af1f76446d62`
- `git diff origin/master...HEAD --check`: passed
- Mac CI run 29741908061: `PR Checks` passed (format, sanity, unit tests, release smoke); `Static Analysis` passed
- No local Xcode/Swift toolchain is available on this Linux worker
- Sensitive paths touched: none

### Counts

- Blocking: 1
- Suggestions: 0
- Nitpicks: 0
- Follow-up issues: none

agent-p1p:
Addressed the blocking deferred-invalidation finding in rebased head 8595fc8.

- suppressed media changes record one O(1) earliest-sort-key watermark
- unrelated backward pagination, including media rows, does not consume it
- the first authoritative window retaining media at/after the watermark clears it and invalidates once
- the end-to-end regression primes, installs, suppresses, accepts, and resolves the same sourceEpoch-0 reference (expected listMedia count: 1 → 2)

Post-rebase Mac CI run 29744634227 is green: format, sanity, unit tests, release smoke, and Static Analysis all passed. The branch is based on current master 58305d0 and the PR is ready for review. Earlier runs on this PR were red (format, then test-fixture compilation), so it must not auto-merge. CodeRabbit was requested but is rate-limited and produced no actionable inline findings.

agent-p1p:
@coderabbitai review

PR #767, opened 2026-08-13
Update accounts avatar badges as messages arrive

pepina-dev:
The badges with unread counts by account in the sidebar only get updated when switching to another account. This PR fixes that issue.


pepina-dev:
@coderabbitai review

PR #807, opened 2026-08-23
style: resolve the medium rung to Regular weight

erskingardner:
## Summary
- The type ramp's lightest rung (`WNFontWeight.medium`) now resolves to **Regular (400)** instead of Medium (500) in both framework spellings (`WNFontWeight.swiftUI` / `WNFontWeight.appKit`), so body copy and every non-emphasised run renders one step lighter than before.
- Token names and the three-rung vocabulary are unchanged — `.medium` keeps its name for the Flutter rung it descends from (`lib/theme/app_typography.dart`); only its resolution differs, and the header comment documents that divergence.
- `TypographyTests.bothFrameworksSpellTheSameWeight` updated to pin `.medium` → `.regular` in both frameworks; all 1127 unit tests pass.

## Test plan
- [x] `xcodebuild build-for-testing` succeeds
- [x] `xcodebuild test-without-building` (PR test plan) — 1127 tests pass
- [x] Eyeballed the running app: body text reads lighter than the previous Medium setting

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the medium font weight to render as Regular consistently across SwiftUI and AppKit.
  * Improved typography consistency for text using the medium weight.

### DIFF
--- a/whitenoise-mac/Views/Palette/WNFontWeight.swift
+++ b/whitenoise-mac/Views/Palette/WNFontWeight.swift
@@ -8,17 +8,20 @@
 //  rather than a separate file in the bundle.
 //
 //  The vocabulary stays three deep because the ramp is the Flutter client's ladder
-//  and that ladder names three: body copy is `medium`, and anything a design draws
-//  lighter maps up to it. Each case carries both frameworks' spelling of the same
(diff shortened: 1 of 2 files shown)


pepina-dev (approved):
LGTM!

